PR #1, opened 2020-08-15
[1] Created an initial boilerplate for the project

ankurnarkhede:
- Added the initial boilerplate for the project
- The values for the CORS and security headers are set to the default for now as per offered by the respective packages. It will be modified later if required.
- Added the `GET /healthcheck` route to get the service status

The current state is deployed at: https://rds-apis.herokuapp.com/healthcheck


Kratika0907 (approved):
Few doubts in .gitignore

PR #7, opened 2020-08-16
[2] Route to handle the callback from GitHub Auth

ankurnarkhede:
- Added a route to handle the callback from GitHub Auth
- Passport.js being used to handle the GitHub Oauth
- Added in integration tests


Kratika0907:
Route to handle the callback from GitHub Auth

Issue #5, opened 2020-08-15
Integrate GitHub Actions for CI


ankurnarkhede:
Integrate Github actions for the repo to:
- Run the tests on every pull request raised
- Run on each push, merge commit on `main` branch

swarajpure:
I'd like to work on this issue. Please assign me.

PR #10, opened 2020-08-16
Integrating Github actions for CI

swarajpure:
Integrating Github actions for CI as per #5 

- The Github actions are triggered on push and pull request events.
- There is no trigger event for commit. 
- Triggered on all branches.
- The tests are being run on node v10.x, 12.x, 14.x (please comment if all these are needed).
- I have kept the **build** script (optional) as well if we plan to add any, can be removed if not needed.

## When the tests fail (tested by expecting HTTP reponse 400 instead of 200):

image
image

## When the tests pass:
image
image


swarajpure:
> 
> 
> Can we have a better name to this file - node.js.yml does not tell what this action is supposed to do. I think test.yml or similar name would be okay

Done. Anything else to be done?

PR #23, opened 2020-08-29
Set cookie for realdevsquad.com on successful login

ankushdharkar:


controllers/authController.js
```diff
@@ -35,6 +35,7 @@ const githubAuth = (req, res, next) => {
 
       // respond with a cookie
       res.cookie(config.get('userToken.cookieName'), token, {
+        domain: 'realdevsquad.com',
```
ankurnarkhede, line 38: @ankushdharkar , the domain name won't be the same everytime, e.g, in stage environments or in local.
Can you make this configurable via config files?

PR #33, opened 2020-09-06
[2] Refresh JWT after expiry

ankurnarkhede:
- Refresh JWT after expiry if the token init time is within the limits of the refreshTTL configured
- Minor typo fixes in package.json


whyDontI:
LGTM!

PR #34, opened 2020-09-12
Reduced config redundancy

ankurnarkhede:
- Reduced config redundancy
- Removed default config which was already defined in `default.js`


whyDontI:
Awesome!

PR #49, opened 2020-09-27
Setting incompleteUserDetails for new user

swarajpure:
- Setting incompleteUserDetails for new user who has not complete profile yet.

Please suggest changes if any. Thank you!


models/users.js
```diff
@@ -28,20 +28,22 @@ const addOrUpdate = async (userData, userId = null) => {
         })
       }
 
-      return { isNewUser, userId }
```
whyDontI, line 31: This `isNewUser` is different than what we are storing in DataBase.
We use `isNewUser` in the return response to indicate if we Added a new user or Updated.
This is not related to profile completion.
Please revert to old changes.

models/users.js
```diff
@@ -28,20 +28,22 @@ const addOrUpdate = async (userData, userId = null) => {
...
+      return { incompleteUserDetails, userId }
     }
 
     // userId is null, Add or Update user
     const user = await userModel.where('github_id', '==', userData.github_id).limit(1).get()
     if (!user.empty) {
+      userData.incompleteUserDetails = false
```
whyDontI, line 37: This flag can only be false when user updates his/her `id`.
We should change this flag until the `id` is updated.

whyDontI (approved):

Issue #63, opened 2020-10-13
Throw errors from model functions


ankurnarkhede:
Currently, model functions return an error, while they should be throwing them. The functions calling the model functions never get the error in the catch block due to the current implementation.

swarajpure:
I'd like to take this up.

swarajpure:
Issue solved.

PR #68, opened 2020-10-14
Throw err from model functions instead of returning them

swarajpure:
Solves #63 
- Throw error from model functions
- Added `this.skip()` to the tests which were failing because of lack of test suite as suggested by @ankurnarkhede 

image

image

Feel free to suggest changes. Thank you! 😄


ankurnarkhede (approved):
LGTM

PR #69, opened 2020-10-15
Revert "Throw err from model functions instead of returning them"

ankurnarkhede:
Reverts Real-Dev-Squad/website-backend#68


swarajpure (approved):

swarajpure:
Reason to revert: pushed in the main branch instead of develop.

PR #72, opened 2020-11-06
crypto products api added

prakashchoudhary07:
Products api 
get /crypto/products
get /crypto/products/id
post /crypto/products    body:{data}

Crypto user transaction will be done in next pr


controllers/crypto/productsController.js
```diff
@@ -0,0 +1,79 @@
...
+    const product = await productQuery.fetchProduct(req.params.id)
+    if (product) {
+      return res.status(200).json({
+        message: 'Products returned successfully!',
+        Products: product
+      })
+    }
+    return res.status(404).json({
+      message: 'Products not found!'
+    })
```
whyDontI, line 43: We are using boom for errors,
We can write `return res.boom.notFound('<Error Message>')`

Boom docs -
https://npmdoc.github.io/node-npmdoc-boom/build/apidoc.html

models/crypto/products.js
```diff
@@ -0,0 +1,84 @@
...
+
+    const allProducts = []
+
+    snapshot.forEach((doc) => {
+      allProducts.push({
+        id: doc.id,
+        ...doc.data()
+      })
+    }
+    )
```
whyDontI, line 27: NIT: Can we get these parenthesis on a single line.

models/crypto/products.js
```diff
@@ -0,0 +1,84 @@
...
+    const product = await productsModel.doc(id).get()
+    if (product.exists) {
+      return {
+        product: {
+          id: id,
+          ...product.data()
+        }
+      }
+    }
+    return ''
```
whyDontI, line 52: This line of code is not reachable if the user exists it'll return otherwise promise will be rejected with an error.

Can we refactor this part of the code?
Please let me know if I am missing something.

models/crypto/products.js
```diff
@@ -0,0 +1,84 @@
...
+const addProduct = async (productData) => {
+  try {
+    let productInfo = await productsModel.where('productId', '==', productData.productId).limit(1).get()
+
+    if (productInfo.empty) {
+      productInfo = await productsModel.doc().set(productData)
+      return { productId: productInfo.id }
+    }
+
+    return ''
```
whyDontI, line 73: This sort of return statement doesn't seem right, return some error message or come up with some better approach?

routes/crypto/index.js
```diff
@@ -0,0 +1,7 @@
+const express = require('express')
+const app = express()
+
+app.use('/products/', require('./products'))
+// app.use('/users', require('./users'))
```
whyDontI, line 5: Can we remove this commented code?

routes/crypto/products.js
```diff
@@ -0,0 +1,75 @@
+const express = require('express')
+const router = express.Router()
+const authenticate = require('../../middlewares/authenticate')
+const productsController = require('../../controllers/crypto/productsController')
+
+// TODO: swager documantation
```
whyDontI, line 6: We can remove this `TODO` statement as swagger docs are there in the file.

routes/crypto/products.js
```diff
@@ -0,0 +1,75 @@
...
+ *       401:
+ *         description: Unauthorized
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/errors/unAuthorized'
+ */
+router.post('/', authenticate, productsController.addNewProduct)
+// router.patch('/:id', authenticate, productsController.getSelfDetails)
+// router.delete('/:id', authenticate, productsController.)
```
whyDontI, line 31: Commented code

whyDontI (changes requested):
Great work 🥳 👍🏽 
Requested some little changes.

models/crypto/products.js
prakashchoudhary07: This would return the product if found or else it would return an empty string. We can remove the return statement which means it would return undefined which is the same, but I think this is easy to understand, If not then let me know I will change it 😀

models/crypto/products.js
prakashchoudhary07: If the user does not exist then it will return undefined

models/crypto/products.js
Kratika0907: Instead of empty string, can we have message that product was not found.

prakashchoudhary07: That has been taken care of in the controller.js file [`here`](https://github.com/Real-Dev-Squad/website-backend/pull/72/files/184c32c5c2771fd61aa937ac213474fbaafbc21f#diff-1bf082f4fa1b819960342e413ebff7efc8893704defa73f4cf3ed3e434cd1a6eR60-R72)

ankurnarkhede:
@prakashchoudhary07 , any update here?

prakashchoudhary07:
API contracts have been changed, So closing the PR

PR #75, opened 2020-11-07
Addition of PATCH users/self route

Rucha1499:
1) Added the PATCH users/self route. 
2) Added the username field to user Schema.
3) Deleted the PATCH users/:id route.

### DIFF
--- a/controllers/usersController.js
+++ b/controllers/usersController.js
@@ -104,9 +104,9 @@ const updateSelf = async (req, res) => {
     const token = (req.headers.authorization.split(' '))[1]
     const { userId } = decodeAuthToken(token)
 
-    if(req.body.hasOwnProperty('username')) {
+    if (req.body.username) {
       const { user } = await userQuery.fetchUser(userId)
-      if(!user.incompleteUserDetails) {
+      if (!user.incompleteUserDetails) {
         return res.boom.forbidden('Cannot update username again')
       }
       await userQuery.setIncompleteUserDetails(userId)
--- a/controllers/usersController.js
+++ b/controllers/usersController.js
@@ -127,10 +127,9 @@ const updateUser = async (req, res) => {
  */
 const updateSelf = async (req, res) => {
   try {
-    const authorization = req.headers.authorization
-    const token = (authorization.split(' '))[1]
+    const token = (req.headers.authorization.split(' '))[1]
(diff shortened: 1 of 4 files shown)


middlewares/validators/user.js
```diff
@@ -30,6 +31,7 @@ const createUser = async (req, res, next) => {
 
 const updateUser = async (req, res, next) => {
   const schema = joi.object().keys({
+    user_name: joi.string().optional(),
```
swarajpure, line 34: Shouldn't we be making `user_name` and other fields like `yoe`, `company` required for the `updateUser` function? Otherwise someone may set `incompleteUserDetails` to false through postman by calling this API and get away without filling the form which sujit is working on.

whyDontI: - One can't update `incompleteUserDetails` field because we don't have validation for it. Validator won't allow that field in data.
- We should not set other fields as `required`, because if I only want to change my YOE I won't send other data. User does not update the whole information all the time, only required fields are updated which we can't predict for each user.

controllers/usersController.js
```diff
@@ -92,21 +93,30 @@ const addNewUser = async (req, res) => {
...
  * @param req {Object} - Express request object
- * @param req.params.id {string} - User id
  * @param req.body {Object} - User object
  * @param res {Object} - Express response object
  */
-const updateUser = async (req, res) => {
+const updateSelf = async (req, res) => {
   try {
-    const user = await userQuery.addOrUpdate(req.body, req.params.id)
+    const token = (req.headers.authorization.split(' '))[1]
```
ankushdharkar, line 104: I thought we take our token from the `rds-session` cookie. What is this authorization header?

ankushdharkar (changes requested):

controllers/usersController.js
whyDontI: Thats `Bearer JWT` token

ankushdharkar: What is that @whyDontI?

ankushdharkar (approved):
Thanks @Rucha1499!

controllers/usersController.js
whyDontI: Sorry, I totally forgot that we use Cookie in prod.
Well have to change this code to get token from cookie.
My bad 😶

Issue #51, opened 2020-10-01
Automation for checking local development environment is properly setup or not.


takshch:
# Description
Automation for checking local development environment is properly setup or not.

# Steps
1. Check whether local.js exits or not.
2. Check whether local.js contains GitHub Oauth clientId and ClientSecret.
3. Validation for the GitHub Oauth clientId and ClientSecret.
4. Check whether firestore-private-key.json exits or not
5. Checking for the connection for firestore and permission to read and write by adding dummy data and removing that dummy data.

Tasks

- [x] Firestore connection
- [ ] Github Authentication
- [ ] JWT Token

PR #82, opened 2020-11-17
Automation for checking local development environment is properly setup or not.

takshch:
#51

### DIFF
--- a/local-test.js
+++ b/local-test.js
@@ -23,21 +23,15 @@ if (!firestoreData) {
   throw new Error('Please make sure firestore-private-key.json file is not empty')
 }
 
-readWriteCheck()
-
 // check local development have permission to read and write in firestore or not
-async function readWriteCheck () {
-  try {
-    const docRef = await Firestore.collection('dummy').doc('users')
-    await docRef.set({
-      user: 'dummy'
-    })
-    const resp = await docRef.get('user')
-    if (resp.data().user !== 'dummy') {
-      throw new Error('Problem with permission of read and write.\nCheck your firestore permissions')
-    }
-    await docRef.delete()
-  } catch (err) {
-    throw new Error(err)
+(async () => {
+  const docRef = await Firestore.collection('dummy').doc('users')
+  await docRef.set({
+    user: 'dummy'
+  })
+  const resp = await docRef.get('user')
+  if (resp.data().user !== 'dummy') {
+    throw new Error('Problem with permission of read and write.\nCheck your firestore permissions')
   }
-}
+  await docRef.delete()
+})()
--- a/local-test.js
+++ b/local-test.js
@@ -1,0 +1,43 @@
+const fs = require('fs')
+const Firestore = require('./utils/firestore.js')
+const config = require('config')
+
+// check whether github credentials are not falsy
+if (
+  !config.githubOauth ||
+  !config.githubOauth.clientId ||
+  !config.githubOauth.clientSecret
+) {
+  throw new Error('Github credentials are not properly set')
+}
+
+// throw an error if unable to read file
+try {
+  var firestoreData = fs.readFileSync('./firestore-private-key.json', 'utf8')
+} catch (error) {
+  throw new Error('Please make sure firestore-private-key.json file is correct')
+}
+
+// check whether firestoreData is empty, null, and  undefined
+if (!firestoreData) {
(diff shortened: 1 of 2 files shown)


takshch:
@ankurnarkhede Sir previous PR was having some problem which can't be solved from GitHub. So I created a New PR. I have made the required changes. 
Please review and merge this PR.

local-test.js
```diff
@@ -0,0 +1,43 @@
...
+try {
+  var firestoreData = fs.readFileSync('./firestore-private-key.json', 'utf8')
+} catch (error) {
+  throw new Error('Please make sure firestore-private-key.json file is correct')
+}
+
+// check whether firestoreData is empty, null, and  undefined
+if (!firestoreData) {
+  throw new Error('Please make sure firestore-private-key.json file is not empty')
+}
```
takshch, line 24: Fixed!

package.json
```diff
@@ -7,6 +7,7 @@
     "dev": "cross-env NODE_ENV=development nodemon server.js",
     "lint": "eslint .",
     "lint-fix": "eslint . --fix",
+    "test-config": "node local-test.js",
```
takshch, line 10: Now there will be no local config check for running development server.
Only test-config variable will be responsible for checking local configuration.

local-test.js
```diff
@@ -0,0 +1,43 @@
...
+  !config.githubOauth ||
+  !config.githubOauth.clientId ||
+  !config.githubOauth.clientSecret
+) {
+  throw new Error('Github credentials are not properly set')
+}
+
+// throw an error if unable to read file
+try {
+  var firestoreData = fs.readFileSync('./firestore-private-key.json', 'utf8')
```
nimish-gupta, line 16: Can we use const over here, since the variable is not changing?

local-test.js
```diff
@@ -0,0 +1,43 @@
+const fs = require('fs')
+const Firestore = require('./utils/firestore.js')
+const config = require('config')
+
+// check whether github credentials are not falsy
+if (
```
nimish-gupta, line 6: Will it be a good thing if we wrap this thing in a function and then call the function within readWriteCheck? We can do this for both GitHub/ firebase validations.

local-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+// check whether firestoreData is empty, null, and  undefined
+if (!firestoreData) {
+  throw new Error('Please make sure firestore-private-key.json file is not empty')
+}
+
+readWriteCheck()
+
+// check local development have permission to read and write in firestore or not
+async function readWriteCheck () {
```
nimish-gupta, line 29: Can we use arrow functions here? Is eslint setup for the repository?

local-test.js
```diff
@@ -0,0 +1,43 @@
...
+    await docRef.set({
+      user: 'dummy'
+    })
+    const resp = await docRef.get('user')
+    if (resp.data().user !== 'dummy') {
+      throw new Error('Problem with permission of read and write.\nCheck your firestore permissions')
+    }
+    await docRef.delete()
+  } catch (err) {
+    throw new Error(err)
```
nimish-gupta, line 41: Are we returning the same error? I think then we can skip try-catch. We can use try-catch if we want to override the error or add some logging or any other customization on error.

local-test.js
nimish-gupta: Reason - we can unit test the validations. Also, we will be knowing where validations are required explicitly.

local-test.js
takshch: var can't be access outside  of block scope.
If i will use const then variable will not be accessed for if condition on line 22.

If i declare a const variable but doesn't initialize then i will not be able to assign other value to the variable.
So, var is instead of const.

local-test.js
takshch: I want to know more. Can we come talk for few minutes on discord?

local-test.js
takshch: Same.

local-test.js
takshch: I removed try-catch and also used IIFE for async arrow functions.
I don't know async arrow function with IIFE is  best practices or not.

local-test.js
nimish-gupta: yes sure

ankurnarkhede (approved):

PR #85, opened 2020-11-23
added library function for network call

### LINKED ISSUE #84: Make a wrapper function for network calls
Make a wrapper function for network calls so that whenever we want to perform a network call, we can call this function.
In case we choose to use another library for network calls we can just make changes in this particular function rather than changing the code at every network call. This makes our code modular.

swarajpure:
fixes #84


lib/networkCall.js
```diff
@@ -0,0 +1,17 @@
+const fetch = require('node-fetch')
+
+const networkCall = async (url, method = 'get', body = null, headers = null) => {
```
ankurnarkhede, line 3: What is the benefit of this function?
Does directly using `fetch` in our service functions lack anything?
Is this done to stringify the response?

ankurnarkhede, line 3: What is we need to introduce request timeout here as we interact with external services?

ankurnarkhede (changes requested):

lib/networkCall.js
swarajpure: Mentioned the purpose in issue description
https://github.com/Real-Dev-Squad/website-backend/issues/84

```
Make a wrapper function for network calls so that whenever we want to perform a network call, we can call this function.
In case we choose to use another library for network calls we can just make changes in this particular function rather than changing the code at every network call. This makes our code modular.
```

swarajpure: Will see how to go about it.

ankurnarkhede, line 3: Constructing the whole URL everytime is a tedious task. Can we accept the query params in an argument tot teh function?

ankurnarkhede (changes requested):

lib/networkCall.js
```diff
@@ -0,0 +1,17 @@
...
+  try {
+    const response = await fetch(url, {
+      method: method,
+      body: body,
+      headers: headers
+    })
+    const json = await response.json()
+    return json
+  } catch (err) {
+    return err
```
ankurnarkhede, line 13: When we are returning error, the error will be obtained in the then block only. We will need to throw it from here.
Also, can we handle throwing of 4XX and 5XX errors too.
Refer the limitation of the library here: https://medium.com/@thejasonfile/fetch-vs-axios-js-for-making-http-requests-2b261cdd3af5

ankurnarkhede (changes requested):

lib/networkCall.js
swarajpure: Error throwing done.

lib/networkCall.js
```diff
@@ -0,0 +1,20 @@
...
+    const response = await axios({
+      method: method,
+      url: url,
+      params: queryParams,
+      data: body,
+      headers: headers
+    })
+    return response
+  } catch (err) {
+    logger.error('Something went wrong. Please contact admin')
```
ankurnarkhede, line 15: Add the error message in the logs.

lib/networkCall.js
```diff
@@ -0,0 +1,20 @@
+const axios = require('axios')
+const logger = require('../utils/logger')
+
+const networkCall = async (url, queryParams, method = 'get', body = null, headers = null) => {
```
whyDontI: Please add default value for `queryParams` too

lib/networkCall.js
```diff
@@ -0,0 +1,20 @@
...
+  try {
+    const response = await axios({
+      method: method,
+      url: url,
+      params: queryParams,
+      data: body,
+      headers: headers
```
whyDontI, line 11: You can use ShortHand syntax here, 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

whyDontI (changes requested):
Great work 🔥 👍🏽 
Requested few changes, rest all is good 👍🏽

lib/networkCall.js
```diff
@@ -0,0 +1,20 @@
+const axios = require('axios')
+const logger = require('../utils/logger')
+
+const networkCall = async (url, method = 'get', queryParams = null, body = null, headers = null) => {
```
ankurnarkhede, line 4: Can we also add `options = {}` argument to the function and add `...options` in the axios call.
This will be useful in case we need to add other provided oprions to the call rather than the frequently used ones.

Add JSDOC.

And, `networkCall` seems a long name for the function which is to be used too frequently in the codebase.
Can we change the name to anything like `request` or anything else short? 😬

ankurnarkhede (changes requested):
LGTM, will be merged post the comments are addressed.

lib/networkCall.js
swarajpure: Thanks for pointing out the shorthand syntax. Actually axios uses `data` keyword for `body` and I suppose `body` word is more familar to developers than `data`. I personally never came across `data` for `body`. So I was thinking to use something like `data: body`
What do you say?

whyDontI: body is an argument to the function, where ever we are going to call the function, it will require 4th argument to be of type `object`, the argument name doesn't matter for the user of the function.

If you think body will be a better idea then go with it, else `data` with shorthand syntax will always be a better option.

swarajpure: I agree. Will stick to the shorthand syntax.

lib/networkCall.js
swarajpure: done.

lib/networkCall.js
ankurnarkhede: Same for `params` object! @swarajpure

swarajpure: please check again @ankurnarkhede

ankurnarkhede (dismissed):

ankurnarkhede (approved):

whyDontI (approved):
👍🏽

PR #95, opened 2020-12-09
Changed API routes and functions to take `username` of user along with `id`

### LINKED ISSUE #94: Change API routes such that they take `username` instead of `id`
Currently our APIs like `pullrequests/:id` and `users/:id` take `id` as a path param. 
We need to change it so that it takes `username` as path param and proceed in the exact same way.

swarajpure:
Changed following API routes:
- `pullrequests/:id`
- `users/:id`

Also changed corresponding functions to accomodate and work with `username` of user ~instead of~ along with `id` of user

Fixes #94 

Please review. Thank you!


middlewares/authenticate.js
```diff
@@ -34,7 +34,7 @@ module.exports = async (req, res, next) => {
     const decoded = authService.verifyAuthToken(token)
 
     // add user data to `req.userData` for further use
-    const userData = await users.fetchUser(decoded.userId)
+    const userData = await users.fetchUser(null, decoded.userId)
```
whyDontI, line 37: **Suggestion: **
Passing `null` as an argument does not feel good.
Is it a better idea to make the argument as an Object which will contain `userId` and `userName`

The function call will look like this
```
const userData = await users.fetchUser({
    userId: 21389374,
    userName: 'Swarajpure'
})
```

Opinions: @ankurnarkhede @swarajpure

swarajpure: Awesome idea @whyDontI !
Had not at all thought of it.
Pushed commit with this change. Please review.

Thank you so much! 😄

models/users.js
```diff
@@ -79,20 +79,31 @@ const fetchUsers = async (query) => {
...
 /**
- * Fetches the user data from the passes userId
+ * Fetches the user data from the the provided username or userId
  *
- * @param userId { string }: User id
+ * @param { Object }: Object with username and userId, any of the two can be used
  * @return {Promise<{userExists: boolean, user: <userModel>}|{userExists: boolean, user: <userModel>}>}
  */
-const fetchUser = async (userId) => {
+const fetchUser = async ({ userId, username }) => {
```
whyDontI, line 87: Add default values,
```
{ userId = null, userName = null }
```

models/users.js
```diff
@@ -79,20 +79,31 @@ const fetchUsers = async (query) => {
...
-    const user = await userModel.doc(userId).get()
+    let userData, id
+    if (username) {
+      const user = await userModel.where('username', '==', username).limit(1).get()
 
+      user.forEach(doc => {
+        id = doc.id
+        userData = doc.data()
+      })
+    } else {
```
whyDontI, line 97: Put a check for userId, both values can also be null.

whyDontI (changes requested):

middlewares/authenticate.js
```diff
@@ -58,7 +58,7 @@ module.exports = async (req, res, next) => {
         })
 
         // add user data to `req.userData` for further use
-        req.userData = await users.fetchUser(decoded.userId)
+        req.userData = await users.fetchUser({ userId: decoded.userId })
```
whyDontI, line 61: Destructure `decoded` it's being used at multiple places

whyDontI (changes requested):

models/users.js
swarajpure: I didn't understand how can both values be null?

middlewares/authenticate.js
swarajpure: Yes sure

models/users.js
swarajpure: Done.

middlewares/authenticate.js
swarajpure: Done with requested changes. Please review.

models/users.js
whyDontI: Edge case

middlewares/authenticate.js
whyDontI: Did you push your changes?
Can't see destructured `decoded` varialble

middlewares/authenticate.js
```diff
@@ -31,10 +31,10 @@ module.exports = async (req, res, next) => {
       token = req.headers.authorization.split(' ')[1]
     }
 
-    const decoded = authService.verifyAuthToken(token)
+    const { userId } = authService.verifyAuthToken(token)
```
swarajpure, line 34: @whyDontI destructured it here

whyDontI (approved):
Great work @swarajpure 👍🏽

PR #97, opened 2020-12-14
Fetch oldest Open PRs

### LINKED ISSUE #96: API for N-oldest Open PRs
For the feature at Real-Dev-Squad/website-status#15

ankushdharkar:
Fetch oldest PRs from Real Dev Squad sorted by creation date and pagination

Closes #96

### DIFF
--- a/controllers/pullRequestsController.js
+++ b/controllers/pullRequestsController.js
@@ -43,6 +43,52 @@ const getPRdetails = async (req, res) => {
   }
 }
 
+const getStalePRs = async (req, res) => {
+  try {
+    const { data } = await githubService.fetchOpenPRs()
+
+    if (data.total_count) {
+      const allPRs = []
+      data.items.forEach(({
+        title,
+        html_url: url,
+        state,
+        created_at: createdAt,
+        updated_at: updatedAt,
+        draft,
+        labels,
+        user,
+        assignees
+      }) => {
+        const allAssignees = assignees.map(object => object.login)
+        const allLabels = labels.map(object => object.name)
+        allPRs.push({
+          title,
+          state,
+          createdAt,
+          updatedAt,
+          url,
+          username: user.login,
+          labels: allLabels,
+          assignees: allAssignees
+        })
+      })
+      return res.json({
+        message: 'Stale PRs',
+        pullRequests: allPRs
+      })
+    }
+    return res.json({
+      message: 'No pull requests found!',
+      pullRequests: []
+    })
+  } catch (err) {
+    logger.error(`Error while processing pull requests: ${err}`)
+    return res.boom.badImplementation('Something went wrong please contact admin')
+  }
+}
+
 module.exports = {
-  getPRdetails
+  getPRdetails,
+  getStalePRs
 }
--- a/controllers/pullRequestsController.js
+++ b/controllers/pullRequestsController.js
@@ -43,6 +43,13 @@ const getPRdetails = async (req, res) => {
   }
 }
 
+/**
+ * Get stale PRs in open state for Real Dev Squad repos
+ *
+ * @param {Object} req
+ * @param {Object} res
+ * @todo create cache for RDS usernames <> github usernames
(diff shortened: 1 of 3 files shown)


controllers/pullRequestsController.js
```diff
@@ -43,6 +43,52 @@ const getPRdetails = async (req, res) => {
...
+        draft,
+        labels,
+        user,
+        assignees
+      }) => {
+        const allAssignees = assignees.map(object => object.login)
+        const allLabels = labels.map(object => object.name)
+        allPRs.push({
+          title,
+          state,
```
swarajpure, line 67: Do we really need to return the state when we already know it's stale (open) ?

controllers/pullRequestsController.js
```diff
@@ -43,6 +43,52 @@ const getPRdetails = async (req, res) => {
...
+          assignees: allAssignees
+        })
+      })
+      return res.json({
+        message: 'Stale PRs',
+        pullRequests: allPRs
+      })
+    }
+    return res.json({
+      message: 'No pull requests found!',
```
swarajpure, line 82: I guess this will make more sense:
```No stale pull requests found!```

What do you think?

File: routes/pullrequests.js [hunk 3]
```diff
@@ -2,6 +2,37 @@ const express = require('express')
...
+ *              schema:
+ *                type: object
+ *                properties:
+ *                  message:
+ *                    type: string
+ *                    example: Stale PRs
+ *                  pullRequests:
+ *                    type: array
+ *                    items:
+ *                      $ref: '#/components/schemas/pullRequests'
```
swarajpure, line 26: This schema seems to be just slightly different from the `pullRequests` schema, like this one has got `username` of the author as well.

I'm not sure about this, but should we create a new schema just for one/two change(s)?

controllers/pullRequestsController.js
```diff
@@ -43,6 +43,52 @@ const getPRdetails = async (req, res) => {
   }
 }
 
+const getStalePRs = async (req, res) => {
```
swarajpure, line 46: Please add JSDoc.

services/githubService.js
```diff
@@ -25,6 +75,24 @@ const fetchPRsByUser = async (username) => {
   }
 }
 
+const fetchOpenPRs = async () => {
```
swarajpure, line 78: JSDoc annotations missing

swarajpure (changes requested):
Requesting some changes and clarifying some things.
Also, please consider adding API contracts for the same.

Thank you! A very much needed feature to keep track of stale PRs....
Finally the `shoes` on the crypto site will be put to use now 😛

controllers/pullRequestsController.js
ankushdharkar: Good point. I deliberately kept it there to control the text we will show on the Frontend against that PR row. In future, instead of "Open", if we want to call it "Unattended", we simply need to make backend changes, and when the user sees it will not be affected by the assets caching on browser or CDN

controllers/pullRequestsController.js
ankushdharkar: We aren't using this message anywhere. So, it doesn't really make much of a difference.

routes/pullrequests.js
ankushdharkar: You are absolutely right though @swarajpure! Great catch!

There is a BIG problem with the username, or rather usernames.
And I want @sumitd94 @whyDontI to resolve it.
We need to implement a Remote Dictionary Server (Redis) for a few reasons. We will have a session on it.

services/githubService.js
ankushdharkar: Added in the review changes. Thank you for catching this.

controllers/pullRequestsController.js
ankushdharkar: Added in the review-related changes. Thanks!

PR #106, opened 2020-12-22
Document about getting cookie in local

whyDontI:


README.md
```diff
@@ -37,3 +37,16 @@ $ npm run dev
...
+- Run the project locally, make sure the server is listening to requests
+- Open a new tab in the browser
+- Open the `Network` tab in Browsers Developer Tools
+- Make sure the `Preserve log` checkbox is selected  
+  ![Preserve logs checkbox](https://user-images.githubusercontent.com/26569942/102855521-cea32880-444a-11eb-84e3-02ee6fa5b830.png)
+- Put the following URL in browsers URL bar with the clientId you got from GitHub
+    <br />
+    `https://github.com/login/oauth/authorize?client_id=<clientId>`
+- The link will take you to the GitHub login flow and it will redirect to registered callback url after authentication
+- In the browsers Dev tools Network tab, look for `http://localhost:3000/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>` request. This is the response that actually sets the cookie on the website, but because we made the call from localhost, it will not be honoured by the browser, but we can manually copy the cookie from **Response Headers** and use it for our Postman and Swagger calls.
```
ankurnarkhede, line 51: Can we simply add the URL `<HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>` in the third point here https://github.com/Real-Dev-Squad/website-backend/blob/develop/CONTRIBUTING.md#generating-authentication-token to make it more explicit?

ankurnarkhede (changes requested):

README.md
whyDontI: Sure, I think it'll be better if we move all this to the `CONTRIBUTION.md`.
I'll make the changes

CONTRIBUTING.md
```diff
@@ -80,9 +80,18 @@ The following project structure should be followed:
...
+- Run the project locally, make sure the server is listening to requests
+- Open a new tab in the browser
+- Open the `Network` tab in Browsers Developer Tools
+- Make sure the `Preserve log` checkbox is selected
+    <br />
+    ![Preserve logs checkbox](https://user-images.githubusercontent.com/26569942/102855521-cea32880-444a-11eb-84e3-02ee6fa5b830.png)
+- Navigate to `https://github.com/login/oauth/authorize?client_id=<GITHUB_CLIENT_ID>` in the same tab
 - Authorize the application.
-- Copy the response cookie named `rds-session` from the redirected request from the API.
+- In the browsers Dev tools Network tab, look for `<HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>` request. This is the response that actually sets the cookie on the website. If we are making the call from localhost, it will not be honoured by the browser, but we can manually copy the cookie from **Response Headers** and use it for our Postman and Swagger calls.
```
ankurnarkhede, line 91: Can we just change this line to:

- Copy the response cookie named `rds-session` from the redirected request `<HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>`.


We already have mentioned about using the token as bearer token in non production environments.
The other instructions mentioned about opening tab, preserve log is something too explicit and can be removed IMO. 🤔

ankurnarkhede (changes requested):
Minor comment, can be merged post its addressed.

CONTRIBUTING.md
ankurnarkhede: Also, the cookie will be set to the requests from localhost for running it locally and on staging environment as the cookie's SameSite attribute is set to `None`.

ankushdharkar: @ankurnarkhede The cookie will most likely be ignored because it's secure, so needs to be sent over https. Can you look into running localhost with https?

ankurnarkhede: For running UI locally, requests to staging environment will contain the auth cookie as the `samesite` is set to `None` and the secure attribute is not valid for the requests originating from localhost (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Attributes)

ankurnarkhede (changes requested):

CONTRIBUTING.md
ankushdharkar: @ankurnarkhede I checked on localhost, the cookie was being returned with `secure` attribute

ankurnarkhede: The secure flag is set but the requests to localhost will have the cookie added as localhost is exempted in case the secure flag is set. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Attributes

Am I missing anything? 🤔

ankushdharkar: @ankurnarkhede Oh. Interesting! I see what you are saying now. Thanks

I wonder why Chrome was complaining about it then. @whyDontI Did you also see that warning message?

Will have to check again, thanks Ankur!

ankurnarkhede (approved):
LGTM 👍 
@whyDontI

Issue #105, opened 2020-12-20
Return success message on successful setup of backend


swarajpure:
Currently the backend shows errors when the backend is not properly setup, but it doesn't show anything if the setup is successful.

It can be tested using `npm run test-config` command.

PR #109, opened 2020-12-25
Success message added on successful setup of backend

adnanraghib1308:
Success message added on successful setup of backend. This is the solution for the issue #105

### DIFF
--- a/local-test.js
+++ b/local-test.js
@@ -37,4 +37,4 @@ if (!firestoreData) {
   await docRef.delete()
 })()
 
-logger.info("SUCCESS!!! backend setup successful")
+logger.info('SUCCESS!!! backend setup successful')
--- a/local-test.js
+++ b/local-test.js
@@ -35,5 +35,5 @@ if (!firestoreData) {
     throw new Error('Problem with permission of read and write.\nCheck your firestore permissions')
   }
   await docRef.delete()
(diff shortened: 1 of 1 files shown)


local-test.js
```diff
@@ -35,3 +36,5 @@ if (!firestoreData) {
   }
   await docRef.delete()
 })()
+
+logger.info('SUCCESS!!! backend setup successful')
```
swarajpure, line 40: Are you sure this is on the right line? Have you tried it out?
Shouldn't it be inside the async function?

adnanraghib1308: Yeah right, it should be inside async function. Required changes done.

local-test.js
```diff
@@ -35,6 +35,5 @@ if (!firestoreData) {
     throw new Error('Problem with permission of read and write.\nCheck your firestore permissions')
   }
   await docRef.delete()
+  logger.info('SUCCESS!!! backend setup successful')
```
swarajpure, line 38: Great! Just one suggestion:
Can we have it in a form language? More formal and professional?

`Success! The backend was successfully set up.`

This is just one suggestion, you can have something of your own as well.

adnanraghib1308: Sure!
Changes are done. Thanks

swarajpure (approved):
Great! Thank you for the quick fix! 🎉

PR #110, opened 2020-12-28
Remove unneeded response headers

ankurnarkhede:
- Removed unneeded security headers from the library helmet
- Reference taken from: https://stackoverflow.com/a/60709460/8001821


whyDontI (approved):
Great 👍🏽 🔥

PR #113, opened 2021-01-03
Make `logger` and `config` accessible globally

ankurnarkhede:
- Made logger and config accessible globally to avoid importing it to every other file


swarajpure (approved):
Great idea!
Was not aware about making any variable global earlier.

Thank you! 😁

PR #114, opened 2021-01-03
Username cache

whyDontI:


services/userService.js
```diff
@@ -0,0 +1,39 @@
...
+const cacheUser = (user) => {
+  if (!user) {
+    return false
+  }
+  const userObject = {
+    github_id: user.github_id,
+    id: user.id,
+    username: user.username
+  }
+  set(user.id, userObject)
```
ankurnarkhede, line 15: What about the TTL?
Do we need to store this forever? We will need to modify this if the username is changed.

ankurnarkhede (changes requested):

services/cacheService.js
```diff
@@ -0,0 +1,10 @@
+const cache = require('memory-cache')
+
+const set = (key, value, time = 7200000) => {
+  return put(key, value, time)
+}
```
whyDontI, line 5: @ankurnarkhede Added default TTL as 2 hours

whyDontI (commented):
Added TTL

services/cacheService.js
```diff
@@ -0,0 +1,10 @@
+const cache = require('memory-cache')
+
+const set = (key, value, time = 7200000) => {
+  return cache.put(key, value, time)
+}
```
whyDontI, line 5: Set default TTL to 2 hours @ankurnarkhede

whyDontI (commented):
Added TTL

services/cacheService.js
```diff
@@ -0,0 +1,10 @@
+const cache = require('memory-cache')
+
+const set = (key, value, time = 7200000) => {
```
ankurnarkhede, line 3: We should set the `ttl` of different instances of data to be stored in cache as the config.
e.g.
```
cache: {
  ttl: {
      defaut: <defaultValue>,
      userCache: <userCacheTTL>
  }
}
```

services/userService.js
ankurnarkhede, line 15: Can we store the user object into the cache?
In that we, we can also utilise the same cache to get the user data during auth and will reduce the response time.

It would be great if you can create an issue for this cache implementation(if not created) and document all stuff into it.

ankurnarkhede, line 15: What's the TTL here?
According to the docs, this will stay in cache forever. We need to expose the cache set, get wrapper functions from the cache service which will have proper defaults and then use it here. 

Also, how will we manage the invalidation/updation of the data in the cache after user caches his profile info?

ankurnarkhede (changes requested):
- Added comments around implementation
- Pending integration and unit tests

PR #118, opened 2021-01-05
Fixed missing config imports

ankurnarkhede:
- Fixed missing config imports

### DIFF
--- a/local-test.js
+++ b/local-test.js
@@ -1,7 +1,6 @@
 const fs = require('fs')
 const Firestore = require('./utils/firestore.js')
 const config = require('config')
-const logger = require('./utils/logger')
 
 // check whether github credentials are not falsy
 if (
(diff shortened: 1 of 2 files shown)


local-test.js
```diff
@@ -35,5 +34,7 @@ if (!firestoreData) {
     throw new Error('Problem with permission of read and write.\nCheck your firestore permissions')
   }
   await docRef.delete()
-  logger.info('Success! The backend was successfully set up.')
+
+  // eslint-disable-next-line no-console
+  console.log('Success! The backend was successfully set up.')
```
ankurnarkhede, line 39: Using console log here as using logger doesn't seem to make sense here as this is a utility task, not an application file.
It also solves the issue of missing config import in logger.

swarajpure (approved):
Thank you! 🎉

PR #120, opened 2021-01-06
pagination capability for PRs API

aman7870:
I did the changes for getting the selected page value from user

### DIFF
--- a/controllers/pullRequestsController.js
+++ b/controllers/pullRequestsController.js
@@ -107,7 +107,6 @@ const getStalePRs = async (req, res) => {
  */
 const getOpenPRs = async (req, res) => {
   try {
-    
     const { data } = await githubService.fetchOpenPRs((req.query.page) || 1)
 
     if (data.total_count) {
--- a/controllers/pullRequestsController.js
+++ b/controllers/pullRequestsController.js
@@ -107,7 +107,12 @@ const getStalePRs = async (req, res) => {
  */
 const getOpenPRs = async (req, res) => {
   try {
-    const { data } = await githubService.fetchOpenPRs(req.query.page)
+    let pageNumber = 1
+
+    if (req.query.page > 1) {
+      pageNumber = req.query.page
+    }
+    const { data } = await githubService.fetchOpenPRs(pageNumber)
 
     if (data.total_count) {
       const allPRs = []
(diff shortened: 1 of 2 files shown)


swarajpure (changes requested):
Checks are failing

Please do `npm run lint-fix` to fix the errors and commit again

aman7870:
fixed the build errors

swarajpure (changes requested):
Please pass a default value of 1 in function call
Otherwise if the user simply calls `pullrequests/open` without `page` as query, a blank response will be returned.

aman7870:
fixed the page number to be 1 by default

controllers/pullRequestsController.js
```diff
@@ -107,7 +107,12 @@ const getStalePRs = async (req, res) => {
  */
 const getOpenPRs = async (req, res) => {
   try {
-    const { data } = await githubService.fetchOpenPRs()
+    let pageNumber = 1
+
+    if (req.query.page > 1) {
+      pageNumber = req.query.page
+    }
+    const { data } = await githubService.fetchOpenPRs(pageNumber)
```
swarajpure, line 115: There is a better and short way to pass default arguments

Please have a look here:
https://github.com/Real-Dev-Squad/website-backend/blob/d17ea626918d1424cc8ebd8181794c572e96ac24/models/users.js#L60

Also, please make sure to convert the argument to integer using `parseInt` as done in above function.

swarajpure (changes requested):

aman7870:
did the changes do check

controllers/pullRequestsController.js
```diff
@@ -107,7 +107,8 @@ const getStalePRs = async (req, res) => {
  */
 const getOpenPRs = async (req, res) => {
   try {
-    const { data } = await githubService.fetchOpenPRs()
+    const pageNumber = ((req.query.page) || 1)
+    const { data } = await githubService.fetchOpenPRs(pageNumber)
```
swarajpure, line 111: Can we simply pass it to the function than saving it in variable and then passing it?
`fetchOpenPRs(req.query.page || 1)` would be short and sweet

swarajpure (changes requested):
Just one final, simple change and we'll be able to merge it!

aman7870:
did the changes do check

swarajpure (approved):
Great job!
Thank you! 🎉

Issue #119, opened 2021-01-05
Create /tasks api for status-site


whyDontI:
# Find below schema
```
{
  type: 'dev' | 'group',
  links: [ array of related links ],
  endsOn,
  startedOn,
  status,
  ownerId,
  percentCompleted,
  dependsOn: [list of tasks that this task depends on],
  participants: [list of users participating in this tasks. For 'group' type only],
  completionAward: { gold: 3, bronze: 300 },
  lossRate: { gold: 1 }  // Loss per day of overshoot after deadline
}
```

# Create
- API contract
- POST /tasks
- GET /tasks
- PATCH /tasks/:id

PR #124, opened 2021-01-08
Add logic to fetch & create tasks

sumitd94:
API for issue #119


routes/index.js
```diff
@@ -5,6 +5,7 @@ app.use('/healthcheck', require('./healthCheck.js'))
 app.use('/auth', require('./auth.js'))
 app.use('/users', require('./users.js'))
 app.use('/members', require('./members.js'))
+app.use('/tasks', require('./tasks'))
```
ankushdharkar, line 8: GET /tasks
POST /task (singular?)

models/tasks.js
```diff
@@ -0,0 +1,36 @@
...
+    const taskInfo = await tasksModel.add(taskData)
+    return { taskid: taskInfo.id }
+  } catch (err) {
+    logger.error('Error in creating task', err)
+    throw err
+  }
+}
+
+async function fetchTasks () {
+  try {
```
swarajpure, line 15: Not using arrow function here?

models/tasks.js
```diff
@@ -0,0 +1,36 @@
...
+    return (function () {
+      const tasks = []
+      tasksSnapshot.forEach((task) => {
+        tasks.push({
+          id: task.id,
+          ...task.data()
+        })
+      })
+      return tasks
+    })()
```
swarajpure, line 26: Do we really need IIFE here?

swarajpure (changes requested):
Requesting some changes.

Thank you! 😄

routes/index.js
sumitd94: Yes, Makes sense. I have updated the PR.

controllers/tasksController.js
```diff
@@ -0,0 +1,37 @@
+const taskQuery = require('../models/tasks')
+
+const addNewTask = async (req, res) => {
+  try {
+    const task = await taskQuery.addTask(req.body)
+    return res.json({
+      message: 'Task created successfully!',
+      task: req.body,
+      id: task.taskid
```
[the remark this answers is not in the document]
sumitd94: Yes, Added the changes.

controllers/tasksController.js
```diff
@@ -0,0 +1,47 @@
+const taskQuery = require('../models/tasks')
```
ankurnarkhede, line 1: Rename the modal as simply 'tasks'!

controllers/tasksController.js
```diff
@@ -0,0 +1,47 @@
...
+  try {
+    const task = await taskQuery.addTask(req.body)
+    return res.json({
+      message: 'Task created successfully!',
+      task: req.body,
+      id: task.taskId
+    })
+  } catch (err) {
+    logger.error(`Error while creating new task: ${err}`)
+    return res.boom.serverUnavailable('Something went wrong please contact admin')
```
ankurnarkhede, line 19: Return 500.

controllers/tasksController.js
```diff
@@ -0,0 +1,47 @@
...
+    if (allTasks.length > 0) {
+      return res.status(200).json({
+        message: 'Tasks returned successfully!',
+        tasks: allTasks
+      })
+    }
+    return res.boom.notFound('No tasks found')
+  } catch (err) {
+    logger.error(`Error while fetching tasks ${err}`)
+    return res.boom.serverUnavailable('Something went wrong please contact admin')
```
ankurnarkhede, line 40: 500.

routes/index.js
```diff
@@ -5,6 +5,7 @@ app.use('/healthcheck', require('./healthCheck.js'))
 app.use('/auth', require('./auth.js'))
 app.use('/users', require('./users.js'))
 app.use('/members', require('./members.js'))
+app.use(require('./tasks'))
```
swarajpure, line 8: I understand what you're doing here, but I'm not very sure about it 😅 
@ankurnarkhede @whyDontI What do you suggest?

routes/tasks.js
```diff
@@ -0,0 +1,38 @@
...
+const tasksController = require('../controllers/tasksController')
+
+/**
+ * @swagger
+ * /tasks/:
+ *   summary: Used to get all the tasks
+ *   responses:
+ *     200:
+ *       description: returns tasks
+ *     404:
```
ankurnarkhede, line 12: Add response schema to error response.

routes/tasks.js
```diff
@@ -0,0 +1,38 @@
+const express = require('express')
+const router = express.Router()
+const tasksController = require('../controllers/tasksController')
+
+/**
+ * @swagger
+ * /tasks/:
```
swarajpure, line 7: No need of the trailing slash
Should be just `/tasks:`

routes/tasks.js
```diff
@@ -0,0 +1,38 @@
...
+ *   summary: Used to get all the tasks
+ *   responses:
+ *     200:
+ *       description: returns tasks
```
swarajpure, line 11: Please add schema for task object in `utils/swaggerDefinition.js` and include that schema in 200 response

ankurnarkhede (changes requested):

routes/tasks.js
```diff
@@ -0,0 +1,38 @@
...
+ *   summary: Used to create new task
+ *   responses:
+ *     200:
+ *       description: returns newly created task
+ *     503:
+ *       description: serverUnavailable
+ *       content:
+ *         application/json:
+ */
+router.post('/task', tasksController.addNewTask)
```
ankurnarkhede, line 36: Shouldn't this be `/tasks`?

ankurnarkhede (changes requested):

routes/tasks.js
Rucha1499, line 7: I think we don't need an extra `/` here. Can you please changes it to `/tasks` ?

swarajpure (changes requested):
Please take a pull from develop.
Suggesting some changes 😅

routes/tasks.js
```diff
@@ -0,0 +1,38 @@
...
+ *       description: serverUnavailable
+ *       content:
+ *         application/json:
+ */
+
+router.get('/tasks', tasksController.fetchTasks)
+
+/**
+ * @swagger
+ * /tasks/:
```
Rucha1499, line 26: An extra `/` here also. 😅

Rucha1499 (changes requested):

routes/tasks.js
```diff
@@ -0,0 +1,38 @@
...
+ *     200:
+ *       description: returns tasks
+ *     404:
+ *       description : no tasks found
+ *       content:
+ *         application/json:
+ *     503:
+ *       description: serverUnavailable
+ *       content:
+ *         application/json:
```
Rucha1499, line 19: - for 404 we need to write 
`application/json:
        schema:
            $ref: '#/components/schemas/errors/notFound'`

- for 503
 `application/json:
         schema:
              $ref: '#/components/schemas/errors/serverUnavailable'`

Please add them wherever missing. Thank you! 😃

controllers/tasksController.js
sumitd94: Did you mean to rename `const taskQuery` to `cont tasks` ?

routes/tasks.js
sumitd94: Got it, Thanks.

routes/index.js
sumitd94: I will make it easy for you ;)

swarajpure (changes requested):
Since you've made changes in `routes/index.js`, the GET call will have an endpoint as `/tasks/tasks` instead of just `/tasks`
Please see to it
Thank you! 😁

routes/index.js
```diff
@@ -5,6 +5,8 @@ app.use('/healthcheck', require('./healthCheck.js'))
 app.use('/auth', require('./auth.js'))
 app.use('/users', require('./users.js'))
 app.use('/members', require('./members.js'))
+app.use('/tasks', require('./tasks'))
+app.use('/roadmap-site', require('./roadmap-site/index'))
```
swarajpure, line 9: Please remove this in order to fix the failing checks
We're not using `roadmap-site` anymore

swarajpure (changes requested):

swarajpure (dismissed):
Looks good to me.

Thank you! 😁

Rucha1499 (commented):
Thanks for quick changes! 🎉 👍

Rucha1499 (approved):

controllers/tasksController.js
ankurnarkhede: yes

routes/tasks.js
```diff
@@ -0,0 +1,52 @@
...
+ *           schema:
+ *             $ref: '#/components/schemas/errors/badImplementation'
+ */
+
+router.get('/', tasksController.fetchTasks)
+
+/**
+ * @swagger
+ * /tasks:
```
ankurnarkhede, line 34: Where is the HTTP method defined here?

routes/tasks.js
```diff
@@ -0,0 +1,52 @@
+const express = require('express')
+const router = express.Router()
+const tasksController = require('../controllers/tasksController')
+
+/**
+ * @swagger
+ * /tasks:
```
ankurnarkhede, line 7: HTTP method missing.

controllers/tasksController.js
```diff
@@ -0,0 +1,47 @@
...
+const fetchTasks = async (req, res) => {
+  try {
+    const allTasks = await taskQuery.fetchTasks()
+    if (allTasks.length > 0) {
+      return res.status(200).json({
+        message: 'Tasks returned successfully!',
+        tasks: allTasks
+      })
+    }
+    return res.boom.notFound('No tasks found')
```
ankurnarkhede, line 37: Ideally this should be a 200 response.
Its a success response with no tasks. 🤔

controllers/tasksController.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param req {Object} - Express request object
+ * @param req.body {Object} - Task object
+ * @param res {Object} - Express response object
+ */
+const addNewTask = async (req, res) => {
+  try {
+    const task = await taskQuery.addTask(req.body)
+    return res.json({
+      message: 'Task created successfully!',
+      task: req.body,
```
ankurnarkhede, line 14: We shouldn't return the request body here. All fields from the request body are not to be added to the DB.
We should return the data stored in DB by the model function.
Can you check what data is obtained at https://github.com/Real-Dev-Squad/website-backend/pull/124/files#diff-aa74c20d1c66caed73c8f28f809dff80428137d5f605696c02ac8494c02bd46fR12 and utilise it?

controllers/tasksController.js
```diff
@@ -0,0 +1,47 @@
+const taskQuery = require('../models/tasks')
+/**
+ * Creates new task
+ *
+ * @param req {Object} - Express request object
+ * @param req.body {Object} - Task object
+ * @param res {Object} - Express response object
+ */
+const addNewTask = async (req, res) => {
```
ankurnarkhede, line 9: Isn't there any validation added for the request data?
Any user can thus add any long request body and it will just land in the DB!

utils/swaggerDefinition.js
```diff
@@ -53,6 +53,43 @@ const swaggerOptions = {
...
         },
+        tasks: {
+          type: 'object',
+          properties: {
+            type: {
+              type: 'string'
+            },
+            links: {
+              type: 'array',
+              items: []
```
ankurnarkhede, line 64: Can you define the data to be required in the items array?

utils/swaggerDefinition.js
```diff
@@ -53,6 +53,43 @@ const swaggerOptions = {
...
+            },
+            ownerId: {
+              type: 'string'
+            },
+            percentCompleted: {
+              type: 'number'
+            },
+            dependsOn: {
+              type: 'array',
+              items: []
```
ankurnarkhede, line 83: Also define data required in dependsOn array.

utils/swaggerDefinition.js
```diff
@@ -53,6 +53,43 @@ const swaggerOptions = {
...
+            },
+            completionAward: {
+              type: 'object'
```
ankurnarkhede, line 86: We should be defining the data in these objects too as those keys won't be changing anytime.

controllers/tasksController.js
ankurnarkhede: Also, can you add tests for those validators and other functions defined in PR?

ankurnarkhede (changes requested):
Added a few comments.

controllers/tasksController.js
sumitd94: So , if there are no tasks .. we still return a success response, but the `tasks` in the response message will be an empty object , is that what you are trying to tell ?

utils/swaggerDefinition.js
sumitd94: Like this `items: [link1, link2]` ?

ankurnarkhede: Also data types for those.
Refer:
- https://swagger.io/docs/specification/2-0/adding-examples/
- https://stackoverflow.com/questions/35955243/following-swagger-specifications-how-can-i-define-json-of-nested-objects-to-yam

controllers/tasksController.js
ankurnarkhede: Yes, because its actually not a client error, right?

controllers/tasksController.js
```diff
@@ -0,0 +1,47 @@
...
+ * Fetches all the tasks
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const fetchTasks = async (req, res) => {
+  try {
+    const allTasks = await taskQuery.fetchTasks()
+    if (allTasks.length > 0) {
+      return res.status(200).json({
```
ankurnarkhede, line 32: No need to specify the status code explicitly unless its !== 200.

ankurnarkhede (changes requested):

utils/swaggerDefinition.js
sumitd94: What to do if we do not know exactly the object keys for this ? like for `completionAward` we can reward the user with only `gold` or with `gold & silver` or `gold, silver, bronze`
in this case should I define the properties as `gold, silver & bronze` ?

controllers/tasksController.js
sumitd94: I will add the validations, regarding tests I need to check .. let me get these validations in place and then lets check on this tests

controllers/tasksController.js
sumitd94: Ok Makes sense, Thanks.

utils/swaggerDefinition.js
ankurnarkhede: Yes

models/tasks.js
sumitd94: Done

models/tasks.js
sumitd94: Thanks , I have improved this.

controllers/tasksController.js
sumitd94: Thanks, I have updated the logic accordingly.

utils/swaggerDefinition.js
sumitd94: Ok Thanks.

routes/tasks.js
```diff
@@ -0,0 +1,77 @@
...
+ *           schema:
+ *             $ref: '#/components/schemas/errors/badImplementation'
+ */
+router.post('/', tasksController.addNewTask)
+
+/**
+ * @swagger
+ * /tasks:
+ *  patch:
+ *   summary: Used to update task details
```
Rucha1499, line 58: - For the patch route we are sending the data which needs to be updated right ? So, we need to mention it here in swagger docs. Please add `requestBody` after summary. 
- You can refer this: https://swagger.io/docs/specification/basic-structure/

Rest everything looks good to me! Thank you. 😃

Rucha1499 (changes requested):

routes/tasks.js
sumitd94: Done, Please check.

Rucha1499 (approved):
Looks good to me. Thank you!! @sumitd94 🎉 😄

controllers/tasksController.js
ankurnarkhede: This variable name hasn't been changed. Its a reference to the model, not to a query, hence we should not name it as tasksQuery.

controllers/tasksController.js
ankurnarkhede: Still the same. Comment not addressed.

controllers/tasksController.js
ankurnarkhede: It should be 500 error, not 503.

ankurnarkhede (changes requested):
Few comments aren't addressed.
Tests missing.

controllers/tasksController.js
ankurnarkhede: It should return an empty tasks array in this case with 200 response.
The client has requested the correct resource in this case.

ankurnarkhede (changes requested):

sumitd94:
> Few comments aren't addressed.
> Tests missing.

https://github.com/Real-Dev-Squad/website-backend/pull/124/commits/ba355b7840d794d8aea204995d2bd62a1ecdba6b

controllers/tasksController.js
```diff
@@ -0,0 +1,66 @@
...
+    const allTasks = await tasks.fetchTasks()
+    return res.json({
+      message: 'Tasks returned successfully!',
+      tasks: allTasks.length > 0 ? allTasks : []
+    })
+  } catch (err) {
+    logger.error(`Error while fetching tasks ${err}`)
+    return res.boom.badImplementation('An internal server error occurred')
+  }
+}
```
Rucha1499, line 39: Are we not returning the error `404` if no tasks are found ? I can see it mentioned in the swagger docs below! Please add it here as well.

image


Thank you!

Rucha1499 (changes requested):
Nice work! 🎉 
Suggesting small change! 😅

routes/tasks.js
```diff
@@ -0,0 +1,90 @@
...
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/components/schemas/tasks'
+ *     404:
+ *       description : no tasks found
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/components/schemas/errors/notFound'
```
Rucha1499, line 23: I think we don't need it here. We are not returning `404` error response for this request. Can you please remove it?

Thank you! 😄

Rucha1499 (changes requested):

controllers/tasksController.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+const updateTask = async (req, res) => {
+  try {
+    const task = await tasks.fetchTask(req.params.id)
+    if (!task.taskData) {
+      return res.boom.notFound('Task not found')
+    }
+
+    await tasks.updateTask(req.body, req.params.id)
+    return res.status(204).send()
```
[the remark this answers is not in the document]
sumitd94: If the response is 204, we will be assuming the task has been updated. If there are any errors, we are returning the respective error responses .. so that can be handled at frontend.

ankurnarkhede (approved):
LGTM 👍 
Created a task for added tests to be taken up later: https://github.com/Real-Dev-Squad/website-backend/issues/135

PR #133, opened 2021-01-12
Ref(middleware/user): Few extra form fields added

SagarBajpai:


middlewares/validators/user.js
```diff
@@ -2,9 +2,12 @@ const joi = require('joi')
 
 const createUser = async (req, res, next) => {
   const schema = joi.object().keys({
+    phone_number: joi.string().optional(),
```
Rucha1499, line 5: The phone number field is mentioned as `phone` in the API contract of users. You can have a look at this : 
https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/users#users

So, what do you think should we changes it to `phone` here instead of `phone_number`, or vice-versa?
Thank you! 😃

ankurnarkhede: @SagarBajpai , can we change this to `phone` as per the contracts?

ankurnarkhede: @SagarBajpai , can we change this to phone as per the contracts?

ankurnarkhede (changes requested):

middlewares/validators/user.js
sagarokcredit: DOne

swarajpure (changes requested):
There have been changes in ESLint rules, please revert them. Also please disable prettier for this project and just use the ESLint VSCode extension. It'll take care of linting + formatting

Thank you! 😁

swarajpure (approved):
Great! Thank you! 🎉

Issue #32, opened 2020-09-06
Create a test suite


ankurnarkhede:
Create a test suite using the mocha APIs for the project to facilitate the following:
- Add firestore config for test or use any firestore mocking libraries(if available), and thus remove all model function mocks ✅  
  - https://firebase.google.com/docs/emulator-suite 
  - https://stackoverflow.com/questions/46563885/running-firestore-local-e-g-for-testing
- Define custom test specific config.
  - Be able to define custom config for tests to achieve https://github.com/Real-Dev-Squad/website-backend/blob/develop/middlewares/authenticate.js#L14
- Reduce redundancy in each test files ✅ 
  - Use Mocha API https://mochajs.org/api/mocha to start the test suite and define globals
  - Start the server only once rather than passing the server.js to chai HTTP module
- Mocking XHR calls ✅ 
- Mocking function calls ✅ 
- Test coverage in the repo ✅ 

====================================================================================
## Current issues in test suite:
- `firebase emulators:start` ([code](https://github.com/Real-Dev-Squad/website-backend/blob/develop/scripts/tests/testIntegration.sh#L11)) is a blocking process and does not get terminates after the test process gets finished. ✅ 
- Create an interface/utility to flush the DB. It can be used in the mocha tests as per required.
- Define custom test specific config.
  - Be able to define custom config for tests to achieve https://github.com/Real-Dev-Squad/website-backend/blob/develop/middlewares/authenticate.js#L14

ankurnarkhede:
Reopening the issue. More work required on the test suite. Updating the description.

Cartmanishere:
@ankurnarkhede 

> Create an interface/utility to flush the DB. It can be used in the mocha tests as per required.

The library for firestore we are using (`@google-cloud/firestore`) does not have an operation for clearing all data in firestore. If we had used a library like `@firebase` then, we would have access to interface like `clearFirestoreData`.

In the current package, we can do something like fetching and deleting all the collections (and subcollections) in the firestore in appropriate test fixture (`beforeEach` or `afterAll`). 

I think this is a bit of an overkill for a test suite. Instead, we can use the rest following REST endpoint for the firestore emulator. 

`http://localhost:8080/emulator/v1/projects/<project-name>/databases/(default)/documents`

We can clear the documents after the tests are run so the state of one test block would not affect the other.

shubham-y:
[edit] Sorry, I pushed the wrong button!

PR #134, opened 2021-01-16
Fix firestore emulator issue

Cartmanishere:
As per this issue https://github.com/Real-Dev-Squad/website-backend/issues/32, the firestore emulator would keep running even after the tests are done running.

This change fixes --
- Firestore emulator is stopped after the tests are done.
- Remove unnecessary teardown step of erasing data.

We don't need to erase data because firestore emulator does not persist it by default.

If we want to persist data, we have to use `emulators:import` and `emulators:export`. 

Ref:
- [Install, configure and integrate Local Emulator Suite](https://firebase.google.com/docs/emulator-suite/install_and_configure)
- [Test your Cloud Firestore Security Rules](https://firebase.google.com/docs/firestore/security/test-rules-emulator)

### DIFF
--- a/scripts/tests/testIntegration.sh
+++ b/scripts/tests/testIntegration.sh
@@ -3,18 +3,5 @@
 # set 'test' environment
 export NODE_ENV='test'
 
-# Set environment variable for firestore emulator host
-# @todo: Get port value from '../../firebase.json'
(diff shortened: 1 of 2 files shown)


ankurnarkhede (approved):
LGTM 👍

Issue #141, opened 2021-01-24
Authorize the user for post and update of the tasks


Harshi7016:
Add a middleware that checks for the user who is logged in has the role as an app owner for post or update of the new task.

PR #142, opened 2021-01-24
PR for Authorize the user for post and update of the tasks #141

Harshi7016:


middlewares/authorizeOwner.js
```diff
@@ -0,0 +1,35 @@
+const usersController = require('../controllers/usersController')
+/**
+ * Middleware to validate the authorized routes to fetch open PRs
```
sumitd94, line 3: I think this should be something like this
`Middleware to validate the authorized routes to be able to create & Update tasks`

middlewares/authorizeOwner.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ * @param next {Function} - Express middleware function
+ * @return {Object} - Returns unauthorized user if the role is not assigned
+ */
+module.exports = async (req, res, next) => {
+  try {
+    // get user data from `req.userData` for further use
+    const accountOwners = await usersController.getAccountOwners()
+    if (!req.userData.incompleteUserDetails) {
```
sumitd94, line 22: NIT: you can destructure the `incompleteUserDetails & username` from `req.userData` by adding a new line of code above this. something like this
```javascript
const { incompleteUserDetails , username } = req.userData
```

mockdata/appOwners.json
```diff
@@ -0,0 +1,63 @@
...
+        "img": "test",
+        "github_id": "whydonti",
+        "linkedin_id": "test",
+        "twitter_id": "test",
+        "instagram_id": "test",
+        "site": "test",
+        "isMember": "boolean",
+        "github_display_name": "Nikhil Bhandarkar",
+        "incompleteUserDetails": false
+    }
```
sumitd94, line 62: Why these many fields when you are using only `username` ? Let's keep this data simple by including `first_name, last_name, username`
Rest all the fields are not required.
Also please add other App owners
`kratika, ankur, swaraj` (these are their usernames)

sumitd94 (changes requested):
Few changes requested, other than that things looks good. Great work :)

PR #154, opened 2021-02-01
Allows empty `website` field

### LINKED ISSUE #153: Make `website` field optional from users
The [frontend](https://my.realdevsquad.com/signup) says that the `website` is optional but if we send blank in website field, the backend sends that `website` cannot be empty and hence form cannot be submitted

swarajpure:
Fixes #153

### DIFF
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -52,7 +52,7 @@ const updateUser = async (req, res, next) => {
     site: joi.string().optional(),
     isMember: joi.boolean().optional(),
     tokens: joi.object().optional(),
-    website: joi.string().optional()
+    website: joi.string().optional().allow('')
   })
 
   try {


middlewares/validators/user.js
```diff
@@ -52,7 +52,7 @@ const updateUser = async (req, res, next) => {
     site: joi.string().optional(),
     isMember: joi.boolean().optional(),
     tokens: joi.object().optional(),
-    website: joi.string().optional()
+    website: joi.string().optional().allow('')
```
ankurnarkhede, line 55: Why is this change being made?
The website name shouldn't be an empty string, right?

ankurnarkhede (changes requested):

middlewares/validators/user.js
swarajpure: Yes. Currently the signup form has `website` as optional and when the user leaves it blank, the backend sends 400 response saying `website cannot be empty` 

Source: https://github.com/sideway/joi/blob/v10.6.0/API.md#string---inherits-from-any
image

ankurnarkhede: Yeah, but this doesn't mean that we should explicitly allow the empty strings. This is an issue from the UI. The UI shouldn't be sending empty strings for website value.

swarajpure:
Handling on the frontend

PR #156, opened 2021-02-02
returning `phone` and `email` only for users/self route when query `private` is true

### LINKED ISSUE #155: Send `phone` and `email` only for `users/self` route if query `private` is true
Since the signup form is now taking phone number and email address, we should not send them as response for every API. No user should be able to see other user's phone number and email address.

No route other than `GET users/self` should return `phone` and `email` and that too only if query `private` is set to true

swarajpure:
Fixes #155 

According to the API contract: https://github.com/Real-Dev-Squad/website-api-contracts/pull/23

### DIFF
--- a/test/integration/usersController.test.js
+++ b/test/integration/usersController.test.js
@@ -33,11 +33,14 @@ describe('Users', function () {
         .send({
           first_name: 'Nikhil',
           last_name: 'Bhandarkar',
+          username: 'nikhil',
           yoe: 0,
           img: './img.png',
           github_id: 'whydonti',
           linkedin_id: 'nikhil-bhandarkar',
-          twitter_id: 'whatifi'
+          twitter_id: 'whatifi',
+          phone: '1234567890',
+          email: 'abc@gmail.com'
         })
         .end((err, res) => {
           if (err) { return done() }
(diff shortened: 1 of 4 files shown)


ankurnarkhede (changes requested):
@swarajpure , as per the conversation with @ankushdharkar , we will go forward with the private query param.
Can you add tests for this change?
We can merge his post that.

swarajpure:
> 
> 
> @swarajpure , as per the conversation with @ankushdharkar , we will go forward with the private query param.
> Can you add tests for this change?
> We can merge his post that.

Tests added!

test/integration/usersController.test.js
```diff
@@ -106,6 +111,65 @@ describe('Users', function () {
...
+            message: 'Unauthenticated User'
+          })
+
+          return done()
+        })
+    })
+  })
+
+  describe('GET /users/self?private=true', function () {
+    it('Should return the logged user\'s details with phone and email', function (done) {
```
ankurnarkhede, line 161: @swarajpure, can you add this test in the earlier describe block? We do not need a separate describe block for this. Its the same route.

ankurnarkhede (changes requested):

ankurnarkhede (approved):

Issue #148, opened 2021-01-30
Add integration/unit tests for contribution APIs


ankurnarkhede:
- Add integration/unit tests for contribution APIs

PR #159, opened 2021-02-05
Test for contribution/user

prakashchoudhary07:
#148 Added Test for contribution/user


test/integration/contributionsController.test.js
```diff
@@ -0,0 +1,155 @@
...
+
+describe('Contributions route', function () {
+  describe('Create User', function () {
+    before(async function () {
+      const userId = await addUser()
+      jwt = authService.generateAuthToken({ userId })
+    })
+
+    afterEach(function () {
+      sinon.restore()
```
ankurnarkhede, line 27: This is not required as you are not mocking any function calls.

test/integration/contributionsController.test.js
```diff
@@ -0,0 +1,155 @@
...
+    })
+
+    describe('POST /users - create new user', function () {
```
ankurnarkhede, line 30: Why is the test for `POST /users` added here? 🤔

test/integration/contributionsController.test.js
```diff
@@ -0,0 +1,155 @@
...
+const app = require('../../server')
+
+chai.use(chaiHttp)
+
+// Import fixtures
+const githubPRInfo = require('../fixtures/contributions/githubPRInfo')()
+
+let jwt
+
+describe('Contributions route', function () {
```
ankurnarkhede, line 19: Name it as simply Contributions.

test/integration/contributionsController.test.js
```diff
@@ -0,0 +1,155 @@
...
+        .post('/tasks')
+        .send({
+          title: 'Task for contribution (GET contributions/:username API)',
+          purpose: 'To Test mocha',
+          featureUrl: '<testUrl>',
+          type: 'Dev | Group',
+          links: [
+            'https://github.com/Real-Dev-Squad/website-backend/pull/145'
+          ],
+          endsOn: '<unix timestamp>',
```
ankurnarkhede, line 70: Can you add the actual values here?

test/fixtures/contributions/githubPRInfo.js
```diff
@@ -0,0 +1,366 @@
+/**
+ * USer PR info for GitHub response
+ * Multiple responses can be added to the array if required
```
ankurnarkhede, line 3: Which array?
I see the function is returning an Object.

test/integration/contributionsController.test.js
```diff
@@ -0,0 +1,155 @@
...
+            return done()
+          })
+      })
+    })
+  })
+  describe('POST /tasks - creates a new task', function () {
+    it('Should return success response after adding the task (isNoteWorthy = true)', function (done) {
+      chai
+        .request(app)
+        .post('/tasks')
```
ankurnarkhede, line 61: This isn't a contributions route. The tests for the tasks should be added in a separate file for tasksController.

If the motive here is to use the already added data of tasks into the contributions test, then you can simply add the data using the `tasks.updateTasks()` function ([code](https://github.com/Real-Dev-Squad/website-backend/blob/develop/models/tasks.js#L11)) in the test itself, or in before blocks and then test the response from contributions route.
This is because our actual aim here is to test the contributions routes, not the tasks route.

test/integration/contributionsController.test.js
```diff
@@ -0,0 +1,155 @@
...
+          expect(res.body).to.be.a('object')
+          expect(res.body.message).to.equal('Task created successfully!')
+          expect(res.body.id).to.be.a('string')
+          expect(res.body.task).to.be.a('object')
+          return done()
+        })
+    })
+  })
+  describe('GET /contributions/{username}', function () {
+    it('Should get all the contributions of the user', function (done) {
```
ankurnarkhede, line 130: https://github.com/Real-Dev-Squad/website-backend/blob/develop/routes/contributions.js#L19 shows that this route also has 404, 500 responses. Can you also add tests for that?

ankurnarkhede, line 130: Also, add tests for no tasks, no PRs, etc.

ankurnarkhede (changes requested):

test/integration/contributionsController.test.js
prakashchoudhary07: Github api calls are mocked

test/integration/contributionsController.test.js
prakashchoudhary07: Removed it and populated Db using before()

ankurnarkhede (approved):

PR #163, opened 2021-02-08
Badge GET api

whyDontI:
## Added
- [x] GET /badges

## API Contracts
https://github.com/Real-Dev-Squad/website-api-contracts/pull/29


Cartmanishere (commented):
LGTM

Rucha1499 (commented):
Great work!! @whyDontI 🎉 😄 

Can we generate the API Schema for this route?

Rucha1499 (approved):
Looks great!! 🎉

test/integration/badgeController.test.js
```diff
@@ -0,0 +1,26 @@
...
+          expect(res).to.have.status(200)
+          expect(res.body).to.be.a('object')
+          expect(res.body.message).to.equal('Badges returned successfully!')
+          expect(res.body.badges).to.be.a('array')
+
+          return done()
+        })
+    })
+  })
+})
```
swarajpure, line 26: Where are we adding the badges in the database to `GET` them? 🤔
How will this test case pass on CI? 🤔

swarajpure (commented):
Having a doubt in the test case 😅

test/integration/badgeController.test.js
whyDontI: Ankush is going to add badges manually in database until we meet the authorization in place.

swarajpure: > 
> 
> Ankush is going to add badges manually in database until we get the authorization in place.

I guess the CI also uses firebase emulators that's why we install it on every commit.
I meant the CI doesn't use the production db, right?
Feel free to correct me 😄

whyDontI: Yes, seems like CI uses emulator, check test.yml file in .git/workflow
CI fetching and adding data in prod DB won't be a good idea

swarajpure: That's what I'm saying 😅 
The prod DB will have the badges details
but the firebase emulator wont and hence the test wouldn't pass 😅 

Let me know if I'm wrong 😄

whyDontI: Oh, that's the convention, even if the GET Endpoint returns an empty array, the response code will still be 200, and the request is considered fulfilled.

swarajpure: Cool

swarajpure (approved):
Looks good! 👌🏻

PR #165, opened 2021-02-08
Adds tests for GET /members API

swarajpure:
Fixes #164

PR #167, opened 2021-02-09
added response in case of no member found

### LINKED ISSUE #166: Return difference response when no members are found in the database
Currently our `membersController` returns the same response when there are members or no member which is not ideal.

swarajpure:
Fixes #166

### DIFF
--- a/controllers/membersController.js
+++ b/controllers/membersController.js
@@ -11,15 +11,9 @@ const getMembers = async (req, res) => {
   try {
     const allMembers = await memberQuery.fetchMembers()
 
-    if (allMembers.length) {
-      return res.json({
-        message: 'Members returned successfully!',
-        members: allMembers
-      })
-    }
-
     return res.json({
-      message: 'No member found!'
+      message: allMembers.length ? 'Members returned successfully!' : 'No member found',
+      members: allMembers
     })
   } catch (error) {
     logger.error(`Error while fetching all members: ${error}`)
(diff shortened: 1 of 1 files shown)


swarajpure:
To be merged before #165

harshith-venkatesh (dismissed):
Good Catch on edge cases @swarajpure and @ankurnarkhede

controllers/membersController.js
```diff
@@ -11,9 +11,15 @@ const getMembers = async (req, res) => {
   try {
     const allMembers = await memberQuery.fetchMembers()
 
+    if (allMembers.length) {
```
ankurnarkhede, line 14: @swarajpure , any reason we want these separate messages?
We can rely on the array length at the UI, right?

controllers/membersController.js
```diff
@@ -11,9 +11,15 @@ const getMembers = async (req, res) => {
...
+      return res.json({
```
ankurnarkhede, line 15: In case if the newer message is required, we can simply return
```js
    return res.json({
      message: allMembers.length ? 'Members returned successfully!' : 'No member found!',
      members: allMembers
    })
```

ankurnarkhede (changes requested):

controllers/membersController.js
swarajpure: Done. Please check

controllers/membersController.js
swarajpure: I guess it would be better to have a different message in case of no member found. Personally, it doesn't make sense to me where the `members` array is empty but still the message says `Members returned successfully!`, since there are no members returned. Again, my opinion. Let me know what you think about it 🙂

swarajpure (commented):
Addressing @ankurnarkhede's comments

ankurnarkhede (approved):

PR #171, opened 2021-02-10
user-POST Route removed

JatinAgrawal94:
POST request route for 'user' was removed.


swarajpure (changes requested):
Please remove your firestore key json file
Anyone can access your db with this file
You're not supposed to commit it

Please revert the change from package-lock

Also, you're supposed to remove the route and not just comment it out.

Thank you!

JatinAgrawal94:
Should I retain the comments regarding the Post Route or remove it ?

Issue #160, opened 2021-02-06
Feature: Display the RDS developers who are not assigned with issues


Harshi7016:
1. Create a page to display list of idle users of RDS.
2. Create a route, controllers for the idleUser Function
3.. Firebase call to efficiently fetch the list of users.

whyDontI:
The idle members API returns all the members who are not in the `participant's` array.
But after updated `API Contract,` we have another field as `assignee`.
We need to exclude members who are assigned to some issues, from idle members response.

I hope I am able to explain this 😁 , if not please feel free to reach out to me 😃

harshith-venkatesh:
Thanks for the update @whyDontI , will work on the changes

whyDontI:
@harshith-venkatesh
As of now, idle members API returns all the users which have ACTIVE tasks assigned to them.
`"status": "active | assigned | unAssigned | blocked | completed"`

I think it should return all the users which have tasks with status anything other than `UNASSIGNED`.

Refer [this](https://github.com/Real-Dev-Squad/website-backend/blob/79afa71a73dcbebc372cb6e734b62503c785447f/models/tasks.js#L64) line

PR #174, opened 2021-02-10
feature: GET inactive/idle members of RDS

harshith-venkatesh:
Please provide your feedback and guide me to improve this code


Harshi7016:
https://github.com/Real-Dev-Squad/website-backend/issues/160

controllers/membersController.js
```diff
@@ -21,6 +22,34 @@ const getMembers = async (req, res) => {
...
+ * Fetches the data about our inactive/idle members
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const getIdleMembers = async (req, res) => {
+  try {
+    let allMemberUsernames = await memberQuery.fetchMemberUsernames()
+    allMemberUsernames = allMemberUsernames.map(member => member.username)
```
sumitd94, line 35: I think this is not needed.
Also change `allMemberUsernames` to `const`

controllers/membersController.js
```diff
@@ -21,6 +22,34 @@ const getMembers = async (req, res) => {
...
+
+    let taskParticipants = await tasks.fetchActiveTaskMembers()
+    taskParticipants = [...new Set(taskParticipants)]
+
+    const idleMemberUserNames = allMemberUsernames.filter(member => taskParticipants.indexOf(member) === -1)
```
sumitd94, line 40: I think instead of destruturing the `sets` back to an array, you can simply make the full of `Sets` like this below

```javascript
    taskParticipants = new Set(taskParticipants)
    const idleMemberUserNames = allMemberUsernames.filter(member => !taskParticipants.has(member))
```

sumitd94 (changes requested):
Requesting few changes, after that thing looks good.

routes/members.js
```diff
@@ -26,4 +26,28 @@ const membersController = require('../controllers/membersController')
 
 router.get('/', membersController.getMembers)
 
+/**
+ * @swagger
+ * /members:
```
swarajpure, line 31: Line 31 should be `/members/idle`

routes/members.js
```diff
@@ -26,4 +26,28 @@ const membersController = require('../controllers/membersController')
...
+ *   get:
+ *     summary: Gets details of inactive/idle members of the Real Dev Squad
+ *     tags:
+ *       - IdleMembers
```
swarajpure, line 35: `tags` should be `Members` only as this route is a part of `Members` routes

models/members.js
```diff
@@ -33,6 +33,31 @@ const fetchMembers = async () => {
   }
 }
 
+/**
+ * Fetches the data about our member usernames
```
swarajpure, line 37: `data about our member usernames`..... 🤔 

What about something like `Returns usernames of all the members in the database`?

controllers/membersController.js
```diff
@@ -21,6 +22,34 @@ const getMembers = async (req, res) => {
   }
 }
 
+/**
+ * Fetches the data about our inactive/idle members
```
swarajpure, line 26: `Returns the usernames of inactive/idle members` seems better to me personally, what do you think?

swarajpure (changes requested):
Some minor queries. Please have a look

Thank you 🎉

ankurnarkhede:
Hey @harshith-venkatesh , any update here?

harshith-venkatesh:
https://github.com/Real-Dev-Squad/website-backend/issues/195

models/tasks.js
```diff
@@ -49,6 +49,28 @@ const fetchTasks = async () => {
   }
 }
 
+/**
+ * Fetch all participants whose task status is active/pending/blocked
```
sumitd94, line 53: NIT: I think this should be only `active` right?

sumitd94 (changes requested):
One minor change, rest all looks good.

sumitd94 (approved):

test/integration/membersController.test.js
```diff
@@ -51,4 +51,42 @@ describe('Members', function () {
...
+          expect(res).to.have.status(200)
+          expect(res.body).to.be.a('object')
+          expect(res.body.message).to.equal('No idle member found')
+          expect(res.body.idleMemberUserNames).to.eql([])
+
+          return done()
+        })
+    })
+
+    it('Get all the idle members in the database', function (done) {
```
ankurnarkhede, line 75: @harshith-venkatesh , we should also have a test when we have tasks assigned to a user and then check that the user is not idle.

ankurnarkhede (changes requested):

test/integration/membersController.test.js
ankurnarkhede: Because, the current tests are written for the cases where there are no tasks.

models/members.js
```diff
@@ -35,6 +35,31 @@ const fetchMembers = async () => {
   }
 }
 
+/**
+ * Fetches usernames of all the members in the database
+ * @return {Promise<userModel|Array>}
+ */
+
+const fetchMemberUsernames = async () => {
```
ankurnarkhede, line 43: Rather than creating a new function for just getting the usernames, can we use this function https://github.com/Real-Dev-Squad/website-backend/blob/develop/models/members.js#L14 and then filter out usernames from it?

models/tasks.js
```diff
@@ -49,6 +49,28 @@ const fetchTasks = async () => {
   }
 }
 
+/**
+ * Fetch all participants whose task status is active
+ *
+ * @return {Promise<tasks|Array>}
+ */
+
+const fetchActiveTaskMembers = async () => {
```
ankurnarkhede, line 58: Unit tests pending for this function.
Can you add its tests in `test/unit/models/tasks.test.js`?

ankurnarkhede (changes requested):

models/tasks.js
harshith-venkatesh: I will create new issue for these test cases

ankurnarkhede (approved):
Merging the PR in the absence of all tests as the UI integration is blocked.
Issue for adding tests in the repo: https://github.com/Real-Dev-Squad/website-backend/issues/195

cc: @ankushdharkar

PR #178, opened 2021-02-13
`/members` API no longer returns user's private data

sanyogita14:
`phone` and `email` entered by the user is not returned by `/members` API.


swarajpure (approved):
Great! Thank you for the fix! 😁🎉

PR #182, opened 2021-02-15
rename non-production auth cookies

sanyogita14:
Authentication cookies renamed as `rds-session-<NODE_ENV>` for non-production environments

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -6,7 +6,7 @@
  * Documentation: https://github.com/lorenwest/node-config/wiki/Configuration-Files
  */
 
-var NODE_ENV = process.env.NODE_ENV
+const NODE_ENV = process.env.NODE_ENV
 module.exports = {
   port: 3000,
   enableFileLogs: true,
--- a/config/default.js
+++ b/config/default.js
@@ -5,6 +5,8 @@
  *
  * Documentation: https://github.com/lorenwest/node-config/wiki/Configuration-Files
  */
+
+var NODE_ENV = process.env.NODE_ENV
 module.exports = {
   port: 3000,
   enableFileLogs: true,
@@ -38,7 +40,7 @@ module.exports = {
   },
 
   userToken: {
-    cookieName: 'rds-session',
+    cookieName: `rds-session-${NODE_ENV}`,
     ttl: 30 * 24 * 60 * 60, // in seconds
     refreshTtl: 180 * 24 * 60 * 60, // in seconds
     publicKey: '<publicKey>',
--- a/test/integration/healthController.test.js
+++ b/test/integration/healthController.test.js
@@ -1,12 +1,14 @@
 const chai = require('chai')
 const { expect } = chai
 const chaiHttp = require('chai-http')
+// const { config } = require('winston')
 chai.use(chaiHttp)
 
 const app = require('../../server')
 const authService = require('../../services/authService')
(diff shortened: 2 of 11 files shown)


config/default.js
```diff
@@ -5,6 +5,8 @@
  *
  * Documentation: https://github.com/lorenwest/node-config/wiki/Configuration-Files
  */
+
+var NODE_ENV = process.env.NODE_ENV
```
swarajpure, line 9: Can we please use `const` here instead of `var` since the value will not be changed?

swarajpure (changes requested):
One minor change in the config file required.

Also, the tests are failing and they're failing because the cookie used to authenticate the user is being set by the name `rds-session` and according to the implementation of this PR, the cookie name will be called `rds-session-test`.

Since the tests are looking for a cookie named `rds-session`, they wont be able to find it and hence, the tests are failing. So you might want to declare a variable `cookie` or `cookieName` in all the files where we are using the cookie and let it take the cookie name from the config and pass that variable wherever we are passing the cookie in the request. You can try doing a global search for ```.set('cookie', `rds-session=${jwt}\`)``` in VSCode and you'll understand what I want to say.

Thank you! 😄

config/default.js
sanyogita14: Yes. Will carry out the changes. Thanks for the help!

config/default.js
```diff
@@ -38,7 +40,7 @@ module.exports = {
   },
 
   userToken: {
-    cookieName: 'rds-session',
+    cookieName: `rds-session-${NODE_ENV}`,
```
ankurnarkhede, line 43: We just need to rename the non-production cookies.

test/integration/healthController.test.js
```diff
@@ -1,12 +1,14 @@
 const chai = require('chai')
 const { expect } = chai
 const chaiHttp = require('chai-http')
+// const { config } = require('winston')
```
ankurnarkhede, line 4: Why is this line added?

test/integration/healthController.test.js
```diff
@@ -1,12 +1,14 @@
...
 chai.use(chaiHttp)
 
 const app = require('../../server')
 const authService = require('../../services/authService')
 const addUser = require('../utils/addUser')
-
+const config = require('config')
```
ankurnarkhede, line 10: `config` is accessible globally. Can you check if its accessible in the test scripts?
If yes, we can remove the import for it.

config/default.js
ankurnarkhede: @sanyogita14 , can you override the cookie name in production config to be `rds-session`?
We do not want the cookie name in production to be `rds-session-production`.

ankurnarkhede (changes requested):

config/default.js
sanyogita14: Alright..so I'll just specify it separately for the production file, right?

test/integration/healthController.test.js
sanyogita14: Yes I tried but the tests were failing so I had to import it into all the files

config/default.js
ankurnarkhede: Yes @sanyogita14

ankurnarkhede (approved):

PR #197, opened 2021-02-23
Updates packages

swarajpure:


package.json
```diff
@@ -14,45 +14,45 @@
...
+    "eslint-plugin-standard": "^4.1.0",
+    "mocha": "^8.3.0",
+    "nock": "^13.0.7",
+    "nodemon": "^2.0.7",
     "nyc": "^15.1.0",
     "pre-commit": "^1.2.2",
-    "sinon": "^9.0.3",
-    "standard": "^14.3.4"
+    "sinon": "^9.2.4",
+    "standard": "^16.0.3"
```
ankurnarkhede, line 55: Are we using this package now?

package.json
```diff
@@ -14,45 +14,45 @@
...
+    "joi": "^17.4.0",
     "jsdoc": "^3.6.6",
     "jsonwebtoken": "^8.5.1",
-    "morgan": "~1.9.1",
+    "morgan": "~1.10.0",
     "passport": "^0.4.1",
     "passport-github2": "^0.1.12",
-    "swagger-jsdoc": "^4.0.0",
-    "swagger-ui-express": "^4.1.4",
+    "swagger-jsdoc": "^6.0.6",
```
ankurnarkhede, line 33: @swarajpure , have you verified that the upgrading to the packages like swagger-jsdoc and eslint-config-standard, there are no breaking changes, as its a major version upgrade?

Post that, we can merge the PR.

ankurnarkhede (changes requested):

package.json
swarajpure: We're having a new warning:
image

Other than that, I tried to generate api schema, checked swagger docs, they look fine to me. Not sure how to check for jsdoc
Is there any proper way to check for breaking changes?

package.json
swarajpure: I read up the docs for this package and it is used an alternative to using ESLint. 
Since we're already using ESLint, this can be removed. I tried to run `lint` script and it works just fine. Getting output same as mentioned in the screenshot above

package.json
ankurnarkhede: @swarajpure , can you fix the new warnings occurring due to the upgraded version, if those can be fixed?

You can go to the changelog for that package and view it there. Check if any of the change affects us. I suppose there might nor be any change in the schema generation.

Changelog: https://github.com/Surnet/swagger-jsdoc/releases

Semantic versioning: https://github.com/Surnet/swagger-jsdoc/releases. This might help in understanding the versioning...

ankurnarkhede:
@swarajpure , can you also create a task as "Update dependencies" and link it with the PR?

package.json
swarajpure: Will fix the warning 👍🏻 

Yes I had seen that link already and comparing between our current package and the latest stable package just shows me some commits 🤔

package.json
swarajpure: Removed it.

package.json
swarajpure: Fixed the warning.

swarajpure:
> 
> 
> @swarajpure , can you also create a task as "Update dependencies" and link it with the PR?

Done

ankurnarkhede (approved):

PR #202, opened 2021-02-27
Sets sameSite attribute of cookie as lax

### LINKED ISSUE #196: Make the same site attribute of auth cookie as Lax
The auth cookie `rds-session` is currently called only from the pages with URL *.realdvsquad.com. Also, we do not expect any other page to call the API with auth cookies in it, although the request won't succeed as CORS check will fail.
Currently, the Same Site attribute is marked as `None`.
Also, we expect the cookies to be with the requests when a user clicks on a link to RDS on any other webpage, which creates navigation. Hence, it shouldn't be a `Strict` cookie.

Thus, owing to the above-mentioned reasons, we should convert it to `Lax`.

cc: @ankushdharkar

Rucha1499:
The `sameSite` attribute of cookie is set to `lax`. Fixes #196

### DIFF
--- a/controllers/authController.js
+++ b/controllers/authController.js
@@ -38,7 +38,8 @@ const githubAuth = (req, res, next) => {
         domain: rdsUiUrl.hostname,
         expires: new Date(Date.now() + config.get('userToken.ttl') * 1000),
         httpOnly: true,
-        secure: true
+        secure: true,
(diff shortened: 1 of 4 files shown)


ankurnarkhede (approved):
LGTM ✅

PR #206, opened 2021-02-28
Added unit tests for addOrUpdate user model function

ankurnarkhede:
- Added unit tests for addOrUpdate user model function


ankurnarkhede:
@swarajpure , this PR is added as a reference for adding unit tests for other model functions.

swarajpure (approved):

PR #209, opened 2021-02-28
Feature/api crypto get product [ 2 ]

prakashchoudhary07:
Merge After #208 

Closes https://github.com/Real-Dev-Squad/website-crypto/issues/155

PR #210, opened 2021-02-28
Feature/api crypto post products [ 3 ]

prakashchoudhary07:
Merge after #209


models/crypto.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+const addProduct = async (productData) => {
+  try {
+    const { id } = productData
+    const product = await cryptoProductsCollection.doc(id).get()
+    if (!product.exists) {
+      await cryptoProductsCollection.doc(id).set(productData)
+      return productData
+    } else {
+      return undefined
```
Kratika0907, line 39: By default it will return undefined, we can return proper message or error

Kratika0907 (changes requested):

models/crypto.js
prakashchoudhary07: Yes but that will give a linting error, This is DAO, is returning an error message necessary hear

prakashchoudhary07:
This is being deleted as the contracts were updated, and the project is on hold currently

PR #223, opened 2021-03-10
Create function for Mapping username-userId 

sanyogita14:
Function `toUserId` receives the username and returns the corresponding userId for that user.
Function `toUsername` does the opposite - userId received as a parameter and the corresponding username returned.


swarajpure (dismissed):
Looks good to me!
Thanks for this. Very much needed!

utils/idMapping.js
```diff
@@ -0,0 +1,37 @@
+const { fetchUser } = require('../models/users')
+
+/**
+ * Used for receiving userId when providing username
+ *
+ * @param username {String} - username of the User.
+ * @returns id {String} - userId of the same user
+ */
+const toUserId = async (username) => {
```
ankurnarkhede, line 9: Can we simply name the functions as getUserId and getUsername?

toUserId seems fine when its a class method, but here, its a function.

utils/idMapping.js
```diff
@@ -0,0 +1,37 @@
+const { fetchUser } = require('../models/users')
```
ankurnarkhede, line 1: Can we change the file name as users.js, so that in future, we can add other utilities related to user data?

ankurnarkhede (changes requested):
Need unit tests.

sanyogita14:
> Need unit tests.

Can I make a separate PR for the tests?

utils/idMapping.js
sanyogita14: yeah ok

ankurnarkhede:
> > Need unit tests.
> 
> Can I make a separate PR for the tests?

@sanyogita14 , we should add tests to the same PR if possible.

swarajpure (approved):
@ankurnarkhede Since the changes which you had mentioned were made, I'm merging this PR now. 
It's a small PR but this PR is critical for all the backend devs working on crypto tasks for public release on 14 March

I've created an issue for unit tests and assigned to @sanyogita14 #228

PR #230, opened 2021-03-12
Saving `userId` in db and returning `username` for /tasks API

sanyogita14:
`/tasksController` and `/models/tasks` modified to save userId to the database while the frontend sends and receives the username.


controllers/tasksController.js
```diff
@@ -1,4 +1,41 @@
 const tasks = require('../models/tasks')
+const userMapping = require('../utils/users')
+/**
+ * Converts the userIds entered in the array to corresponding usernames
+ * @param participantArray {array} : participants array to be updated
+ * @returns participantUsernames {array} : array of usernames of all participants
+ */
+const getParticipantUsernames = async (participantArray) => {
```
ankurnarkhede, line 8: @sanyogita14 , we should move these functions to utils/users file as these do not directly correspond to a route.

ankurnarkhede (changes requested):
@sanyogita14 , looks like this is modifying the response. Can you check the failing tests?

controllers/tasksController.js
sanyogita14: okay

sanyogita14:
> @sanyogita14 , looks like this is modifying the response. Can you check the failing tests?

Yes. I'm looking into it now. I'll have to change the POST Route function right? To return the original response instead of modifying it.

controllers/tasksController.js
```diff
@@ -27,7 +27,7 @@ const addNewTask = async (req, res) => {
  */
 const fetchTasks = async (req, res) => {
   try {
-    const allTasks = await tasks.fetchTasks()
+    const allTasks = await tasks.fetchTasks(req.body)
```
swarajpure, line 30: Any specific reason of passing `req.body` here?

swarajpure (commented):
A minor query.

Amazing job btw! ⚡

controllers/tasksController.js
sanyogita14: no actually, I think I added it previously when I added the id-mapping logic in this file. But the code's working fine otherwise. Should I remove it?

sanyogita14:
> A minor query.
> 
> Amazing job btw! ⚡

Thanks!!

controllers/tasksController.js
swarajpure: If we don't need it, we should remove it

sanyogita14: yeah ok. I'll remove it then

swarajpure (approved):
Looks good!

Good job! 🥳

Approving from my side. Let @ankurnarkhede have a look once.

models/tasks.js
```diff
@@ -1,6 +1,7 @@
 const firestore = require('../utils/firestore')
 const tasksModel = firestore.collection('tasks')
 const { fetchUser } = require('./users')
+const userMapping = require('../utils/users')
```
ankurnarkhede, line 4: We can change the name to `userUtils`, or any other descriptive name, as the file does not only contain the mapping functions.

ankurnarkhede (changes requested):
@sanyogita14, the tests for this have already been written. Can you modify the existing tests or write new ones to assert on this?

models/tasks.js
sanyogita14: ok

sanyogita14:
> @sanyogita14, the tests for this have already been written. Can you modify the existing tests or write new ones to assert on this?

Yes ok. I'll modify the existing tests

ankurnarkhede (approved):
LGTM ✅

PR #244, opened 2021-03-14
API implemented for fetching latest 10 transactions on basis of userId

rkmr039:


controllers/transactionsController.js
```diff
@@ -0,0 +1,25 @@
...
+ * Collects all transactions and sends only required data for each userId
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const fetchLatestTransactions = async (req, res) => {
+  try {
+    const userId = req.params.userId
+    const data = await transactionsModel.fetchLatestTransactions(userId)
```
swarajpure, line 13: We should be sending `username` in the URL. The userId is a random string, we shouldn't be using that to make the calls. 

We should be getting the `username` from the URL and then find the `userId` for that `username`. We already have functions which do that, please check: `utils/users.js`

models/transaction.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * This file contains wrapper functions to interact with the DB.
+ * This will contain the DB schema if we start consuming an ORM for managing the DB operations
+ */
+
+const firestore = require('../utils/firestore')
+const transactionsRef = firestore.collection('transaction')
```
swarajpure, line 7: Line 7 can be `transactionModel`

models/transaction.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * fetch latest 10 transactions from DB from specific userId provided
+ * @param userId { String }: User Id String to be used to fetch latest transactions
+ * @returns {Promise<{isUserExist: boolean, transactions: []}>}
+ */
+const fetchLatestTransactions = async (userId) => {
+  const transactions = await transactionsRef.where('userId', '==', userId).get
```
swarajpure, line 15: Line 15 can be transactionsRef

models/transaction.js
```diff
@@ -0,0 +1,31 @@
...
+  transactions.forEach((doc) => {
+    const transaction = doc.data()
+    transaction.dateInMillis = doc.data().dateTime._seconds * 1000
+    delete transaction.dateTime
+    res.push(transaction)
+  })
+  res.sort((a, b) => {
+    return a.dateInMillis > b.dateInMillis ? 1 : -1
+  })
+  return res.slice(0, 10)
```
swarajpure, line 26: I thought `res` is the express response object at first.

Please use suitable variables which will not confuse others.

`res` can be called `transactions` now if you see my comments above. (Sticking to the naming convention system we already have)

routes/transactionsRoute.js
```diff
@@ -0,0 +1,36 @@
...
+ *                    items:
+ *                      $ref: '#/components/schemas/transactions'
+ *       500:
+ *         description: badImplementation
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/errors/badImplementation'
+ */
+router.get('/fetchLatestTransactions/:userId', transactionsController.fetchLatestTransactions)
```
swarajpure, line 34: Do we really need such a long endpoint? It could be simply `/latest/:username` (`username` : the first review comment)

models/transaction.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+const firestore = require('../utils/firestore')
+const transactionsRef = firestore.collection('transaction')
+
+/**
+ * fetch latest 10 transactions from DB from specific userId provided
+ * @param userId { String }: User Id String to be used to fetch latest transactions
+ * @returns {Promise<{isUserExist: boolean, transactions: []}>}
+ */
```
swarajpure, line 13: I guess we shouldn't hardcode the value `10`
We could pass query params for the number of transactions required. Default value can be 10 though.

swarajpure (changes requested):
Great one!

Thanks for this! 👍🏻 

Requesting some changes. Please check 🙂

swarajpure:
Also, can you please link the API contracts?

rkmr039:
> Also, can you please link the API contracts?

Yeah Sure ......will make all the necessary changes as mentioned above.

rkmr039:
API Contract available here https://github.com/Real-Dev-Squad/website-api-contracts/pull/48

controllers/transactionsController.js
swarajpure: Please check this comment @rkmr039

routes/transactionsRoute.js
```diff
@@ -0,0 +1,36 @@
...
+ *                    items:
+ *                      $ref: '#/components/schemas/transactions'
+ *       500:
+ *         description: badImplementation
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/errors/badImplementation'
+ */
+router.get('/fetchLatest/:username/:noOfRecords', transactionsController.fetchLatest)
```
swarajpure, line 34: Don't you think passing `noOfRecords` as a query param instead of path param would be better? 🤔
It should be optional in my opinion. Let me know what you think about it 😉

controllers/transactionsController.js
```diff
@@ -0,0 +1,29 @@
...
+ * Collects all transactions and sends only required data for each userId
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const fetchLatest = async (req, res) => {
+  try {
+    const userId = await getUserId(req.params.username)
+    const noOfRecords = req.params.noOfRecords
```
swarajpure, line 14: (read the comment on routes files first)
Once we have it as query param, since query params are optional, we should have a default value for number of transactions. Something like 10?

controllers/transactionsController.js
```diff
@@ -0,0 +1,29 @@
...
+    const data = await transactionsModel.fetchLatest(userId, noOfRecords)
+    return res.json({
+      message: 'Transactions returned successfully!',
+      responseCode: 200,
```
swarajpure, line 18: Line 18 is not needed. `res.json` without `status` means the response code is 200

controllers/transactionsController.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param res {Object} - Express response object
+ */
+
+const fetchLatest = async (req, res) => {
+  try {
+    const userId = await getUserId(req.params.username)
+    const noOfRecords = req.params.noOfRecords
+    const data = await transactionsModel.fetchLatest(userId, noOfRecords)
+    return res.json({
+      message: 'Transactions returned successfully!',
```
swarajpure, line 17: Please handle the case when there are no transactions by the given user

models/transaction.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * This file contains wrapper functions to interact with the DB.
+ * This will contain the DB schema if we start consuming an ORM for managing the DB operations
+ */
+
+const firestore = require('../utils/firestore')
+const transactionsModel = firestore.collection('transaction')
+
+/**
+ * fetch latest 10 transactions from DB from specific userId provided
```
swarajpure, line 10: Please have a look at line 10, needs to be changed. I meant the number of transactions

controllers/transactionsController.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Collects all transactions and sends only required data for each userId
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const fetchLatest = async (req, res) => {
+  try {
+    const userId = await getUserId(req.params.username)
```
swarajpure, line 13: Please handle the case where given `username` doesn't exist

routes/transactionsRoute.js
```diff
@@ -0,0 +1,36 @@
+const express = require('express')
+const router = express.Router()
+const transactionsController = require('../controllers/transactionsController.js')
+
+/**
+ * @swagger
+ * /transactions/fetchLatestTransactions:
```
swarajpure, line 7: Please generate API schema for new route. Check scripts in package.json for the same.

controllers/transactionsController.js
```diff
@@ -0,0 +1,29 @@
+const transactionsModel = require('../models/transaction')
```
swarajpure, line 1: This file since it is already inside `controllers` folder, need not be called `transactionController`. Simply `transactions` would do.

swarajpure (changes requested):
Requesting some changes about edge cases and API endpoint. Please check.

Thank you! 😁

routes/transactionsRoute.js
rkmr039: Yes you are right....noOfRecords should be optional.

controllers/transactionsController.js
rkmr039: All existing controllers are following same naming convention.

routes/transactionsRoute.js
rkmr039: Generated

controllers/transactionsController.js
rkmr039: If user doesn't exist then returned userId will be undefined, on basis of this if condition added

models/transaction.js
rkmr039: Changed to  more generic description

controllers/transactionsController.js
rkmr039: Done

controllers/transactionsController.js
rkmr039: responseCode removed

controllers/transactionsController.js
rkmr039: Default value set to 10 if query param noOfRecords is not available in url

controllers/transactionsController.js
```diff
@@ -0,0 +1,39 @@
...
+      } else {
+        return res.json({
+          message: 'No record exist!'
+        })
+      }
+    } else {
+      return res.json({
+        message: 'User does not exist!'
+      })
+    }
```
swarajpure, line 30: So many `if`, `else` and `return`, it looked confusing to me at the first glance 😅 
Can we simplify this? (Hint: ternary operator)

controllers/transactionsController.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Collects all transactions and sends only required data for username spicified in url
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const fetchLatest = async (req, res) => {
+  try {
+    const userId = await getUserId(req.params.username)
+    if (userId !== undefined) {
```
swarajpure, line 13: Can this be simply `if (userId)`?

controllers/transactionsController.js
```diff
@@ -0,0 +1,39 @@
...
+      const noOfRecords = req.query.noOfRecords
```
swarajpure, line 14: Does the query param need to be this long? Can it be simply `n`? We could specify what `n` is in API contract

Also, wouldn't it be better if the default value `10` is set here itself (using `||` operator)

controllers/transactionsController.js
```diff
@@ -0,0 +1,39 @@
...
+        })
+      } else {
+        return res.json({
+          message: 'No record exist!'
+        })
+      }
+    } else {
+      return res.json({
+        message: 'User does not exist!'
+      })
```
swarajpure, line 29: In case of user not existing, the status code should not be 200, it should be 404
Use `res.boom.notFound.....`

swarajpure (changes requested):
Requesting some changes related to readability of the code. Please check

controllers/transactionsController.js
rkmr039: Oh ok...will change it

controllers/transactionsController.js
rkmr039: Ok will try

controllers/transactionsController.js
rkmr039: Done

controllers/transactionsController.js
rkmr039: Order = o (ASC or DESC)
NoOfRecords = n (1,2,3....)

controllers/transactionsController.js
rkmr039: Done

rkmr039: Done

controllers/transactionsController.js
rkmr039: Done

controllers/transactions.js
```diff
@@ -0,0 +1,36 @@
...
+      const o = req.query.o || 'DESC'
+      const data = await transactionsModel.fetch(userId, n, o)
+      if (data.length > 0) {
+        return res.json({
+          message: 'Transactions returned successfully!',
+          data
+        })
+      } else {
+        return res.boom.notFound('No transactions exist!')
+      }
```
swarajpure, line 24: Transactions not existing is not the user's problem. 4xx response is when there's something wrong from the user's side.
The API call isn't bad when there are no transactions. It should be 200 with appropriate message.

Have a look here;
https://github.com/Real-Dev-Squad/website-backend/blob/7bc36f4efe28d434f9f55bcbdab10c66bb1f7aed/controllers/members.js#L16

controllers/transactionsController.js
rkmr039: Done

controllers/transactions.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param res {Object} - Express response object
+ */
+
+const fetch = async (req, res) => {
+  try {
+    const userId = await getUserId(req.params.username)
+    if (userId) {
+      const n = req.query.n || 10
+      const o = req.query.o || 'DESC'
+      const data = await transactionsModel.fetch(userId, n, o)
```
swarajpure, line 16: Now I'm seeing the problem. If we keep using alphabets, it might become difficult to keep track of what's what.
`size` instead of `n` and `order` instead of `o` would make more sense, right? We wouldn't have to explicitly mention in the docs what's doing what
Apologies for the inconvenience.

swarajpure (approved):
Requesting some minor changes. Please check

swarajpure (changes requested):
Requesting changes, not approved yet. Approved earlier by mistake.

controllers/transactions.js
rkmr039: renamed according to purpose

controllers/transactions.js
rkmr039: Done

controllers/transactions.js
```diff
@@ -0,0 +1,33 @@
...
+ * Collects all transactions and sends only required data for username spicified in url
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const fetch = async (req, res) => {
+  try {
+    const userId = await getUserId(req.params.username)
+    if (userId) {
+      const orderBy = parseInt(req.query.orderBy, 10) || 'DESC'
```
swarajpure, line 14: How is `orderBy` going to be parsed into integer? 🤔

routes/transactions.js
```diff
@@ -0,0 +1,44 @@
...
+const router = express.Router()
+const transactions = require('../controllers/transactions.js')
+
+/**
+ * @swagger
+ * /transactions/:username:
+ *   get:
+ *     summary: Transactions done by username in Real Dev Squad
+ *     tags:
+ *       - Transaction Management
```
swarajpure, line 11: tag can be just `Transactions`

models/transaction.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * fetch latest N transactions from transactions collection for specific userid provided
+ *
+ * @param userId { String }: User Id String to be used to fetch latest transactions
+ * @param limit { number }: no of records to fetch from transaction table, if not specified in URL then default will be 10
+ * @param offset { number }: starting index of set of records, default value is 0, useful for pagination
+ * @param orderBy { String }: to order the transactions in ascending or descending manner, by default will fetch latest first
+ * @returns {Promise<{transactions: []}>}
+ */
+const fetch = async (userId, offset, limit, orderBy) => {
```
swarajpure, line 18: Better to name this function as `fetchTransactions`? 🤔

swarajpure (changes requested):
Some queries, please check

controllers/transactions.js
rkmr039: Oh ...my bad

rkmr039: parseInt removed

models/transaction.js
rkmr039: fetch is defined under transaction controller, it's means fetching the transactions

routes/transactions.js
rkmr039: Done

models/transaction.js
swarajpure: We already have a util function called `fetch` which is used for network calls. That might become confusing. Also if I use your function in some other file and I destructure it there, it won't make any sense as to what the function is doing there.. I hope that makes sense

rkmr039: For the functions like (toString or fetch), same name is suggested and to distinguish we have documentation. User can't remember diff names for same kind of task related to diff APIs.

swarajpure: > Also if I use your function in some other file and I destructure it there, it won't make any sense as to what the function is doing there.. I hope that makes sense

Say I use this function in `models/wallets.js`, and I write `fetch(.....)` how do you know what is being fetched? The function names should be self-explanatory. You shouldn't need to look at the imports to know where the file is coming from.

Imagine all the places we replace `fetchUser => fetch`, `fetchTransactions => fetch` and if I had to use both the functions in one single file, how would I differentiate?

rkmr039: For that we have documentation. `fetch(..)` is like atomic operation which every API supposed to have, if you provide different name for each then you will end up with 1000s of method name.

swarajpure: You didn't say anything about:
> Imagine all the places we replace `fetchUser => fetch`, `fetchTransactions => fetch` and if I had to use both the functions in one single file, how would I differentiate?

rkmr039: every fetch is followed an entity like user or transaction, if you are creating object for that entity the you should be aware of object name(keep object name in that way). At last we have documentation in place for every help required.

swarajpure: I'm talking about function names themselves, not the objects returned by them

Also, how do you know that this is going to be the only `fetch` function in this file? Can we not have `fetchTransaction` to get one particular transaction details given it's id tomorrow? What would you do that in that case?

We need to differentiate between `fetchTransactions` and `fetchTransaction`

rkmr039: You can fetch any number of transactions with this API but value of limit is required for that otherwise it will return only 10 maximum records

harshith-venkatesh: Hey @rkmr039 , Even I agree with what @swarajpure is mentioning here, even though the function is defined inside models/transaction , it will create confusions when we try to import fetch from general util and usually function names are self explanatory...

swarajpure: This function will fetch the transactions by a user. I'm talking about a function which would return transaction details given transaction id

harshith-venkatesh: With name fetchTransactions , we are defining a function in general which has a capability to fetch N transactions 0<=N<=infinity

I hope I am describing it right

harshith-venkatesh: fetchTransactions/: transactionId can provide unique transaction details
Pardon for some spelling mistakes as I am using mobile to comment here

swarajpure: In fact, the name shouldn't be just `fetchTransactions`, it should be `fetchTransactionsByUser`

Tomorrow we could have:
- `fetchAllTransactions`: Returns ALL of the transactions in the DB (by all users)
- `fetchTransactionById`: Returns a transaction by transactionId

So let's go with `fetchTransactionsByUser` @rkmr039

swarajpure: Regarding the 1000s of method names, would you prefer 1000 functions with different names (self-explanatory) or 1000 functions with same name (all called `fetch`)? 😉

rkmr039: Every function performs same kind of task can have same name although in diff packages.

rkmr039: renamed as `fetchByUsername`

harshith-venkatesh: @swarajpure @rkmr039 great discussion guys, we have come to collective conclusion now

rkmr039: Actually I always want to be a part of team like this, but in my current role i can name methods anything ....nobody checks

swarajpure: > Every function performs same kind of task can have same name although in diff packages.

Why are you not getting the point?
```JS
const { fetch } = require('./models/transactions')
const { fetch } = require('./models/users')
```

Is this the confusion you would like to work with?

Every function HAS to be named as per the job it is supposed to do irrespective of the folder/file name. If the function was named `fetchTransactionsModel` then keep the `model` in function name makes no sense, but `transactions` MUST be kept. 

This function may not remain exclusive to this route. We may use this in other routes as well. Do you want the developer to look at the imports every single time to make sense of what the function is supposed to do? That defeats the purpose of having self-explanatory function names in the first place. 

Why not just name all the functions as `a`? Short and sweet.

swarajpure: > renamed as `fetchByUsername`

Even this doesn't make much sense.
Why should I have to look at the file name to make sense of what this function fetches? 
Why should I need to look at documentation if having self-explanatory function name could solve the problem?

rkmr039: Do you want me to rename it as `fetchTransactionByUsername` ?

swarajpure: `fetchTransactionByUserId` in this case

harshith-venkatesh: Yeah, fetchTransactionsByUserId would be better in this case right 
And the name remains same in model as well as in controller for easy navigation to different functions

swarajpure: Not exactly the same, the controller is taking `username` and the model is taking `userId`, but they'll be similar, yeah

harshith-venkatesh: Even though it is taking username in query parama , it is used to fetch userId and then the model operation occurs right... My implications was in that fashion

rkmr039: Done

harshith-venkatesh (dismissed):

swarajpure (changes requested):
Good job in there! 🥳 
Please add tests, after which we can merge this 👍🏻

test/fixtures/transaction/transactions.js
```diff
@@ -0,0 +1,27 @@
...
+    {
+      userId: 'kratika',
+      userTo: 'Rishab',
+      amount: 400,
+      typeOfTransaction: 'Credit',
+      typeOfCurrency: 'Gold',
+      dateInMillis: 1615707000000
+    }
+  ]
+}
```
swarajpure, line 27: I guess we don't have routes/model functions to make a transaction, right? That's why fixtures, right?

swarajpure (changes requested):
Having some queries, please check.
Also, please add integration tests.

Thanks for your patience till now 🎉 🙂

test/fixtures/transaction/transactions.js
rkmr039: Yes....we need some method to record transaction

test/unit/models/transaction.test.js
```diff
@@ -0,0 +1,32 @@
...
+    await cleanDb()
+  })
+
+  describe('when fetchTransactionsByUserId called, it ', function () {
+    it('should fetch transaction', async function () {
+      const transactionData = transactionDataArray[0]
+      const data = await transactionModelsImpl.fetchTransactionsByUserId('kratika', 0, 2, 'DESC')
+      Object.keys(transactionData).forEach(key => {
+        expect(data[key]).to.deep.equal(data[key])
+      })
```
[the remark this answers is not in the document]
rkmr039: Yeah...my mistake 
Have done required changes

Issue #246, opened 2021-03-14
Remove extra validation fields from user validator


swarajpure:
Currently we are having extra fields in users validators
We have to remove the fields which are not as per the API contract: https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/users#user-object

File concerned: middlewares/validators/user.js

PR #259, opened 2021-03-15
Removed extra validation fields from user validator

TD-17:
issue  #246

### DIFF
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -11,8 +11,6 @@ const updateUser = async (req, res, next) => {
     company: joi.string().optional(),
     designation: joi.string().optional(),
     img: joi.string().optional(),
-    github_display_name: joi.string().optional(),
-    github_id: joi.string().optional(),
     linkedin_id: joi.string().optional(),
     twitter_id: joi.string().optional(),
     instagram_id: joi.string().optional(),
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -16,8 +16,7 @@ const updateUser = async (req, res, next) => {
     linkedin_id: joi.string().optional(),
     twitter_id: joi.string().optional(),
     instagram_id: joi.string().optional(),
-    website: joi.string().optional(),
-    isMember: joi.boolean().optional()
+    website: joi.string().optional()
   })
 
   try {
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -18,7 +18,7 @@ const updateUser = async (req, res, next) => {
     instagram_id: joi.string().optional(),
     site: joi.string().optional(),
     isMember: joi.boolean().optional(),
-    tokens: joi.object().optional(),
+    tokens: joi.object().optional()
   })
 
   try {
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -16,9 +16,8 @@ const updateUser = async (req, res, next) => {
     linkedin_id: joi.string().optional(),
     twitter_id: joi.string().optional(),
     instagram_id: joi.string().optional(),
-    site: joi.string().optional(),
-    isMember: joi.boolean().optional(),
-    tokens: joi.object().optional()
+    website: joi.string().optional(),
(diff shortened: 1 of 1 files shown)


middlewares/validators/user.js
```diff
@@ -19,8 +18,7 @@ const updateUser = async (req, res, next) => {
     instagram_id: joi.string().optional(),
     site: joi.string().optional(),
     isMember: joi.boolean().optional(),
-    tokens: joi.object().optional(),
-    website: joi.string().optional()
```
swarajpure, line 23: I just realised we're sending `website` from the signup form, I'll change the API contract, please remove `site` and keep `website`

swarajpure:
Need to have clarification from the frontend engineer who worked on the signup form as what key are we using for the company name of the user. I highly doubt it's `company_name`. Some users in the DB have got `company` while some have got `company_name`

If we make any changes to the validators, the signup form will stop working which is not expected at this point of time. Let's keep it on hold till then.

TD-17:
> Need to have clarification from the frontend engineer who worked on the signup form as what key are we using for the company name of the user. I highly doubt it's `company_name`. Some users in the DB have got `company` while some have got `company_name`
> 
> If we make any changes to the validators, the signup form will stop working which is not expected at this point of time. Let's keep it on hold till then.

yeah okay!

middlewares/validators/user.js
```diff
@@ -19,8 +18,7 @@ const updateUser = async (req, res, next) => {
...
+    tokens: joi.object().optional()
```
ankurnarkhede, line 21: @swarajpure , we won't actually be storing tokens via this route, right?
So this validation can be removed.

ankurnarkhede (changes requested):

middlewares/validators/user.js
swarajpure: Yes, that can be removed as well

middlewares/validators/user.js
swarajpure: @TD-17

middlewares/validators/user.js
swarajpure: @TD-17

middlewares/validators/user.js
swarajpure: Let's remove `company_name`

TD-17: yeah okay

middlewares/validators/user.js
```diff
@@ -17,10 +16,8 @@ const updateUser = async (req, res, next) => {
     linkedin_id: joi.string().optional(),
     twitter_id: joi.string().optional(),
     instagram_id: joi.string().optional(),
-    site: joi.string().optional(),
-    isMember: joi.boolean().optional(),
-    tokens: joi.object().optional(),
-    website: joi.string().optional()
+    website: joi.string().optional(),
+    isMember: joi.boolean().optional()
```
swarajpure, line 20: We don't need this here -- `isMember`
It should not be here from security pov
Otherwise users will mark themselves as members 🤣 

Please remove it from here

TD-17: oh yeahh! Okay

swarajpure (changes requested):
Can we also remove `github_display_name` and `github_id` since this is data coming from github, we shouldn't allow users to change it themselves.

This is last change, after which we can merge this one :P

TD-17:
> Can we also remove `github_display_name` and `github_id` since this is data coming from github, we shouldn't allow users to change it themselves.
> 
> This is last change, after which we can merge this one :P

yess okay :)

swarajpure (approved):
Looks good to me!
Thanks! 

Let's wait for at least one more reviewer to have a look here as 1 mistake here can cause a big problem somewhere. Let's not merge till then.

TD-17:
> Looks good to me!
> Thanks!
> 
> Let's wait for at least one more reviewer to have a look here as 1 mistake here can cause a big problem somewhere. Let's not merge till then.

yeah cool

sumitd94 (approved):

PR #264, opened 2021-03-16
[ON HOLD]Create wallets for all users

prakashchoudhary07:


ankurnarkhede (changes requested):
@prakashchoudhary07 , can you add a description as to why this is required?
Can we create a wallet for the user while fetching their details if we get the error in getting the wallet?
If not, how and when will the current route be executed?

ankushdharkar:
@ankurnarkhede That's a good point. Taking care of it here: https://github.com/Real-Dev-Squad/website-backend/issues/284

ankushdharkar:
We don't need this anymore. Closing it.

PR #266, opened 2021-03-17
bugfix change return err to throw err

prakashchoudhary07:


Rucha1499 (approved):
Looks great! 🔥

swarajpure (approved):
Looks good! Thanks for the fix!

PR #271, opened 2021-03-18
[WIP] challenges API refactored

correaswebert:
TODO:
Request for change in API contract. The post request also does the job of get request.

Doubts:
Should I create a validation function instead of validating inline?

### DIFF
--- a/controllers/challengeController.js
+++ b/controllers/challengeController.js
@@ -3,37 +3,49 @@ const challengeQuery = require('../models/challenges')
 const ERROR_MESSAGE = 'Something went wrong. Please try again or contact admin'
 
 /**
- * Get the challenges or add the challenge
+ * Get the challenges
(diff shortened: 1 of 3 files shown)


controllers/challengeController.js
```diff
@@ -3,37 +3,49 @@ const challengeQuery = require('../models/challenges')
...
-            challenges: challengeAdded
-          })
-        }
-      } else {
-        return res.boom.notFound('Unable to add challenge')
-      }
+    const { title, level, start_date: startDate, end_date: endDate } = req.body
+    if (!title || !level || !startDate || !endDate) {
+      return res.boom.badData('Empty fields received!')
+    }
```
swarajpure, line 35: Please make a middleware for validating. The flow wouldn't reach the controller if there are wrong/missing fields.

PR #277, opened 2021-03-20
User type in GET Requests

rajakvk:
Back end will return newly joined member when hitting /users?userType=new json end point.
We need to add new field called 'userType' in the db. 
Allowed userType values could be 'new', 'blocked', 'member'.
```js
  const allowedUserTypes = ['new', 'blocked', 'member']
```
Please review @ankurnarkhede


models/users.js
```diff
@@ -54,11 +54,21 @@ const addOrUpdate = async (userData, userId = null) => {
...
  */
 const fetchUsers = async (query) => {
+  const allowedUserTypes = ['new', 'blocked', 'member']
+  let snapshot
   try {
-    const snapshot = await userModel
-      .limit(parseInt(query.size) || 100)
-      .offset((parseInt(query.size) || 100) * (parseInt(query.page) || 0))
-      .get()
+    if (typeof query.userType !== 'undefined' && allowedUserTypes.includes(query.userType)) {
```
sumitd94, line 60: is it needed to check for `undefined` ?
cant we use something like 
`if(query.userType && ....)`

sumitd94 (changes requested):
One small change please.

models/users.js
```diff
@@ -54,11 +54,21 @@ const addOrUpdate = async (userData, userId = null) => {
  * @return {Promise<userModel|Array>}
  */
 const fetchUsers = async (query) => {
+  const allowedUserTypes = ['new', 'blocked', 'member']
```
sumitd94, line 57: Please update the `swaggerDocs` and generate the API Schema accordingly by running the command
`npm run generate-api-schema`

sumitd94 (changes requested):

models/users.js
rajakvk: Updated swaggerDefinition.js file with additional field "userType".

rajakvk:
@sumitd94 is it still relevant? If not, we may close

Issue #283, opened 2021-03-21
Tests for /wallet


ankushdharkar:
- [ ] User can GET their own wallet
- [ ] User cannot GET someone else's wallet if not authorized
- [ ] If No wallet is found, an empty wallet is created for the user and 1000 dineros is added
- [ ] Authorized user can GET wallet by username

sebuelias:
Can I take this up?

sebuelias:
Hi @ankushdharkar sorry about the delay.
About this issue, I was not able to find the API contract for the wallet, so what are the expected response/errors?
For the 3rd point, we by default create a wallet when adding a new user #285 , what exactly is the 3rd scenario to check if the new wallet is created for a new user with 1000 dinero?

PR #285, opened 2021-03-21
Create and initialize wallet if not found for user

### LINKED ISSUE #284: Create wallet for existing users on GET
For the current users, if the wallet doesn't already exist, create one, and add 1000 Dineros into it for them as backfill

ankushdharkar:
Closes #284 

Tests ticket: #283

### DIFF
--- a/controllers/usersController.js
+++ b/controllers/usersController.js
@@ -1,7 +1,6 @@
 
 const userQuery = require('../models/users')
 const accountOwners = require('../mockdata/appOwners')
-const { fetchWallet, createWallet } = require('../models/wallets')
 /**
  * Fetches the data about our users
  *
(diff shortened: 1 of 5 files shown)


controllers/usersController.js
```diff
@@ -126,11 +125,9 @@ const updateSelf = async (req, res) => {
     }
 
     const user = await userQuery.addOrUpdate(req.body, userId)
-    const userWallet = await fetchWallet(userId)
-    if (!userWallet.id) {
-      await createWallet(userId, { dinero: 1000 })
-    }
-    if (!user.isNewUser) {
```
ankushdharkar, line 133: Once a new user is signed up, there are additional tasks we may want to do
- Initialize their wallet
- Update the number of users
- Show it on members page, etc

Those post-signup events should exist in a separate function

ankurnarkhede (approved):
LGTM ✅

PR #288, opened 2021-03-22
[fix] stale pr API now does pagination

correaswebert:
1. Stale PRs now accepts pagination params
2. The `githubServices` is updated
3. The controller for `pullrequests` is updated

### DIFF
--- a/controllers/pullRequests.js
+++ b/controllers/pullRequests.js
@@ -3,12 +3,16 @@ const githubService = require('../services/githubService')
 /**
  * Get Latest PRs in open or stale state for Real Dev Squad repos
  *
+ * @param {boolean} page page number to retrieve
+ * @param {boolean} perPage number of PRs per page
  * @param {boolean} isOpen boolean value indication open or stale
- * Get stale PRs in open state for Real Dev Squad repos
+ * @param {string} username retrieve PRs of username
+ *
+ * @returns {Array} List of PRs based on provided conditions
  */
-const getPRs = async (page = 1, n = 10, isOpen = true, username = null) => {
+const getPRs = async ({ page = 1, perPage = 10, isOpen = true, username = null }) => {
   try {
-    const { data } = await githubService.fetchPRs(page, n, isOpen)
+    const { data } = await githubService.fetchPRs({ page, perPage, isOpen, username })
     if (!data.total_count) {
       return []
     }
@@ -28,18 +32,19 @@ const getPRs = async (page = 1, n = 10, isOpen = true, username = null) => {
 const getUserPRs = async (req, res) => {
   try {
     const { username } = req.params
-    const { data } = await githubService.fetchPRsByUser(username)
 
-    if (data.total_count) {
-      const allPRs = githubService.extractPRdetails(data)
-      return res.json({
-        message: 'Pull requests returned successfully!',
-        pullRequests: allPRs
-      })
+    let message
+    const prs = await getPRs({ username })
+
+    if (prs.length) {
+      message = 'User PRs'
+    } else {
+      message = 'No pull requests found!'
     }
+
     return res.json({
-      message: 'No pull requests found!',
-      pullRequests: []
+      message: message,
+      pullRequests: prs
     })
   } catch (err) {
     return res.boom.badImplementation('Something went wrong please contact admin')
@@ -55,11 +60,11 @@ const getUserPRs = async (req, res) => {
  */
 const getStalePRs = async (req, res) => {
   try {
-    const { page, n } = req.query
+    const { page, n: perPage } = req.query
 
     let message
     const isOpen = false
-    const prs = await getPRs(page, n, isOpen)
+    const prs = await getPRs({ page, perPage, isOpen })
 
     if (prs.length) {
       message = 'Stale PRs'
@@ -85,10 +90,10 @@ const getStalePRs = async (req, res) => {
  */
 const getOpenPRs = async (req, res) => {
   try {
-    const { page, n } = req.query
+    const { page, n: perPage } = req.query
 
     let message
-    const prs = await getPRs(page, n)
+    const prs = await getPRs({ page, perPage })
 
     if (prs) {
       message = 'Open PRs'
--- a/controllers/pullRequests.js
+++ b/controllers/pullRequests.js
@@ -33,18 +33,12 @@ const getUserPRs = async (req, res) => {
   try {
     const { username } = req.params
 
-    let message
-    const prs = await getPRs({ username })
-
-    if (prs.length) {
-      message = 'User PRs'
-    } else {
-      message = 'No pull requests found!'
-    }
+    const pullRequests = await getPRs({ username })
+    const message = pullRequests.length ? 'User PRs' : 'No pull requests found!'
 
     return res.json({
-      message: message,
-      pullRequests: prs
+      message,
+      pullRequests
     })
   } catch (err) {
     return res.boom.badImplementation('Something went wrong please contact admin')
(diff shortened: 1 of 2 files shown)


services/githubService.js
```diff
@@ -99,18 +99,18 @@ const fetchPRsByUser = async (username) => {
...
+const fetchStalePRs = async (pageNumber) => {
   try {
     const url = getGithubURL({
       is: 'open'
     }, {
       sort: 'created',
       order: 'asc',
-      per_page: 5,
-      page: 1
+      per_page: 10,
```
swarajpure, line 112: Can we also keep `per_page` variable? `10` can be default value

correaswebert: Yeah sure 👍🏾

ankurnarkhede (changes requested):
@correaswebert , can we get the changes done and merge this?

services/githubService.js
correaswebert: done

controllers/pullRequestsController.js
```diff
@@ -1,57 +1,82 @@
...
  * @param res {Object} - Express response object
  */
-
 const getUserPRs = async (req, res) => {
   try {
-    const { data } = await githubService.fetchPRsByUser(req.params.username)
+    const { username } = req.params
+
+    let message
+    const prs = await getPRs({ username })
```
swarajpure, line 37: can be called as `pullRequests` so it'd become better while returning (read comment below)

controllers/pullRequestsController.js
```diff
@@ -1,57 +1,82 @@
...
+      message = 'User PRs'
+    } else {
+      message = 'No pull requests found!'
     }
+
     return res.json({
-      message: 'No pull requests found!',
-      pullRequests: []
+      message: message,
+      pullRequests: prs
```
swarajpure, line 47: If you rename `prs` to `pullRequests`, line 47 can be simply `pullRequests`

controllers/pullRequestsController.js
```diff
@@ -1,57 +1,82 @@
...
+      message = 'Stale PRs'
+    } else {
+      message = 'No pull requests found!'
     }
+
     return res.json({
-      message: 'No pull requests found!',
-      pullRequests: []
+      message: message,
+      pullRequests: prs
```
swarajpure, line 77: Same thing here as above.

controllers/pullRequestsController.js
```diff
@@ -65,21 +90,22 @@ const getStalePRs = async (req, res) => {
...
+      message = 'Open PRs'
+    } else {
+      message = 'No pull requests found!'
     }
+
     return res.json({
-      message: 'No pull requests found!',
-      pullRequests: []
+      message: message,
+      pullRequests: prs
```
swarajpure, line 106: Same

services/githubService.js
```diff
@@ -1,29 +1,40 @@
-const { fetch } = require('../utils/fetch')
+const axios = require('axios')
```
swarajpure, line 1: Why remove `fetch`? 🤔

controllers/pullRequestsController.js
```diff
@@ -1,57 +1,82 @@
...
-    if (data.total_count) {
-      const allPRs = githubService.extractPRdetails(data)
-      return res.json({
-        message: 'Pull requests returned successfully!',
-        pullRequests: allPRs
-      })
+    if (prs.length) {
+      message = 'User PRs'
+    } else {
+      message = 'No pull requests found!'
```
swarajpure, line 42: We need not do this
While returning we can simply do
`message: prs.length ? 'User PRs' : 'No pull requests found!'`

controllers/pullRequestsController.js
```diff
@@ -1,57 +1,82 @@
...
     })
   } catch (err) {
-    logger.error(`Error while processing pull requests: ${err}`)
     return res.boom.badImplementation('Something went wrong please contact admin')
   }
 }
 
 /**
- * Get stale PRs in open state for Real Dev Squad repos
+ * Get Latest PRs in stale state for Real Dev Squad repos
```
swarajpure, line 55: We return `oldest stale PRs` first
Check here: https://api.realdevsquad.com/pullrequests/stale

swarajpure (changes requested):
Requesting some changes. Please check

controllers/pullRequestsController.js
swarajpure: Latest stale PRs would become same as `pullrequests/open`

correaswebert: Oops, forgot to update the comment here ig

services/githubService.js
correaswebert: We could use the custom fetch if we were doing something extra in our custom function...
Like we could add the `credentials: include` code in our custom fetch, making the call in the controller/service easier to read.
But here it was a simple get request, reduces the call-stack in case of finding errors

controllers/pullRequestsController.js
correaswebert: Latest PR in stale state means that latest stale PRs first right?

services/githubService.js
swarajpure: Then the whole purpose of wrapper function gets lost. We might not use axios tomorrow. Will you then sit and manually change `axios` to `axios-alternative` in every network call? 🤔 
For stale, open PRs and PRs by users.. we have limited network calls now.. but what if have 100s of them?

correaswebert: Cool... I just changed the name in the module scope so that there is no confusion between node's fetch and our fetch wrapper 👍🏾

correaswebert: Done 👍🏾

controllers/pullRequests.js
```diff
@@ -1,57 +1,69 @@
...
+    if (!data.total_count) {
+      return []
+    }
+    return githubService.extractPRdetails(data)
+  } catch (err) {
+    logger.error(`Error while processing pull requests: ${err}`)
+    throw err
+  }
+}
+
```
swarajpure, line 25: What is this doing in controllers?

controllers/pullRequests.js
```diff
@@ -1,57 +1,69 @@
...
  * @param req {Object} - Express request object
  * @param res {Object} - Express response object
  */
-
 const getUserPRs = async (req, res) => {
   try {
-    const { data } = await githubService.fetchPRsByUser(req.params.username)
+    const { username } = req.params
+
+    const pullRequests = await getPRs({ username })
```
swarajpure, line 36: Why are we not taking care of query params passed?

controllers/pullRequests.js
```diff
@@ -1,57 +1,69 @@
...
  * @param {Object} req
  * @param {Object} res
  * @todo create cache for RDS usernames <> github usernames
  */
 const getStalePRs = async (req, res) => {
   try {
-    const { data } = await githubService.fetchStalePRs()
+    const { page, n: perPage } = req.query
+
+    const pullRequests = getPRs({ page, perPage, isOpen: false })
```
swarajpure, line 59: Why `isOpen: false`?
Stale PRs = Open PRs sorted in reverse order (oldest first)

services/githubService.js
```diff
@@ -121,19 +108,28 @@ const fetchStalePRs = async () => {
...
+    if (username) {
+      const user = await fetchUser({ username })
+      searchParams.author = user.github_id
+    }
+
+    const url = getGithubURL(
+      searchParams,
+      {
+        sort: 'created',
+        order: isOpen ? 'desc' : 'asc',
```
swarajpure, line 127: Now I see why was `isOpen: false` somewhere in other file
Wrong choice of variable name causing confusions 😛

swarajpure (changes requested):
Having some doubts and requesting some changes
Please check

swarajpure:
Closing this in lieu of #328

Issue #270, opened 2021-03-18
Capture sign up time for new user


ankushdharkar:

anuragKhandelwal1:
@ankushdharkar, can you please look at the pr?

ankurnarkhede:
@anuragKhandelwal1 , can you check if there is a firestore functionality to add createdAt and updatedAt timestamps implicitly?

PR #289, opened 2021-03-26
#270 Time stamp added

anuragKhandelwal1:


models/users.js
```diff
@@ -40,6 +40,7 @@ const addOrUpdate = async (userData, userId = null) => {
 
     // Add user
     userData.incompleteUserDetails = true
+    userData.createdOn = Date.now()
```
ankurnarkhede, line 43: @anuragKhandelwal1 , rather than this, can you check if there is any way in FireStore to automatedly add `createdAt` and `updatedAt` timestamps implicitly?

ankurnarkhede (changes requested):

models/users.js
anuragKhandelwal1: Would `firebase.firestore.Timestamp` help? 
(Someone in discord channel said you shouldn't be including firebase just for this.)

anuragKhandelwal1: cc: @swarajpure

swarajpure: We're now using `firebase-admin` package. Can you please check if we can use the implicit method now?

utils/time.js
```diff
@@ -0,0 +1,5 @@
+const getUnixEpochTime = (time = Date.now()) => {
+  return Math.floor(time / 1000)
+}
+
+module.exports = { getUnixEpochTime }
```
swarajpure, line 5: Can you add JSDoc for this function?

swarajpure (changes requested):
JSDoc missing for `getUnixEpochTime`
Also, can you please assert on the tests that we're returning `createdOn` from `GET /users` and `GET /users/:username` and that is a number? (modification in existing integration tests required for `/users`)
Also, if possible, we could also cross verify the value of the `createdOn` field

utils/time.js
anuragKhandelwal1: Do we have some JSDocs available in our project structure?
I couldn't find any.

If not, where should I create it?

swarajpure: Existing code is having JSDoc annotations
Check this link and it'll all make sense 😉 https://jsdoc.app/about-getting-started.html#adding-documentation-comments-to-your-code

anuragKhandelwal1: I feel stupid now.😛

Pushed in new changes.

test/fixtures/user/user.js
```diff
@@ -21,6 +21,7 @@ module.exports = () => {
       isMember: true,
       phone: '1234567890',
       email: 'abc@gmail.com',
+      createdAt: 1621025657,
```
swarajpure, line 24: The values in fixtures are being used by our integration and unit tests to check for the expected behavior.
Can we also have this timestamp coming from the getUnixEpochTime function? We'll be then actually going close to the expected behaviour and our tests would be checking real-life case than having to run tests on hardcoded things.. Then the tests themselves won't make any sense..The tests would be checking on hardcoded values, they'll pass, developers become happy but when it goes to production, something might break because we didn't test on real life scenario, we checked on hardcoded values..

I hope you're getting what I'm trying to say 😅

anuragKhandelwal1: I've to get the time from the created utility method. Correct me if I am wrong.

swarajpure: Yes

utils/time.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Converts Time in milliseconds to Epoch Time
```
swarajpure, line 2: Shouldn't this be `seconds`?

anuragKhandelwal1: We are converting from milliseconds to Epoch (seconds).
`Date.now( )` method returns the number of milliseconds elapsed since January 1, 1970

swarajpure: Got it

swarajpure:
> JSDoc missing for `getUnixEpochTime`
> Also, can you please assert on the tests that we're returning `createdOn` from `GET /users` and `GET /users/:username` and that is a number? (modification in existing integration tests required for `/users`)
> Also, if possible, we could also cross verify the value of the `createdOn` field

@anuragKhandelwal1 can we have modifications in tests as well? We need to assert on the response on above mentioned routes to have `createdOn` field, it should be a **number** and it's value should be equal to the value of `createdOn` field coming from the fixtures as we discussed.

I realise that I'm being too picky but users collection is THE most important collection right now. All the routes are dependent on this one. If at this thing breaks, we'll break all the dependent routes as well. I hope that makes sense :)

utils/time.js
ankushdharkar: Slight clarification here:
Epoch is a particular point in time: 00:00:00:0000 on Jan 1, 1970
Unix timestamp (or Unix time or POSIX time) is the number of seconds since epoch.

Issue #234, opened 2021-03-12
Tests for auctions APIs


swarajpure:

ankurnarkhede:
Reopening because unit tests for the functions are missing.

PremanandChowdhury:
> Reopening because unit tests for the functions are missing.
> @PremanandChowdhury

O yes, the `unit test` is remaining. It will be done post `10th May`
@ankurnarkhede

PR #295, opened 2021-04-03
Checks if user has sufficient items to auction

### LINKED ISSUE #294: Check if the user auctioning something actually has sufficient items

swarajpure:
Fixes #294

### DIFF
--- a/controllers/auction.js
+++ b/controllers/auction.js
@@ -1,4 +1,5 @@
 const auctions = require('../models/auctions')
+const wallet = require('../models/wallets')
 
 /**
  * Fetches all the active (ongoing) auctions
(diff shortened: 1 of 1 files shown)


controllers/auction.js
```diff
@@ -49,6 +50,12 @@ const createNewAuction = async (req, res) => {
   try {
     const { id: seller } = req.userData
     const { initial_price: initialPrice, item_type: itemType, end_time: endTime, quantity } = req.body
+
+    const { currencies } = await wallet.fetchWallet(seller)
+    const itemQuantity = parseInt(currencies[`${itemType}`])
```
swarajpure, line 55: Reason for ``currencies[`${itemType}`]``: https://stackoverflow.com/questions/51273469/how-to-resolve-eslint-generic-object-injection-sink-error

controllers/auction.js
```diff
@@ -49,6 +50,12 @@ const createNewAuction = async (req, res) => {
...
+    if (!itemQuantity) return res.boom.forbidden(`You do not have any ${itemType}!`)
```
ankurnarkhede, line 56: Rather than explicitly showing the user does not have any coins, we can simply say that `You do not have enough of ${itemType}s!` to avoid showing explicit details.

We can return that message for `if(!itemQuantity || itemQuantity < quantity)`

ankurnarkhede (dismissed):
Can you also add tests? 😬

controllers/auction.js
swarajpure: Done.

swarajpure:
> Can you also add tests? grimacing

Ticket assigned to @correaswebert #234

ankurnarkhede (approved):
LGTM

PR #296, opened 2021-04-04
New members API for members page

### LINKED ISSUE #262: GET new members who recently signed up

rajakvk:
Created new PR with new branch

Closes #262 

This PR should only be merged after successful run/merge of this PR #351

### DIFF
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -13,7 +13,7 @@ const getMembers = async (req, res) => {
     const allMembers = await memberQuery.fetchMembers()
 
     return res.json({
-      message: allMembers.length ? 'Members returned successfully!' : 'No member found',
+      message: allMembers.newMembers.length || allMembers.oldMembers.length ? 'Members returned successfully!' : 'No member found',
       members: allMembers
     })
   } catch (error) {
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -33,7 +33,7 @@ const getIdleMembers = async (req, res) => {
   try {
     const allMembers = await memberQuery.fetchMembers()
     const taskParticipants = await tasks.fetchActiveTaskMembers()
-    const idleMembers = allMembers?.filter(({ id }) => !taskParticipants.has(id))
+    const idleMembers = allMembers.oldMembers?.filter(({ id }) => !taskParticipants.has(id))
     const idleMemberUserNames = idleMembers?.map((member) => member.username)
 
     return res.json({
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -13,7 +13,7 @@ const getMembers = async (req, res) => {
     const allMembers = await memberQuery.fetchMembers()
 
     return res.json({
-      message: allMembers.newMembers.length || allMembers.oldMembers.length ? 'Members returned successfully!' : 'No member found',
+      message: allMembers.length ? 'Members returned successfully!' : 'No member found',
       members: allMembers
     })
   } catch (error) {
@@ -33,7 +33,7 @@ const getIdleMembers = async (req, res) => {
   try {
     const allMembers = await memberQuery.fetchMembers()
     const taskParticipants = await tasks.fetchActiveTaskMembers()
-    const idleMembers = allMembers.oldMembers?.filter(({ id }) => !taskParticipants.has(id))
+    const idleMembers = allMembers?.filter(({ id }) => !taskParticipants.has(id))
     const idleMemberUserNames = idleMembers?.map((member) => member.username)
 
     return res.json({
--- a/models/members.js
+++ b/models/members.js
@@ -19,13 +19,19 @@ const fetchMembers = async () => {
 
     if (!snapshot.empty) {
       snapshot.forEach((doc) => {
-        allMembers.push({
+        const memberData = doc.data()
+        const curatedMemberData = {
           id: doc.id,
-          ...doc.data(),
+          ...memberData,
           tokens: undefined,
           phone: undefined,
           email: undefined
-        })
+        }
+        if (typeof memberData.roles === 'undefined') {
+          allMembers.newMembers.push(curatedMemberData)
+        } else if (memberData.roles && memberData.roles.member) {
+          allMembers.oldMembers.push(curatedMemberData)
+        }
       })
     }
 
--- a/models/members.js
+++ b/models/members.js
@@ -27,7 +27,7 @@ const fetchMembers = async () => {
           phone: undefined,
           email: undefined
         }
-        if (typeof memberData.roles === 'undefined') {
+        if (memberData.roles && memberData.roles.member) {
           allMembers.newMembers.push(curatedMemberData)
         } else {
           allMembers.oldMembers.push(curatedMemberData)
--- a/models/members.js
+++ b/models/members.js
@@ -15,7 +15,7 @@ const fetchMembers = async () => {
   try {
     const snapshot = await userModel.where('isMember', '==', true).get()
 
-    const allMembers = []
+    const allMembers = { oldMembers: [], newMembers: [] }
 
     if (!snapshot.empty) {
       snapshot.forEach((doc) => {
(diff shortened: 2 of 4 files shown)


controllers/members.js
```diff
@@ -32,7 +32,7 @@ const getMembers = async (req, res) => {
 const getIdleMembers = async (req, res) => {
   try {
     let allMemberUsernames = await memberQuery.fetchMembers()
-    allMemberUsernames = allMemberUsernames.map(member => member.username)
+    allMemberUsernames = allMemberUsernames.oldMembers.map(member => member.username)
```
swarajpure, line 35: This is to find the idle members only from old members, right? Just confirming.

models/members.js
```diff
@@ -15,16 +15,32 @@ const fetchMembers = async () => {
...
+      switch (memberData.userType) {
+        case 'new':
+          allMembers.newMembers.push(curatedMemberData)
+          break
+        case 'blocked':
+          // intentionally left blank for future use case
+          break
+        case 'member':
+          // intentionally left break statement to fall through default case
+          // so that doc with no userType field considered as members
```
swarajpure, line 38: I get what is `new` and `blocked` are meant to do, but what is `member` doing? Didn't get that, can you please elaborate a little bit?

test/integration/members.test.js
```diff
@@ -44,8 +45,9 @@ describe('Members', function () {
           expect(res).to.have.status(200)
           expect(res.body).to.be.a('object')
           expect(res.body.message).to.equal('Members returned successfully!')
-          expect(res.body.members).to.be.a('array')
-          expect(res.body.members[0].isMember).to.eql(true)
+          expect(res.body.members.newMembers).to.be.a('array')
+          expect(res.body.members.oldMembers).to.be.a('array')
+          expect(res.body.members.oldMembers[0].isMember).to.eql(true)
```
swarajpure, line 50: `newMembers` would also be having `isMember` true, right?

swarajpure (commented):
Having some minor queries, please check.

controllers/members.js
rajakvk: Basically the structure of the returned json has been changed from allMembers = [{},{},{},...] to allMembers = {newMembers: [{},{},{},...], oldMembers: [{},{},{},...]}.

So I believe, we are considering idle members from oldMembers not from newMembers. That is my take. 

We should confirm this assumption with AO. @sumitd94 

In fact, Sumit suggested this json structure change in his previous review.

models/members.js
rajakvk: TDLR; member case included to handle future scenario, not necessary right now.

IsMember is existing field. userType newly introduced field along with this PR. We need to discuss with AO, @sumitd94 
1. what to do with these two inter-related fields?
2. are we going to keep both the fields?
3. are we going to remove isMember field?
4. If we remove isMember, how to deal with related code changes?
5. If we remove isMemebr, how to deal with existing data?

test/integration/members.test.js
rajakvk: Right now 'Yes'. 
All are members, plus their userType value would be 'new'.

swarajpure: Can we assert on that as well?

controllers/members.js
swarajpure: I'm also assuming the same. @sumitd94 Please confirm.

sumitd94: Yes, idle members should be returned from the list of `old-members` (here by old-members we are referring to all those users who has `isMember` set to `true`

test/integration/members.test.js
rajakvk: This needs discussion based on my questions raised above.

models/members.js
sumitd94: What do you suggest here?

sumitd94: According to me

1. For now let's keep both, userType will contain the values like `new, blocked, member`. here the `member` is same as `isMember = true`
2. For now yes
3. In future yes
4. Need to think, open to suggestion
5. Need to think, open to suggestion

swarajpure: I still didn't get why do we need `member` in `userType` if we already have `isMember` 😅

rajakvk: @swarajpure, if we decide to keep isMember field forever then 'member' in 'userType' is not necessary. But that decision is not yet taken.

swarajpure: @sumitd94 please unblock @rajakvk

models/members.js
```diff
@@ -13,21 +13,26 @@ const userModel = firestore.collection('users')
...
           email: undefined
-        })
-      }
-      )
+        }
+        if (typeof memberData.roles === 'undefined') {
+          allMembers.newMembers.push(curatedMemberData)
+        } else if (memberData.roles && memberData.roles.member) {
+          allMembers.oldMembers.push(curatedMemberData)
+        }
```
sumitd94, line 34: I think we can re-structure this to something like this

```javascript
if (memberData.roles && memberData.roles.member) {
allmembers.oldMembers.push(curatedMemberData)
} else {
...new members
}
```

sumitd94 (changes requested):
One small change, please.

models/members.js
rajakvk: Done with the update.

rajakvk:
Integration test failing when executing /members/idle related tests due to changes in 'tasks' end point changes. I would suggest a discussion needed to proceed further. Also @AO take a call on how to deal with inter-related test cases. @ankurnarkhede @sumitd94 @ankushdharkar

ankurnarkhede:
Will get the PR https://github.com/Real-Dev-Squad/website-backend/pull/296 merged ASAP to resolve the failing CI, and thus merge this.

ankurnarkhede:
@rajakvk , is this PR blocked on any other PR? As per I can see, the CI is passing for develop. The tests for this branch are throwing 500, I suppose this could be because of an incorrectly handled exception. 🤔

ankurnarkhede (changes requested):
@rajakvk  , is this PR blocked on any other PR? As per I can see, the CI is passing for develop. The tests for this branch are throwing 500, I suppose this could be because of an incorrectly handled exception. 🤔

models/members.js
```diff
@@ -13,21 +13,26 @@ const userModel = firestore.collection('users')
...
+          ...memberData,
           tokens: undefined,
           phone: undefined,
           email: undefined
-        })
-      }
-      )
+        }
+        if (memberData.roles && memberData.roles.member) {
+          allMembers.newMembers.push(curatedMemberData)
```
harshith-venkatesh, line 31: I believe this should be oldMembers

ankurnarkhede:
As discussed, can we simply add `isNewMember = <boolean>` (or any other nice name) in the user response object. This will also maintain the existing contract between API and UI and thus won't break things.

sumitd94 (dismissed):
Looks perfect, Thank you so much 🥇

models/members.js
```diff
@@ -13,21 +13,27 @@ const userModel = firestore.collection('users')
...
           email: undefined
-        })
-      }
-      )
+        }
+        if (memberData.roles && memberData.roles.member) {
+          allMembers.isMember = true
+        } else {
+          allMembers.isMember = false
+        }
```
ankushdharkar, line 34: Nit:
```
          allMembers.isMember = !!(memberData.roles && memberData.roles.member)
```

models/members.js
```diff
@@ -13,21 +13,27 @@ const userModel = firestore.collection('users')
 
 const fetchMembers = async () => {
```
ankushdharkar, line 14: Should this be called `fetchUsers` now? Given that all members are users, but all users are not members

ankushdharkar (changes requested):

models/members.js
sumitd94: shouldn't this be `curatedMemberData.isMember`?

sumitd94 (approved):

PR #299, opened 2021-04-05
Updated Contributing.md

anuragKhandelwal1:
Updated the contributing.md file specifying the cookie name in for dev env.


CONTRIBUTING.md
```diff
@@ -93,7 +93,7 @@ The following project structure should be followed:
 - Run the project locally, make sure the server is listening to requests
 - Navigate to `https://github.com/login/oauth/authorize?client_id=<GITHUB_CLIENT_ID>`
 - Authorize the application
-- Copy the response cookie named rds-session from the redirected request <HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>
+- Copy the response cookie named `rds-session` (`rds-session-development` in development mode) from the redirected request <HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>
```
swarajpure, line 96: We don't need to monitor the network tab now for any specific request. We could directly mention that once authorized, they have to go to the cookies tab of their browser and pick up the suitable cookie for that particular domain (remember, after authorizing the app, the user will be taken to `localhost:3000/healthcheck` so they'll be able to see the cookie for localhost directly)

Let me know your thoughts 😉

anuragKhandelwal1: What if someone is using postman for setting the cookie value? 
If the person uses `rds-session` at the time of development, things wouldn't work smooth.

Isn't it?

swarajpure: Yes. you're absolutely right, but these steps are not for `setting` the cookie, but for `getting` the cookie

And to get the cookie, they'll most probably be doing it from their browser. Network tabs are a bit confusing for any newbie, so let's avoid that for now. We could ask them to check the cookies tab instead, much simpler and cleaner to look at.
Does that make sense?

anuragKhandelwal1: Yes.
But I still think the setting up of cookie should also be mentioned somewhere.
This might help us in the long term.
What's your opinion?

swarajpure: Yes, I'm also talking about the same 😅 
We are discussing on how to make it easier for the users to get the cookie

There are two ways to get the cookie for development purposes:
- Keep monitoring the network tab for the particular request `<HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>` amongst other requests and then go to the response headers and then the cookie from there
- Once authorized, simply go to the network tab of your browser and boom! There's your 🍪!

Which one do you think is easier for a newcomer to follow? 😉

anuragKhandelwal1: Wouldn't even a newcomer know about the environment it is working on?
Even if we work on the edge case, asking to check the network tab shouldn't be a complicated task. Or would it be?

swarajpure: I've got on call with 11 devs in RDS to help them with setting up the backend 😅 

I've seen the pain points where new people get stuck and hence I was saying that we better go with the cookies tab than network tab. Network tab also has to be given one more option: `preserve logs`, if you miss checking that, boom. The request you were looking for will come in front of your eyes and vanish 😅 

Why to add more friction points, is what I'm saying

anuragKhandelwal1: In that case, should I document this down separately on the ReadMe.md file?
Something like - For sending cookie in request, use `rds-session-development` as cookie name. (Obviously explaining it a little over there )

anuragKhandelwal1: Also, AFAIK, in swagger the authorization tells the key name itself. Should I mention 'trying api in postman' only?

swarajpure: We're just concerned about getting the cookie here, right? Now it's upto the developer where to use it. We just tell them, `rds-session-development` is the cookie of your concern

anuragKhandelwal1: I still think the commit here makes sense, `Copy the response cookie named rds-session from the redirected request <HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>`. Here it's written the key name would be rds-session, which is not true always.

swarajpure: Can you please read the text in the screenshot once again? We're NOT going with the network tab and any request from there.

anuragKhandelwal1: By this the person working on this would get to know about the cookie name in dev mode.

swarajpure: I'm not talking about the cookie name at all -- you're right about it

I'm saying to discard the network tab and the request thing altogether, just mention to check the cookies tab for `rds-session-development` after authentication. Simple.

anuragKhandelwal1: I guess we are not on the same page here.
Let me state my perspective on this and please correct me if I am going wrong.

- According to the current docs, we are talking about generating a cookie (whose name would be varied in accordance to the env)
- Our objective is to let the devs know how you can set/send **_the cookie_** along with the network with the correct cookie name.

anuragKhandelwal1: > I'm not talking about the cookie name at all -- you're right about it
> 
> I'm saying to discard the network tab and the request thing altogether, just mention to check the cookies tab for `rds-session-development` after authentication. Simple.

I was correct, we weren't on the same page. But, we are now.
Thanks.

anuragKhandelwal1: @swarajpure , pushed in new changes.

CONTRIBUTING.md
```diff
@@ -93,11 +93,13 @@ The following project structure should be followed:
...
+- Once authorized, simply go to the Application tab of your browser's developer tools and boom! There's your 🍪.
```
swarajpure, line 97: Isn't this confusing? We should be including **either** of these lines. Both of them are two different ways to achieve the same thing. Please keep the second line only.

We could mention about the cookie name in the second line and keep that one only 🙂 
Also, please phrase it in proper words. Inside application tab, inside the cookies section for localhost, there is the cookie. In fact, the `application` tab is valid only for chrome. The cookies in firefox are in `storage -> cookies`

anuragKhandelwal1: Pushed in new changes.

swarajpure: Please remove line 96 😅

anuragKhandelwal1: Oopsie! Pushed again.

CONTRIBUTING.md
```diff
@@ -93,11 +93,12 @@ The following project structure should be followed:
 - Run the project locally, make sure the server is listening to requests
 - Navigate to `https://github.com/login/oauth/authorize?client_id=<GITHUB_CLIENT_ID>`
 - Authorize the application
-- Copy the response cookie named rds-session from the redirected request <HOST>/auth/github/callback?code=<SOME_AUTOGENERATED_CODE>
+- Once authorized, check browser's cookies section. There's your 🍪 named `rds-session` (`rds-session-development` in development mode).
```
swarajpure, line 96: > Once authorized, check browser's cookies section. Copy the value of the cookie named `rds-session` (`rds-session-development` for development mode).

This fits well with the grammar, language and style of existing documentation. What do you think?

anuragKhandelwal1: Sounds good to me.

swarajpure (approved):
Awesome. Thanks for your patience with me 😛

PR #301, opened 2021-04-06
Add firebase emulator config to allow using emulator in test environment

rashid301:


rashid301:
Not required. The test-config script is supposed to test the config of remote services. It is not meant to work with emulator.

PR #304, opened 2021-04-08
Add unit tests for utils/users.js

sanyogita14:
Unit tests added for `getUsername` , `getUserId`, `getParticipantUsernames` and `getParticipantUserIds` functions in utils/users.js

### DIFF
--- a/test/unit/utils/users.test.js
+++ b/test/unit/utils/users.test.js
@@ -19,9 +19,12 @@ const tasks = require('../../../models/tasks')
 
 describe('users', function () {
   let userId, taskId
+  let username, userData
   let task, taskData
   beforeEach(async function () {
     userId = await addUser()
+    userData = userDataArray[0]
+    username = userData.username
 
     taskData = {
       title: 'Test task',
@@ -56,25 +59,20 @@ describe('users', function () {
 
   describe('getUsername', function () {
     it('should receive userId of user from database and return username', async function () {
-      const userData = userDataArray[0]
-      const username = await usersUtils.getUsername(userId)
-      expect(username).to.equal(userData.username)
+      const convertedUsername = await usersUtils.getUsername(userId)
+      expect(convertedUsername).to.equal(username)
     })
   })
 
   describe('getUserId', function () {
     it('should receive username of user and return userId', async function () {
-      const userData1 = userDataArray[0]
-      const username1 = userData1.username
-      const userId1 = await usersUtils.getUserId(username1)
-      expect(userId1).to.equal(userId)
+      const convertedUserId = await usersUtils.getUserId(username)
+      expect(convertedUserId).to.equal(userId)
     })
   })
 
   describe('getParticipantUsernames', function () {
     it('should receive userId of users from database and return their usernames', async function () {
-      const userData = userDataArray[0]
-      const username = userData.username
       const participantArray = task.taskData.participants
       const participantUsername = await usersUtils.getParticipantUsernames(participantArray)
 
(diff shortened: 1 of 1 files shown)


test/unit/utils/users.test.js
```diff
@@ -0,0 +1,92 @@
...
+      const username = await usersUtils.getUsername(userId)
+      expect(username).to.equal(userData.username)
+    })
+  })
+
+  describe('getUserId', function () {
+    it('should receive username of user and return userId', async function () {
+      const userData1 = userDataArray[0]
+      const username1 = userData1.username
+      const userId1 = await usersUtils.getUserId(username1)
```
swarajpure, line 69: Why `userData1` and `username1`? 🤔 
`const` won't anyway be accessible from outside this `it` block

test/unit/utils/users.test.js
```diff
@@ -0,0 +1,92 @@
...
+
+      expect(participantUsername).to.include(username)
+    })
+  })
+
+  describe('getParticipantUserIds', function () {
+    it('should receive usernames of users from participant array and return their userIds', async function () {
+      const participantArray = taskData.participants
+      const participantUserId = await usersUtils.getParticipantUserIds(participantArray)
+      expect(participantUserId).to.include(userId)
```
swarajpure, line 89: If we need `username` everytime, we could have it declared at the top in `beforeEach`

swarajpure (changes requested):
Thanks for tests! 🔥 
Requesting some changes, please check!

test/unit/utils/users.test.js
sanyogita14: I wanted to differentiate between the two variables while checking the condition.

test/unit/utils/users.test.js
sanyogita14: Yes. I'll do that

test/unit/utils/users.test.js
```diff
@@ -0,0 +1,90 @@
...
+const tasks = require('../../../models/tasks')
+
+/**
+ * Test the utils functions and validate the data returned
+ */
+
+describe('users', function () {
+  let userId, taskId
+  let username, userData
+  let task, taskData
```
swarajpure, line 23: See `addUser` is being called before every test (beforeEach) and it is returning `userId` which will be unique every single time, so it makes sense to have it declared as `let`, same for `task` and `taskId`

Whereas `username`, `userData`, `taskData` will remain constant throught -- so they can be declared as `const`

Does that make sense? 🤔

sanyogita14: Yes I got your point👍🏻

test/unit/utils/users.test.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * This eslint rule is disabled because of https://github.com/nodesecurity/eslint-plugin-security/issues/21
+ * It gives linting errors in testing the DB data with keys from fixtures
+ */
+/* eslint-disable security/detect-object-injection */
```
swarajpure, line 5: On which lines were you getting the error?
Since it says `security`, let us not avoid it. We should try to fix it.

sanyogita14: Yes ok. I'll re-check where the error is occurring

sanyogita14: > On which lines were you getting the error?
> Since it says `security`, let us not avoid it. We should try to fix it.

I checked the files. Seems to disable the eslint-rule was unnecessary as the code is working fine without it.  I must have encountered the error earlier but forget to check if it has been resolved.

swarajpure (approved):
Looks good!
Thanks for the tests! 😄

PR #312, opened 2021-04-10
Remove firestore-private-key.json usage

### LINKED ISSUE #311: Read firestore config data from env variable
In order to deploy, it's a good practice to use env variables for keeping sensitive data like keys away from the actual code.

But, in case of firestore, the service file is a json file and not a simple string. So, we create an env variable `FIRESTORE_CONFIG` that contains our json as a string, and in our code, we parse the string to create a js object representing the JSON

We have to be careful to remember to escape the newline characters in the JSON string in order have the newlines show up in the JS object.

This will let us remove having to create a `firestore-private-key.json` file and manage it for keys.

Now, in package.json, we can fix our `npm start` command:
\- `   "start": "echo $FIRESTORE_CONFIG_DATA > firestore-private-key.json && node server.js",`
+`   "start": "node server.js",`

ankushdharkar:
We will no longer need to deal with `firestore-private-key.json`

The existing tests will most likely break, because we are changing how we create our firestore admin object.
Updated the tests for the same, please check.

In order to deploy, it's a good practice to use env variables for keeping sensitive data like keys away from the actual code.
This will let us remove having to create a firestore-private-key.json file and manage it for keys.

Now, package.json can have these changes:
\- `"start": "echo $FIRESTORE_CONFIG_DATA > firestore-private-key.json && node server.js",`
\+ `"start": "node server.js",`

Closes #311

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,8 +123,5 @@ dist
 # VSCode files
 .vscode
 
-# Firestore key file
-firestore-private-key.json
-
 # Local config file
 config/local.js
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,7 +82,6 @@ The following project structure should be followed:
     |-- README.md
     |-- CHANGELOG.md
     |-- app.js
-    |-- firestore-private-key.json // Firestore key file
     |-- package-lock.json
     |-- package.json
     |-- server.js // Contains server start logic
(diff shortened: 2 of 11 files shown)


ankurnarkhede (changes requested):
Implementation LGTM ✅ 
Tests need to be fixed.

ankushdharkar:
@ankurnarkhede Running `npm test` locally gives me this:
image

What else should we fix for tests?

ankurnarkhede (commented):
LGTM ✅

ankurnarkhede (approved):

PR #317, opened 2021-04-12
Rename the test-config script to validate-setup

bhk31:


scripts/validateSetup.js
```diff
@@ -2,10 +2,7 @@ const Firestore = require('../utils/firestore.js')
...
 
 // check whether github credentials are not falsy
-if (
-  config.githubOauth.clientId === '<clientId>' ||
-  config.githubOauth.clientSecret === '<clientSecret>'
-) {
+if (config.githubOauth.clientId === '<clientId>' || config.githubOauth.clientSecret === '<clientSecret>') {
   throw new Error('Github credentials are not set.')
 } else {
   // eslint-disable-next-line no-console
```
swarajpure, line 8: Instead of disabling linting errors for `console.log` for each line, would be better if disable it for the whole file at the top. It would help in reading the code easily
Refer: https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments

swarajpure (changes requested):
Just one change and I guess we're good to go.

Also, it seems that your formatter (prettier probably) has added many extra lines in .md files, can you please check?

swarajpure (approved):
Thanks for this one! 🎉

Issue #323, opened 2021-04-13
[Dev Spec] Restrict user actions on backend


Cartmanishere:
## Requirements
- We need a way to restrict a user's action on the backend.
- If it is a bad user, we will restrict the create/edit actions for the user.
- He/she will be able to view all the data but not be able to edit anything.

## Schema

In the `users` schema, we need to add a key/flag which where we can store whether the user is `restricted`.

There are two ways to go about this -- 

#### 1. Use the `roles` 
- We are planning to add `roles` key in the users schema which will define the user's permissions for e.g -- 
```
{
    roles: {
        member: true,
        admin: true
    }
}
```
- We make use of this `roles` map and add `restricted` as one of the roles. for e.g `restricted: true`. 

#### 2. Add a top-level field in the users schema -- 
- Add a key in the user schema which tells us whether the user is restricted -- 
```
{
    ...
    restricted: true
}
```
- [x]  Decide where we should store the flag in the users schema.

## Implementation
- Since this is application-wide validation, it makes sense to implement it as a middleware.
- We can only check for logged-in users, so this `limitAccess` middleware needs to come after the `authenticate` middleware.
- The devs working on adding new routes should not have to explicitly add this middleware while declaring the route middlewares.
- This middleware can be implicitly added after the `authenticate` middleware. So anytime `authenticate` is done, then this check is also performed. 
- Please check linked PR for what I mean by above.

RFC:
- @ankurnarkhede 
- @ankushdharkar 
- @whyDontI

ankurnarkhede:
Yup, this makes sense. We can rely on this mechanism for authorization since we do not have to manage numerous entities and user groups.

The user's roles will be populated in the `authenticated` middleware and can be used further.
@Cartmanishere

Cartmanishere:
@ankurnarkhede Ack. For this feature, I have decided that we can use the `roles` approach for where to store the `restricted` key. I will put a cleaned up PR for this.

Also for authorization, we can use the same `roles` key in User schema to add more authorization (for e.g for superuser, appOwners, etc). I can pick up that task after merging this.

Cartmanishere:
- [ ] Need to update the `user` schema in the `website-data-models` to reflect the addition of field `roles`. cc: @rajakvk 

```
{
    ....
    roles: {
        restricted: true/false
    }
}
```

PR #324, opened 2021-04-13
Add middleware to check restricted user

Cartmanishere:
Closes: #323

### DIFF
--- a/middlewares/authenticate.js
+++ b/middlewares/authenticate.js
@@ -2,6 +2,31 @@ const authService = require('../services/authService')
 const users = require('../models/users')
 
 /**
+ * Middleware to check if the user has been restricted. If user is restricted,
+ * then only allow read requests and do not allow to any edit/create requests.
+ *
+ * Note: This requires that user is authenticated hence must be called after
+ * the user authentication middleware. We are calling it from within the
+ * `authenticate` middleware itself to avoid explicitly adding this middleware
+ * while defining routes.
+ *
+ * @param {Object} req - Express request object
+ * @param {Object} res - Express response object
+ * @param {Function} next - Express middleware function
+ * @returns {Object} - Returns unauthorized object if user has been restricted.
+ */
+const checkRestricted = async (req, res, next) => {
+  const { roles } = req.userData
+  if (
+    roles &&
+    roles.restricted &&
+    req.method !== 'GET') {
+    return res.boom.forbidden('You are restricted from performing this action')
+  }
+  return next()
+}
+
+/**
  * Middleware to validate the authenticated routes
  * 1] Verifies the token and adds user info to `req.userData` for further use
  * 2] In case of JWT expiry, adds a new JWT to the response if `currTime - tokenInitialisationTime <= refreshTtl`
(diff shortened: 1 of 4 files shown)


middlewares/authenticate.js
```diff
@@ -1,6 +1,17 @@
 const authService = require('../services/authService')
 const users = require('../models/users')
 
+/**
+ * Middleware to check if access has been restricted for this user.
+ */
+const checkLimitedAccess = async (req, res, next) => {
+  const { restricted } = req.userData
```
swarajpure, line 8: Is this `restricted` field going to be on the root level of the user data itself or inside an object like `status`? 🤔

middlewares/authenticate.js
```diff
@@ -35,7 +56,7 @@ module.exports = async (req, res, next) => {
     const userData = await users.fetchUser({ userId })
     req.userData = userData.user
 
-    return next()
+    return checkRestricted(req, res, next)
```
ankurnarkhede, line 59: IMO, the authenticate function shouldn't be checking the restricted users, it should just check the auth. 🤔
Is there any way, we can add the `checkRestricted` as a global middleware, so that it will apply to every request?

ankurnarkhede (changes requested):

test/integration/restricted.test.js
```diff
@@ -0,0 +1,75 @@
...
+const userData = require('../fixtures/user/user')()
+const addUser = require('../utils/addUser')
+
+const cookieName = config.get('userToken.cookieName')
+const unrestrictedUser = userData[0]
+const restrictedUser = userData[1]
+
+chai.use(chaiHttp)
+
+describe.only('checkRestrictedUser', function () {
```
ankurnarkhede, line 17: .only needs to be removed.

ankurnarkhede (changes requested):

test/fixtures/user/user.js
```diff
@@ -35,7 +35,10 @@ module.exports = () => {
       linkedin_id: 'nikhil-bhandarkar',
       twitter_id: 'whatifi',
       phone: '1234567891',
-      email: 'abc1@gmail.com'
+      email: 'abc1@gmail.com',
```
ankurnarkhede, line 38: Can we create another user object here?
Adding restricted to this user will affect the other non-GET requests where this is being used.

ankurnarkhede (changes requested):

middlewares/authenticate.js
Cartmanishere: No. It cannot be done that way because to check for `restricted` this middleware needs to come after the authenticate middleware.

Currently we add `authenticate` middleware explicitly while adding any new route. So this middleware would have to be added explicitly after each location where `authenticate` is used.

To avoid this repetition, I have called this middleware from within the authenticate middleware itself.

test/fixtures/user/user.js
Cartmanishere: @ankurnarkhede Good point. I will just add one more user.

ankurnarkhede (approved):

PR #332, opened 2021-04-18
Fixed epoch datatype in tasks and tests #331

### LINKED ISSUE #331: Fix tasks API
startedOn and endsOn should be an epoch
Fix optional values in creating and updating tasks

ankushdharkar:
Closes #331 
Epoch type should be a number in tasks APIs


swarajpure (approved):
LGTM! 👍🏻

PR #334, opened 2021-04-19
fetch idle members based on assignee field in task

harshith-venkatesh:


models/tasks.js
```diff
@@ -67,13 +67,13 @@ const fetchTasks = async () => {
...
-    const tasksSnapshot = await tasksModel.where('status', '==', 'active').get()
+    const tasksSnapshot = await tasksModel.where('type', '==', 'feature').where('status', '==', 'active').get()
     const activeMembers = []
     tasksSnapshot.forEach((task) => {
       const taskData = task.data()
-      if (taskData.participants) {
+      if (taskData.assignee) {
         activeMembers.push(
-          ...taskData.participants
+          taskData.assignee
```
swarajpure, line 76: This will then return those members for which assigned task status is `pending` 🤔 
Is it expected?

harshith-venkatesh: This wont be returning assignees with assigned status pending as there is extra condition to check for status active

swarajpure: That's what I'm saying 😅
An user can have only 1 active task
So let's see they forget to change the status of one of their pending tasks to active, they'll be shown idle, right?

Or if there's someone completed a task and if they have some task in pending and they forget to mark it as active, they'll be shown idle, right?

harshith-venkatesh: Yeah I had a similar discussion, the way forward we need to make sure what ever the task we are working currently we need to mark it as active to keep the active track of all the progress ahead and assign to others accordingly if there are any issues

harshith-venkatesh: Yeah you are right , they will be shown idle and some one would be asking them in the group leading them to mark it active

swarajpure: The user can mark their tasks as `active` from http://my.realdevsquad.com/tasks 👍🏻

swarajpure (dismissed):
Looks good to me!

services/contributions.js
```diff
@@ -30,15 +30,17 @@ const getUserContributions = async (username) => {
...
-      for (const userId of task.participants) {
-        const username = await userUtils.getUsername(userId)
-        const userDetails = participantsDetailsMap.get(username)
-        if (userDetails) {
-          participantsDetails.push(userDetails)
-        } else {
-          const user = await getUserDetails(username)
-          participantsDetailsMap.set(username, user)
-          participantsDetails.push(user)
+      if (Array.isArray(task.participants)) {
```
ankushdharkar, line 33: We are creating an extra `if` block to increase the complexity of a larger piece of code.

This may be more predicatble:

```js
const { participants = [] } = task
for (participants) { ... } 
```

services/contributions.js
```diff
@@ -30,15 +30,17 @@ const getUserContributions = async (username) => {
...
+        for (const userId of task.participants) {
+          const username = await userUtils.getUsername(userId)
```
ankushdharkar, line 35: Are we going to await for every single fetch call? Can we not _parallelize_ this?

services/contributions.js
```diff
@@ -30,15 +30,17 @@ const getUserContributions = async (username) => {
...
+          const userDetails = participantsDetailsMap.get(username)
+          if (userDetails) {
+            participantsDetails.push(userDetails)
+          } else {
+            const user = await getUserDetails(username)
```
ankushdharkar, line 40: Similar to above: `parallelize`?

utils/tasks.js
```diff
@@ -13,7 +13,7 @@ const fromFirestoreData = async (task) => {
     assignee = await getUsername(assignee)
   }
 
-  if (participants) {
+  if (Array.isArray(participants)) {
```
ankushdharkar, line 16: Re: above

ankushdharkar (changes requested):

models/tasks.js
```diff
@@ -67,13 +68,13 @@ const fetchTasks = async () => {
 
 const fetchActiveTaskMembers = async () => {
   try {
-    const tasksSnapshot = await tasksModel.where('status', '==', 'active').get()
+    const tasksSnapshot = await tasksModel.where('type', '==', TASK_TYPE.FEATURE).where('status', '==', TASK_STATUS.ACTIVE).get()
     const activeMembers = []
     tasksSnapshot.forEach((task) => {
       const taskData = task.data()
-      if (taskData.participants) {
```
harshith-venkatesh, line 74: Instead of extra if condition , I will be calling ```const {assignee} = task.data()``` here...
Would there be a possibility of occurrence of undefined in this case?

services/contributions.js
harshith-venkatesh: Sorry for the trouble, The changes are not from the current PR

services/contributions.js
harshith-venkatesh: Sorry for the trouble, The changes are not from the current PR

utils/tasks.js
harshith-venkatesh: Sorry for the trouble, The changes are not from the current PR

harshith-venkatesh: @rashid301 , this is something you might find it useful

models/tasks.js
ankushdharkar: What do the docs say?

harshith-venkatesh: https://firebase.google.com/docs/reference/js/firebase.firestore.DocumentSnapshot#data
If the document doesn't exist it returns undefined 
We can push undefined to array (confirmed in mdn docs example).
Would it be better to stick to current plan when if the assignees exists then they are being added?

ankushdharkar: What's an alternative you suggest?

ankushdharkar (commented):
Why did we not go along the route of [Not Equal](https://cloud.google.com/firestore/docs/query-data/queries#not_equal)?

ankushdharkar (approved):
Looks good to go! Thanks 🎉

Issue #327, opened 2021-04-14
Route : `tasks/:username` for getting tasks of user


sanyogita14:
Create new route `tasks/:username` for getting all tasks of a particular user when `username` provided as a parameter

bhk31:
I will work on this issue. Please assign it to me.

bhk31:
@swarajpure Are we good to close this?

PR #338, opened 2021-04-21
Members active tasks API

bhk31:
PR for API contract [click here](https://github.com/Real-Dev-Squad/website-api-contracts/pull/50)


controllers/tasks.js
```diff
@@ -43,6 +43,24 @@ const fetchTasks = async (req, res) => {
...
   }
 }
+
+const getTaskByUser = async (req, res) => {
+  try {
+    const { username } = req.params
+
+    if (username) {
+      const allTasks = await tasks.fetchAllTaskOfUser(username)
+      return res.json(allTasks)
```
sumitd94, line 53: let's restructure this response like
```javascript
return res.json({ message: 'tasks returned successfully', allTasks })
```

sumitd94 (changes requested):
Few changes requested please

models/tasks.js
```diff
@@ -127,8 +128,11 @@ const fetchUserTasks = async (username, statuses = []) => {
         .where('status', 'in', statuses)
         .get()
     } else {
-      tasksSnapshot = await tasksModel.where('participants', 'array-contains', userId)
-        .get()
+      if (hasUser === false) {
+        tasksSnapshot = await tasksModel.where('participants', 'array-contains', userId)
```
swarajpure, line 132: What is `hasUser`? 🤔

controllers/tasks.js
```diff
@@ -43,6 +43,27 @@ const fetchTasks = async (req, res) => {
     return res.boom.badImplementation('An internal server error occurred')
   }
 }
+
+const getUserTasks = async (req, res) => {
+  try {
+    const { username } = req.params
+
+    if (username) {
+      const allTasks = await tasks.fetchUserTasks(username, [], true)
```
swarajpure, line 52: Once we take `username` from query, we should find `userId` corresponding to the given `username` using existing functions from `utils/users.js`.

We are storing `userId` in DB and not `username`

routes/tasks.js
```diff
@@ -126,4 +126,6 @@ router.post('/', authenticate, authorizeOwner, createTask, tasks.addNewTask)
  */
 router.patch('/:id', authenticate, authorizeOwner, updateTask, tasks.updateTask)
 
+router.get('/:username', authenticate, tasks.getUserTasks)
```
swarajpure, line 129: Please add swagger docs. You can check other routes for more details

controllers/tasks.js
```diff
@@ -43,6 +43,27 @@ const fetchTasks = async (req, res) => {
     return res.boom.badImplementation('An internal server error occurred')
   }
 }
+
+const getUserTasks = async (req, res) => {
+  try {
```
swarajpure, line 48: JSDoc missing, please add

swarajpure (changes requested):
Requesting changes, please check. Also having some minor queries.

Welcome to backend 🎉 
Thank you! 😄

controllers/tasks.js
bhk31: We are fetching `userId`  in `fetchUserTasks` function. It was existing function.

models/tasks.js
bhk31: `fetchUserTasks` this function is used by multiple function Ex. `getSelfTasks` So added a parameter to that function just to identify the function is being called from API of `task:username`. As the API gives us the user so I added a parameter named `hasUser`.

swarajpure: I don't think it's required.  Even for getting self tasks, we can use `fetchUserTasks`. At the end it's job is to find tasks for a particular user. Getting self tasks would be a special case but we did not modify the function. What do you think?

controllers/tasks.js
swarajpure: No. You're passing `username` from here, how will the `fetchUserTasks` function know the corresponding userId accordingly by itself? 😅 
Just to make it clear, my `username` would be `swaraj` and my userId would be some random string generated by firestore and it would be unique and it won't change any time soon. My username may change, but the userId won't and hence we're using `userId` in the DB to store things.

controllers/tasks.js
```diff
@@ -43,6 +43,33 @@ const fetchTasks = async (req, res) => {
...
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const getUserTasks = async (req, res) => {
+  try {
+    const { username } = req.params
+
+    if (username) {
+      const allTasks = await tasks.fetchUserTasks(username, [], true)
```
swarajpure, line 58: We need to handle the case when the `username` provided is wrong. We should be returning 404 if I pass `obama` as username.
Please check this controller function:
https://github.com/Real-Dev-Squad/website-backend/blob/c32f177d61ccce9890b02d662be5c0031c20e71a/models/auctions.js#L146

And this model function:
https://github.com/Real-Dev-Squad/website-backend/blob/c32f177d61ccce9890b02d662be5c0031c20e71a/models/auctions.js#L146

By these two example functions, you'll get how to handle the wrong username case. Otherwise we'd get 5xx error on production which isn't something we want.

models/tasks.js
swarajpure: What the purpose of identifying from where the function was called? The model function's only is do database calls and return the requested data, it doesn't worry about who's calling those functions and from where. Let me know you think 😉

bhk31: Actually as per the discussion I had with @sumitd94 He said we need the tasks by assignee only. In self tasks it fetches tasks where user is in participant. If we want to include the taskd where user is in participant then we don't need `hasUser` We can move ahead with current implementation.

routes/tasks.js
```diff
@@ -126,4 +126,29 @@ router.post('/', authenticate, authorizeOwner, createTask, tasks.addNewTask)
...
+ *             schema:
+ *               $ref: '#/components/schemas/tasks'
+ *       500:
+ *         description: badImplementation
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/errors/badImplementation'
+ */
+router.get('/:username', authenticate, tasks.getUserTasks)
```
swarajpure, line 152: This route will be used on the members page. It should NOT require authentication. If a recruiter comes to your page, do you expect them to be logged in, just to be able to view your tasks?

swarajpure (changes requested):
Please handle the case when username provided is wrong. We don't want to see 5xx errors on production just because we didn't handle a case 😅 

Also, please remove the authentication from this route. This is going to be a public route.

Good job at your first PR in backend 🥳

models/tasks.js
swarajpure: Now I get what you're trying to say
@sumitd94 need some clarifications here. Group tasks would have `participants` list, right? Don't we want those group tasks to show up on a members' page? I'm getting a bit confused here 😅 
Like the `/tasks` page on my-site, @sumitd94 and I made that page, shouldn't it show up on both of our contributions? 🤔 

cc: @ankushdharkar

routes/tasks.js
bhk31: I was not aware of where we will integrate this API in frontend. But yeah we don't need authentication. Will remove it.

models/tasks.js
```diff
@@ -112,10 +112,11 @@ const fetchTask = async (taskId) => {
  * @return {Promise<tasks|Array>}
  */
 
-const fetchUserTasks = async (username, statuses = []) => {
+const fetchUserTasks = async (username, statuses = [], hasUser = false) => {
   try {
     const { user } = await fetchUser({ username })
```
bhk31, line 117: A small suggestion @swarajpure and @sumitd94. Do we really need to fetch user object ? As we are using username entity of that object in whole function which we already have. We can directly call `userUtils.getUserId(username)` Let me  know if this is helpful/valid.

bhk31:
One more thing as I can see @ankushdharkar changed the PR title to **Members active tasks API**. So as per the current implementation I am fetching all the task of that user irrespective of status. Do we need only active tasks ?
@swarajpure @sumitd94

utils/users.js
```diff
@@ -8,8 +8,15 @@ const { fetchUser } = require('../models/users')
  */
 const getUserId = async (username) => {
   try {
-    const { user: { id } } = await fetchUser({ username })
-    return id
+    const user = await fetchUser({ username })
+
+    if (user.userExists) {
+      const { user: { id } } = user
```
ankushdharkar, line 14: Does this local const variable do anything?

controllers/tasks.js
```diff
@@ -43,6 +43,33 @@ const fetchTasks = async (req, res) => {
...
+    const allTasks = await tasks.fetchUserTasks(username)
+
+    if (allTasks) {
+      return res.json({
+        message: 'Tasks returned successfully!',
+        tasks: allTasks.length > 0 ? allTasks : []
+      })
+    }
+
+    return res.boom.notFound('User doesn\'t exist')
```
ankushdharkar, line 65: Why do we say that the user does not exist, when they simply don't have any tasks assigned to them?
We can have users without any tasks, right?

models/tasks.js
ankushdharkar: Great question @bhk31!
We should ideally use the util (which does also make a read request to get the user, but it can be sped up in future by caching)

ankushdharkar (changes requested):

controllers/tasks.js
```diff
@@ -43,6 +43,33 @@ const fetchTasks = async (req, res) => {
...
+ * @param res {Object} - Express response object
+ */
+const getUserTasks = async (req, res) => {
+  try {
+    const { username } = req.params
+    const allTasks = await tasks.fetchUserTasks(username)
+
+    if (allTasks === false) {
+      return res.boom.notFound('User doesn\'t exist')
+    }
```
swarajpure, line 60: This doesn't make much sense to me. Like how can `allTasks` be false? 😅 

More on this in `fetchUserTasks` function

utils/users.js
```diff
@@ -8,8 +8,13 @@ const { fetchUser } = require('../models/users')
...
   try {
-    const { user: { id } } = await fetchUser({ username })
-    return id
+    const user = await fetchUser({ username })
+
+    if (user.userExists) {
+      return user.user.id
+    } else {
+      return false
+    }
```
swarajpure, line 17: Could we use destructuring here?

```JS
const { userExists, user: { id } } = await fetchUser({ username })
```

Also, the return statements could be simplified with ternary operator:
```JS
return userExists ? id : false
```

What do you think which one is easy to read?

models/tasks.js
```diff
@@ -114,8 +113,12 @@ const fetchTask = async (taskId) => {
 
 const fetchUserTasks = async (username, statuses = []) => {
   try {
-    const { user } = await fetchUser({ username })
-    const userId = await userUtils.getUserId(user.username)
+    const userId = await userUtils.getUserId(username)
+
+    if (!userId) {
+      return false
+    }
```
swarajpure, line 120: We shouldn't just return false here, we saw how it looks when `fetchUserTasks` is called, right?

We could instead return an object:
```JS
return { userNotFound: true }
```

Now at the time of function call, we can write:
```JS
if (allTasks.userNotFound) {
  return res.boom.notFound('User doesn\'t exist')
}
```


Does it feel better?

swarajpure (changes requested):
Good job there!
Suggesting just some minor things to make the code easy to read.

Thank you! 😄

utils/users.js
bhk31: Actually @ankushdharkar previously [commented ](https://github.com/Real-Dev-Squad/website-backend/pull/338#discussion_r619690834). So I thought we can go ahead without destructuring.  As we are just returning the id. Let me know what to finalize.

swarajpure: To answer his question, yes, we're returning the const variable
He didn't say anything about the destructuring, did he? 🤔

bhk31: I answered it No and did the changes. As we are not using that `const` it anywhere in function further then why to store it in variable. 
I am ok with both the approach as it won't need much of changes. Just wanted to clarify why I did it.

bhk31:
Closes #327

routes/tasks.js
```diff
@@ -126,4 +126,35 @@ router.post('/', authenticate, authorizeOwner, createTask, tasks.addNewTask)
...
 
+/**
+ * @swagger
+ * /tasks/username:
+ *   get:
+ *     summary: Use to get all the tasks of the requested user
+ *     tags:
+ *       - Tasks
+ *     security:
+ *       - bearerAuth: []
```
sumitd94, line 137: is this required? we are not checking for authentication here

sumitd94 (changes requested):
One small change, post which we can merge this.

routes/tasks.js
swarajpure: Good catch @sumitd94!

swarajpure: Good catch @sumitd94!

sumitd94 (approved):
Thanks, for the changes, looks good to me

swarajpure (approved):
Looks good to me!

Congrats on your first PR in backend! 🥳

PR #339, opened 2021-04-24
Add authorization middleware based on roles

Cartmanishere:
On backend, some of the routes are only allowed to be called for certain sets of users.
For e.g
- Super user (a.k.a @ankushdharkar )
- App owners (as defined by the list of mockdata in the repo)

Instead of writing different middleware for authorization of each particular type of user, 
this PR aims to provide a unified way of setting authorization for any route.

This is based on the user object having `roles` information.

A user object will contain the roles information as follows:
```
{
  roles: {
    super_user: true/false,
            OR
    app_owner: true/false
  }
}
```

There is hierarchy of required roles. For e.g `superUser` has higher authority than `appOwner`. 
So any route which requires `appOwner` permission, can also be done by a `superUser`.

But a user with `app_owner` role cannot do a request which requires authorization of `superUser`.


middlewares/authorization.js
```diff
@@ -1,13 +1,56 @@
...
+ * for that route. Currently following roles are supported:
+ * - `authorizeUser('superUser')`
+ * - `authorizeUser('appOwner')`
+ * Note: This must be added on routes after the `authenticate` middleware.
+ * @param {String} requiredRole - The least role authority required for a route.
+ * @returns {Function} - A middleware function that authorizes given role.
+ */
+const authorizeUser = (requiredRole) => {
+  return (req, res, next) => {
+    let { roles } = req.userData
```
ankurnarkhede, line 40: Add default as `{}` if roles is `undefined`.

ankurnarkhede (changes requested):

middlewares/authorization.js
```diff
@@ -1,13 +1,56 @@
...
+const REQUIRED_ROLES_PRIORITY = {
+  superUser: ['super_user'],
+  appOwner: ['app_owner', 'super_user'],
+  default: ['default', 'super_user', 'app_owner']
+}
+
+/**
+ * Check if the user has enough authorization based on their role.
+ * User would have following roles schema -
+ * userRoles = {member: true, app_owner: true, super_user: true}
```
Cartmanishere, line 17: `member` is not used. update comment.

firebase.json
```diff
@@ -1,7 +1,7 @@
 {
   "emulators": {
     "firestore": {
-      "port": 8080
+      "port": 8081
```
ankurnarkhede, line 4: This change can be reverted, @Cartmanishere!

routes/stocks.js
```diff
@@ -62,6 +62,6 @@ router.get('/', fetchStocks)
  *           schema:
  *             $ref: '#/components/schemas/errors/badImplementation'
  */
-router.post('/', authenticate, authorization, createStock, addNewStock)
+router.post('/', authenticate, authorizeUser('superUser'), createStock, addNewStock)
```
ankurnarkhede, line 65: Can we have the strings `'superUser', 'appOwner', etc` defined as a constant and then use it here. We might mess up with the strings sometimes leading to an issue!

ankurnarkhede (changes requested):
@Cartmanishere , added few comments.
Let me know if you can complete the roles strings to be defined as a constant by 1st May. Else, we can merge this PR and track it for later in an another issue.

ankurnarkhede (approved):
LGTM ✅ 
To be merged after constants for roles are added or the corresponding issue is created.

Cartmanishere:
@ankurnarkhede Created an issue for the changes you have requested.

I will pick this up in sometime or if someone wants, they can work on it too.

https://github.com/Real-Dev-Squad/website-backend/issues/353

PR #348, opened 2021-04-27
Added status property in user object

ASHWINIAJIT:
@swarajpure @sumitd94 

I have added the Status property in User model.

Please review the PR!

### DIFF
--- a/constants/users.js
+++ b/constants/users.js
@@ -1,1 +1,2 @@
-module.exports = ['ooo', 'idle', 'active']
+const userStatusEnum = ['ooo', 'idle', 'active']
+module.exports = userStatusEnum
--- a/constants/users.js
+++ b/constants/users.js
@@ -1,0 +1,1 @@
+module.exports = ['ooo', 'idle', 'active']
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -1,6 +1,7 @@
 const joi = require('joi')
 
 const updateUser = async (req, res, next) => {
+  const userStatusEnum = ['ooo', 'idle', 'active']
   const schema = joi.object().keys({
     phone: joi.string().optional(),
     email: joi.string().optional(),
@@ -14,7 +15,8 @@ const updateUser = async (req, res, next) => {
     linkedin_id: joi.string().optional(),
     twitter_id: joi.string().optional(),
     instagram_id: joi.string().optional(),
-    website: joi.string().optional()
+    website: joi.string().optional(),
+    status: joi.any().valid(...userStatusEnum)
   })
 
   try {
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -1,7 +1,7 @@
 const joi = require('joi')
+const userStatusEnum = require('../../constants/users')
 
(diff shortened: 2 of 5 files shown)


middlewares/validators/user.js
```diff
@@ -14,7 +15,8 @@ const updateUser = async (req, res, next) => {
     linkedin_id: joi.string().optional(),
     twitter_id: joi.string().optional(),
     instagram_id: joi.string().optional(),
-    website: joi.string().optional()
+    website: joi.string().optional(),
+    status: joi.any().valid(...userStatusEnum)
```
swarajpure, line 19: Do we not provide `.optional()` with this one?

middlewares/validators/user.js
```diff
@@ -1,6 +1,7 @@
 const joi = require('joi')
 
 const updateUser = async (req, res, next) => {
+  const userStatusEnum = ['ooo', 'idle', 'active']
```
swarajpure, line 4: can we please have these under `constants` folder inside a new file called `users.js` maybe?

swarajpure (changes requested):
Suggesting one minor change.

Also, can you please add testcases for the `PATCH users/self` route for the following cases:
1. When we pass correct status with the test (one of the allowed values)
2. When we pass invalid status value

So we'll be confident that this change won't break on production 😄

middlewares/validators/user.js
ASHWINIAJIT: Added .optional() for status property as you suggested!

https://github.com/Real-Dev-Squad/website-backend/pull/348/commits/1f0c6c2a1102c0558dfc2d1cf8cfcc4d9a8ab8d5

middlewares/validators/user.js
ASHWINIAJIT: Created a new file as " users.js" for constant values of status.

https://github.com/Real-Dev-Squad/website-backend/pull/348/commits/1f0c6c2a1102c0558dfc2d1cf8cfcc4d9a8ab8d5

ASHWINIAJIT:
@swarajpure 

I have added the test cases for status property and worked on comments.

Please review the PR!

constants/users.js
```diff
@@ -0,0 +1 @@
+module.exports = ['ooo', 'idle', 'active']
```
swarajpure, line 1: Can we call this list as something like `userStatusTypes` and then export that list? Because if a third person visits this file, they might not be able to make sense of what this list in doing in users constants, I hope that makes sense 😄

swarajpure (changes requested):
Just one small change and we'd be good to go with merging this PR 🥳

Good job with tests! Thanks for that!

constants/users.js
ASHWINIAJIT: I have updated the constants/users.js file as you suggested!

https://github.com/Real-Dev-Squad/website-backend/pull/348/commits/76d5c2900c4f50882b2db2e6fb879b76231885e9

ASHWINIAJIT:
@swarajpure I have made the changes as you suggested. Please check!

swarajpure (approved):
Looks good! Thanks for the changes!
Good job with the tests! 🥳

PR #351, opened 2021-04-28
Script to migrate user roles

sumitd94:
This PR contains the migration script, which will be running only once which will remove the `isMember` field, and add a new field `roles` which will be an object something like this

```javascript
{
 roles: { member: true }
}
```

All those users who have `isMember` set to `true`, will have `member: true` in the roles objects, and for the rest of all users, `member: false`

endpoint - `https://api.realdevsquad.com/members/member-to-role-migration`


ankurnarkhede:
@sumitd94 , executing this will solve the failing tests in https://github.com/Real-Dev-Squad/website-backend/pull/296, right?
Also, is there any other way to execute this query or adding the route is the only way?

ankurnarkhede (approved):
LGTM ✅

Issue #355, opened 2021-05-14
Promote new user to Member


ankushdharkar:
Admins should be able to convert a "new user" into a "member" on click of a button.

Update the role of a new user to "member" (That will give them dedicated space on our members page)

nandita1:
I'll take this up

PR #357, opened 2021-05-15
api to make new members members #355

nandita1:
PR for #355

### DIFF
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -1,5 +1,8 @@
 const memberQuery = require('../models/members')
 const tasks = require('../models/tasks')
+const { fetchUser } = require('../models/users')
+
+const ERROR_MESSAGE = 'Something went wrong. Please try again or contact admin'
 
 /**
  * Fetches the data about our members
@@ -46,7 +49,30 @@ const getIdleMembers = async (req, res) => {
   }
 }
 
+/**
+ * Makes a new member a member
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const moveToMembers = async (req, res) => {
+  try {
+    const username = req.params.username
+    const result = await fetchUser({ username: username })
+    if (result.userExists) {
+      await memberQuery.moveToMembers(result.user)
+      return res.status(204).send()
+    }
+    return res.boom.notFound("User doesn't exist")
+  } catch (err) {
+    logger.error(`Error while retriving contributions ${err}`)
+    return res.boom.badImplementation(ERROR_MESSAGE)
+  }
+}
+
 module.exports = {
   getMembers,
-  getIdleMembers
+  getIdleMembers,
+  moveToMembers
 }
--- a/models/members.js
+++ b/models/members.js
@@ -37,6 +37,29 @@ const fetchMembers = async () => {
   }
 }
 
+/**
+ * changes the role of a new user to member
+ * @param userData { Object }: User data object to be modified
+ * @return {Promise<userModel|object>}
+ */
+
+const moveToMembers = async (userData) => {
+  try {
+    const userDoc = await userModel.doc(userData.id).get()
+    const user = userDoc.data()
+    const roles = user.roles ? { ...user.roles, member: true } : { member: true }
+    await userModel.doc(userData.id).set({
+      ...user,
+      roles
+    })
+    return user
+  } catch (err) {
+    logger.error('Error updating user', err)
+    throw err
+  }
+}
+
 module.exports = {
-  fetchMembers
+  fetchMembers,
+  moveToMembers
 }
--- a/routes/members.js
+++ b/routes/members.js
@@ -3,6 +3,8 @@ const router = express.Router()
 const members = require('../controllers/members')
 const { addRecruiter } = require('../controllers/recruiters')
 const { validateRecruiter } = require('../middlewares/validators/recruiter')
+const authenticate = require('../middlewares/authenticate')
+const { authorizeUser } = require('../middlewares/authorization')
 
 /**
  * @swagger
@@ -83,5 +85,35 @@ router.get('/idle', members.getIdleMembers)
  */
 
 router.post('/intro/:username', validateRecruiter, addRecruiter)
+
+/**
+ * @swagger
+ * /moveToMembers/:username:
+ *   patch:
+ *     summary: Changes the role of a new member(the username provided in params) to member
+ *     tags:
+ *       - Members
+ *     security:
+ *       - bearerAuth: []
+ *     responses:
+ *       204:
+ *         description: No content
+ *
+ *       401:
+ *       description: unAuthorized
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/components/schemas/errors/unAuthorized'
+ *
+ *       500:
+ *         description: serverUnavailable
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/errors/serverUnavailable'
+ */
+
+router.patch('/moveToMembers/:username', authenticate, authorizeUser('superUser'), members.moveToMembers)
 
 module.exports = router


routes/members.js
```diff
@@ -84,4 +86,34 @@ router.get('/idle', members.getIdleMembers)
...
+ *     summary: Changes the role of a new member(the username provided in params) to member
+ *     tags:
+ *       - Members
+ *     security:
+ *       - bearerAuth: []
+ *     responses:
+ *       204:
+ *         description: No content
+ *
+ *       401:
```
swarajpure, line 102: Please add the schema for 404 as well (in case username not found)

routes/members.js
```diff
@@ -84,4 +86,34 @@ router.get('/idle', members.getIdleMembers)
...
+ *
+ *       500:
+ *         description: serverUnavailable
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/errors/serverUnavailable'
+ */
+
+router.patch('/moveToMembers/:username', authenticate, authorizeUser('superUser'), members.moveToMembers)
```
swarajpure, line 117: Better to have `superUser` coming from constants file?

models/members.js
```diff
@@ -37,6 +37,29 @@ const fetchMembers = async () => {
...
+ * @param userData { Object }: User data object to be modified
+ * @return {Promise<userModel|object>}
+ */
+
+const moveToMembers = async (userData) => {
+  try {
+    const userDoc = await userModel.doc(userData.id).get()
+    const user = userDoc.data()
+    const roles = user.roles ? { ...user.roles, member: true } : { member: true }
+    await userModel.doc(userData.id).set({
```
swarajpure, line 51: Can we make use of `update()` method?
https://firebase.google.com/docs/firestore/manage-data/add-data#update-data

So we don't need the whole userData, only `id` would be required as function parameter

controllers/members.js
```diff
@@ -46,7 +49,30 @@ const getIdleMembers = async (req, res) => {
...
+ * Makes a new member a member
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const moveToMembers = async (req, res) => {
+  try {
+    const username = req.params.username
+    const result = await fetchUser({ username: username })
```
swarajpure, line 62: We can use JS shorthand here
https://alligator.io/js/object-property-shorthand-es6/

Would become `fetchUser({ username })` then

swarajpure (changes requested):
Congratulations on your first feature pull request in Real Dev Squad! 🥳 

Great job in there!

Suggesting some changes. Please check.

routes/members.js
```diff
@@ -84,4 +87,40 @@ router.get('/idle', members.getIdleMembers)
...
+ *
+ *       500:
+ *         description: serverUnavailable
+ *         content:
+ *           application/json:
+ *             schema:
+ *               $ref: '#/components/schemas/errors/serverUnavailable'
+ */
+
+router.patch('/moveToMembers/:username', authenticate, authorizeUser(ROLES.SUPERUSER), members.moveToMembers)
```
swarajpure, line 124: Please use the api-schema-generation script in package.json to update the API schema with this new route

swarajpure (changes requested):
Have we handled the case when the user is already a member? 🤔 

Also, can you please write tests for this route? Anything related to users is critical, therefore tests are important here

constants/roles.js
```diff
@@ -0,0 +1,6 @@
+const ROLES = {
+  SUPERUSER: 'superUser',
+  APPOWNER: 'appOwner'
+}
+
+module.exports = { ROLES }
```
sumitd94, line 6: Why exporting as an object when `ROLES` itself is an object?
`module.exports = ROLES` should be a good option, and then in the routes, you can destructure and get `SUPERUSER` and `APPOWNER` values. Makes sense?

controllers/members.js
```diff
@@ -46,7 +49,33 @@ const getIdleMembers = async (req, res) => {
...
+/**
+ * Makes a new member a member
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const moveToMembers = async (req, res) => {
+  try {
+    const username = req.params.username
```
sumitd94, line 61: NIT: destructure?

models/members.js
```diff
@@ -37,6 +37,31 @@ const fetchMembers = async () => {
...
+ * @return existingMember { Boolean }: to show if user is already a member
+ */
+
+const moveToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user.roles && user.roles.member) {
+      return true
+    }
```
sumitd94, line 52: NIT: can be turned into a one liner
```javascript
if (user.roles && user.roles.member) return true;
```

routes/members.js
```diff
@@ -84,4 +87,58 @@ router.get('/idle', members.getIdleMembers)
...
```
sumitd94, line 142: NIT: It would also be nice if we can destructure the `members.moveToMembers`.
The code will look much readable, cleaner, and concise.

sumitd94 (changes requested):
Few Comments added, please check. Apart from this, it looks great. Good Job 🥇

controllers/members.js
```diff
@@ -46,7 +49,33 @@ const getIdleMembers = async (req, res) => {
...
+
+const moveToMembers = async (req, res) => {
+  try {
+    const { username } = req.params
+    const result = await fetchUser({ username })
+    if (result.userExists) {
+      const alreadyMember = await memberQuery.moveToMembers(result.user.id)
+      if (alreadyMember) {
+        return res.status(400).json({ message: 'User Already is a member' })
+      }
```
swarajpure, line 67: We can use boom 400 here:
https://www.npmjs.com/package/boom#boombadrequestmessage-data

models/members.js
```diff
@@ -37,6 +37,29 @@ const fetchMembers = async () => {
...
+const moveToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user.roles && user.roles.member) return true
+    const roles = user.roles ? { ...user.roles, member: true } : { member: true }
+    await userModel.doc(userId).update({
+      roles
+    })
+    return false
```
swarajpure, line 55: The return statements seem a little bit confusing 🤔 
What if we return `{ isAlreadyMember: true }` on line 50?

And for the second case, we could return `{ movedToMember: true }`

What do you think?

test/integration/members.test.js
```diff
@@ -89,4 +103,38 @@ describe('Members', function () {
...
+  describe('PATCH /members/moveToMembers', function () {
+    before(async function () {
+      await cleanDb()
+      const userId = await addUser(superUser)
+      jwt = authService.generateAuthToken({ userId })
+    })
+    it('Should make the user a member', function (done) {
+      chai
+        .request(app)
+        .patch(`/members/moveToMembers/${superUser.username}`)
```
swarajpure, line 116: Here we are generating cookie for ankush which makes sense
However we are also updating him only, can we try updating someone else?

Currently you can use any user, no user in the fixtures is having the role `member`

test/integration/members.test.js
```diff
@@ -89,4 +103,38 @@ describe('Members', function () {
...
+        .set('cookie', `${cookieName}=${jwt}`)
+        .end((err, res) => {
+          if (err) { return done(err) }
+
+          expect(res).to.have.status(400)
+
+          return done()
+        })
+    })
+  })
```
swarajpure, line 139: Perfect after we update the user! 👌🏻

test/integration/members.test.js
```diff
@@ -89,4 +103,38 @@ describe('Members', function () {
...
     })
   })
+
+  describe('PATCH /members/moveToMembers', function () {
+    before(async function () {
+      await cleanDb()
+      const userId = await addUser(superUser)
+      jwt = authService.generateAuthToken({ userId })
+    })
+    it('Should make the user a member', function (done) {
```
swarajpure, line 113: We need test cases for all the 2xx and 4xx responses:
1. Requested user not existing
2. User making the request is not authenticated

test/integration/members.test.js
```diff
@@ -89,4 +103,38 @@ describe('Members', function () {
...
+      chai
+        .request(app)
+        .patch(`/members/moveToMembers/${superUser.username}`)
+        .set('cookie', `${cookieName}=${jwt}`)
+        .end((err, res) => {
+          if (err) { return done(err) }
+
+          expect(res).to.have.status(204)
```
swarajpure, line 121: We could also assert that the body should be empty

test/integration/members.test.js
```diff
@@ -89,4 +103,38 @@ describe('Members', function () {
...
+    })
+    it('Should return 400 if user is already a member', function (done) {
+      chai
+        .request(app)
+        .patch(`/members/moveToMembers/${superUser.username}`)
+        .set('cookie', `${cookieName}=${jwt}`)
+        .end((err, res) => {
+          if (err) { return done(err) }
+
+          expect(res).to.have.status(400)
```
swarajpure, line 134: We could assert on the error message in res.body

swarajpure (changes requested):
Suggesting some changes in model function and tests
Please check 🙂

models/members.js
```diff
@@ -37,6 +37,29 @@ const fetchMembers = async () => {
...
+const moveToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user.roles && user.roles.member) return { isAlreadyMember: true, movedToMember: false }
+    const roles = user.roles ? { ...user.roles, member: true } : { member: true }
+    await userModel.doc(userId).update({
+      roles
+    })
+    return { isAlreadyMember: false, movedToMember: true }
```
swarajpure, line 55: Do we really need false fields? 🤔 
I guess you're trying to make the object being returned consistent, is that so?

swarajpure (changes requested):
Hey thanks for the changes.
I see code being repeated in tests. Can we have `cleanDb` call in `afterEach`?

Adding a user inside `it` block seems more like a hack 😅 
Right now the order of tests will matter. If someone adds/removes/changes one test out of your tests, the following tests will break since the sequence in which you're doing these operations does matter. I hope I'm making sense.

I know sometimes tests can get trickier, but we take out some common stuff out of all the steps and add them in `afterEach` and `beforeEach` When we need a very specific case, like there should be no member in DB, we might have to clean the db in `it` block only. Another case could be for 400, we could use the model function inside `it` maybe?

@ankurnarkhede @Cartmanishere would like to hear your inputs on this one

constants/roles.js
```diff
@@ -0,0 +1,6 @@
+const ROLES = {
+  SUPERUSER: 'superUser',
```
sebuelias, line 2: I think the role key for SUPERUSER is super_user
Please confirm @swarajpure

Cartmanishere: In DB, it is stored as `super_user` and `app_owner`. 

But when adding the authorization to any route using `authorizeUser`, the roles that should be passed in `superUser`.

Please check the implementation in `authorization.js`

PR #359, opened 2021-05-16
Using firebase tools from local node_modules

### LINKED ISSUE #358: Use firebase-tools from local node_modules
Let's use the firestore tools from the local project, instead of having to install the package globally

ankushdharkar:
Closes #358, #361

~~Waiting on resolving the failing test for /tasks~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,6 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm install -g firebase-tools
     - run: npm install
     - run: npm run build --if-present
     - run: npm test
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,17 +108,15 @@ The following project structure should be followed:
 - Pre-requisites:
   - Node.js version 8.0 or higher.
   - Java version 1.8 or higher.
-  - Firebase CLI: `npm install -g firebase-tools`
 
 ## Using Firebase Emulator Locally
 - [Firebase Local Emulator Suite](https://firebase.google.com/docs/emulator-suite) can be used locally as the DB for the project
 - Pre-requisites:
   - Node.js version 8.0 or higher.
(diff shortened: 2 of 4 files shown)


swarajpure (commented):
Wouldn't we need to need to make change in the shell scripts? 
Wouldn't shell script fail to recognise what's `firebase emulators` without it being installed globally? 🤔 
https://github.com/Real-Dev-Squad/website-backend/blob/09c53d861f1133e9ab2bea3a1421bcec855f8716/scripts/tests/testIntegration.sh#L7

What I'm saying is, will we need give path to `firebase-emulators` in this case?


Or are the tests working fine with these changes?

ankurnarkhede (changes requested):
@ankushdharkar , as we are adding firebase-tools in package.json, should we remove the github actions command to install it globally at: https://github.com/Real-Dev-Squad/website-backend/blob/develop/.github/workflows/test.yml#L30 ?

ankushdharkar:
@ankurnarkhede https://github.com/Real-Dev-Squad/website-backend/issues/361 :D

ankurnarkhede:
Also, how is this dependent on the /tasks tests as per you mentioned in the PR description?

swarajpure:
Your local setup would still have firebase-tools globally installed, which makes the tests run on local

Also on CI, this PR might be passing because we're installing firebase-tools globally in CI as well.

I guess the change mentioned by @ankurnarkhede should be a part of this PR itself. What do you think? @ankurnarkhede @ankushdharkar

ankushdharkar:
@ankurnarkhede It's not. My /tasks were failing until yesterday. Not sure why they are passing today 🤔

ankushdharkar:
I like that idea @swarajpure. Made the required change. Thanks!

swarajpure (approved):
Looks good to me! 👌🏻

ankurnarkhede (commented):
LGTM ✅

ankurnarkhede (approved):

ankurnarkhede (approved):

Issue #263, opened 2021-03-16
Create a GET `notifications` API


swarajpure:
API contract: https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/notifications

ganshul34:
@swarajpure, the `page` in the above docs is the `current page`  or the `total number` of pages for which the notifications are to be fetched?

swarajpure:
> @swarajpure, the `page` in the above docs is the `current page` or the `total number` of pages for which the notifications are to be fetched?

It'll be current page
Let's say I've seen the data on first page or first `n` entries, I want to see next `n` entries, so I'll go to next page that is 2nd page in this case
So I'll pass `page=2`
I hope that makes sense :)

PR #326, opened 2021-04-13
Refactor `/pullrequests` API and write integration tests for them

swarajpure:
- [x] Added case for invalid `username` for `/pullrequests/user/:username` route
- [x] Wrote integration tests for `/pullrequests` APIs
- [x] Wrote and used stubs for functions which require network call to GitHub API

Learnt about stubbing 🔪 from this PR 🤣 

Fixes #320

PR #362, opened 2021-05-17
Feature/web api contracts get api notifications

ganshul34:
[GET API for Getting Paginated Notifications for Current LoggedIn User ***#263***](https://github.com/Real-Dev-Squad/website-backend/issues/263)


controllers/notifications.js
```diff
@@ -0,0 +1,41 @@
...
+
+module.exports.getNotificationsForUser = async (req, res) => {
+  try {
+    const { page: currentPage, n: limit } = req.query
+    const { id: userId } = req.userData
+
+    if (!currentPage || !limit) {
+      logger.error('currentPage/ limit not present in the query')
+      return res.boom.badRequest('Please check your query again')
+    }
```
swarajpure, line 17: Query params are optional
We need not do these checks. Instead we should have use some default value if query params are not externally defined.

models/notifications.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ *  Get  the notifications data for current loggedIn user
+ *
+ * @param currentPage { Integer }: Page for which notifications are to be fetched
+ * @param limit { Integer }: Number of Notifications per page
+ * @param userId { String }: userId of current loggedIn user
+ * @return {Promise<notificationModal|Array>}}
+ */
+
+module.exports.fetchNotifications = async ({ currentPage, limit, userId }) => {
```
swarajpure, line 12: Please seperate function definitions from export statements to increase readability and for future changes.

test/integration/notifications.test.js
```diff
@@ -0,0 +1,217 @@
...
+describe('Notifications', function () {
+  let jwt
+  beforeEach(async function () {
+    const userId = await addUser()
+    jwt = authService.generateAuthToken({ userId })
+  })
+
+  afterEach(async function () {
+    await cleanDb()
+  })
```
swarajpure, line 24: We're not getting the notifications from anywhere. Can we use stubbing in this case? Currently we don't have any route which creates notifications. We could create mock of the firestore data and fetch that instead of hitting DB for now. You can refer to #326 for stubbing.

We can later do actual DB calls once we have at least one functions which creates notifications in the DB. For now, we could mock it.

swarajpure (changes requested):
It is the property of query params that they are optional. We need to have some default value in case we don't get query params in URL.

Also, please have functions written seperately and module.exports should be at the end of file. It then becomes really use to see which all functions are being exported from a particular file. Your code might become messy if we wish to add more functions in the same file. You can check existing code for more reference 🙂 

Great job in there! Wasn't expecting tests in your first PR in backend! 🤯

controllers/notifications.js
ganshul34: Is it fine If I set currentPage as 1 and the limit as 10?

swarajpure: Yes, that works 👍🏻

controllers/notifications.js
```diff
@@ -0,0 +1,51 @@
...
+    const { id: userId } = req.userData
+
+    if (!currentPage) {
+      currentPage = 1
+    }
+
+    if (!limit) {
+      limit = 10
+    }
+
```
prakashchoudhary07, line 21: NIT:
you can pass default values for the variable while destructuring,
`const { page: currentPage=1 , n: limit=10 } = req.query` and avoid if statements

controllers/notifications.js
```diff
@@ -0,0 +1,51 @@
...
+    let parsedCurrentPage = +currentPage
+    let parsedLimit = +limit
+
+    if (isNaN(parsedCurrentPage) || parsedCurrentPage <= 0) {
+      parsedCurrentPage = 1
+    }
+
+    if (isNaN(parsedLimit) || parsedLimit <= 0) {
+      parsedLimit = 10
+    }
```
prakashchoudhary07, line 31: NIT: Try ternary operators

ganshul34: But by using **ternary operators*, won't it affect the code readability, since I need to use the ***nested*** ternary operators.

prakashchoudhary07: Well, it's up to you, whether use it or not, If you feel it would be hard to read then you can skip it :smiley:

prakashchoudhary07: But why would you need nested ternary operators?
This would be some thing like this 
`const parsedCurrentPage = (isNaN(parsedCurrentPage) || parsedCurrentPage <= 0) ? 1 : currentPAge`, right or am I missing something?

ganshul34: Yes , correct

controllers/notifications.js
```diff
@@ -0,0 +1,51 @@
...
+  try {
+    let { page: currentPage, n: limit } = req.query
+    const { id: userId } = req.userData
+
+    if (!currentPage) {
+      currentPage = 1
+    }
+
+    if (!limit) {
+      limit = 10
```
ankurnarkhede, line 19: This is the default, right?

We can add it as config then?

ankurnarkhede (changes requested):

PR #364, opened 2021-05-29
Unit tests for fetchUserTasks model function

bhk31:
**Added unit test for these test cases**

fetchUserTasks

### DIFF
--- a/test/unit/models/task.test.js
+++ b/test/unit/models/task.test.js
@@ -131,5 +131,11 @@ describe('tasks', function () {
         expect(allTasks).to.deep.equal(validatedTasks)
       })
     })
+    describe('When user does not exists', function () {
+      it('Should return user not found', async function () {
+        const allTasks = await tasks.fetchUserTasks('dummyUser')
+        expect(allTasks.userNotFound).to.equal(true)
+      })
+    })
   })
 })
--- a/test/unit/models/task.test.js
+++ b/test/unit/models/task.test.js
@@ -1,0 +1,135 @@
+const chai = require('chai')
+const { expect } = chai
+const tasks = require('../../../models/tasks')
+const sinon = require('sinon')
+const addUser = require('../../utils/addUser')
+const userData = require('../../fixtures/user/user')()
+const appOwner = userData[3]
+const otherUser = userData[0]
+
+describe('tasks', function () {
+  before(async function () {
+    await addUser(appOwner)
+    await addUser(otherUser)
+  })
+
+  afterEach(function () {
+    sinon.restore()
+  })
+
+  describe('fetchUserTasks', function () {
+    describe('When user exists and status is not passed', function () {
+      it('Should return blank array if no task are assigned to user', async function () {
+        const allTasks = await tasks.fetchUserTasks(appOwner.username)
+        expect(allTasks).to.eql([])
+      })
+
+      it('Should return group tasks in array', async function () {
+        const taskData = [
+          {
+            title: 'Title of the task 1',
+            type: 'group',
+            endsOn: '1621814400',
+            startedOn: '1621728000',
+            status: 'completed',
+            participants: [
+              appOwner.username
+            ]
+          }]
+
+        await tasks.updateTask(taskData[0])
+        const allTasks = await tasks.fetchUserTasks(appOwner.username)
+        const task1 = allTasks[0]
+        if (task1.type === 'group') {
+          expect(task1.participants).to.include(appOwner.username)
+        }
+      })
+
+      it('Should return both group and feature tasks in array', async function () {
+        const taskData = [{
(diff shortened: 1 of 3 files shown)


swarajpure (changes requested):
This PR seems to be having refactoring along with tets
Can we have seperate PRs for both?
It becomes difficult to review multiple things happening

You can keep ONLY the refactoring part in this PR. 

For tests, you can create a new branch from this branch, keep all the tests in that branch and then commit from there, make a PR for the same. 
Make sure that the target branch for that PR is THIS branch (can be changed anytime from GitHub UI, not a big deal)

Please do the needful 🙂

utils/tasks.js
```diff
@@ -5,15 +5,15 @@ const fromFirestoreData = async (task) => {
...
+  let { createdBy, assignee, participants, type } = task
   if (createdBy) {
     createdBy = await getUsername(createdBy)
   }
   if (assignee) {
     assignee = await getUsername(assignee)
   }
 
-  if (Array.isArray(participants)) {
+  if (type === 'group') {
```
ankurnarkhede, line 16: Can we have this as a constant to be used elsewhere in the app?

ankurnarkhede (changes requested):

bhk31:
The test which is failed is fixed under this PR https://github.com/Real-Dev-Squad/website-backend/pull/369/files

test/unit/models/task.test.js
```diff
@@ -0,0 +1,141 @@
...
+        expect(allTasks).to.eql([])
+      })
+
+      it('Should return group tasks in array', async function () {
+        const taskData = [
+          {
+            title: 'Title of the task 1',
+            type: 'group',
+            endsOn: '1621814400',
+            startedOn: '1621728000',
```
swarajpure, line 33: Shouldn't these be numbers instead of string? 🤔

test/unit/models/task.test.js
```diff
@@ -0,0 +1,141 @@
...
+          expect(task1.participants).to.include(appOwner.username)
+        }
+      })
+
+      it('Should return both group and feature tasks in array', async function () {
+        const taskData = [{
+          title: 'Title of the task 2',
+          type: 'feature',
+          endsOn: '1621814400',
+          startedOn: '1621728000',
```
swarajpure, line 53: re: number

test/unit/models/task.test.js
```diff
@@ -0,0 +1,141 @@
...
+
+        await tasks.updateTask(taskData[0])
+        await tasks.updateTask(taskData[1])
+        await tasks.updateTask(taskData[2])
+
+        const validateStatus = (task) => {
+          return ['active', 'blocked', 'pending'].includes(task.status)
+        }
+
+        const allTasks = await tasks.fetchUserTasks(appOwner.username, ['active', 'pending', 'blocked'])
```
swarajpure, line 129: Since we're using this particular array again and again, better to have it coming from a variable?

test/unit/models/task.test.js
```diff
@@ -0,0 +1,141 @@
...
+
+describe('tasks', function () {
+  before(async function () {
+    await addUser(appOwner)
+    await addUser(otherUser)
+  })
+
+  afterEach(function () {
+    sinon.restore()
+  })
```
swarajpure, line 18: We aren't using sinon anywhere, why do we need this then?
We should be using `cleanDb` instead, in afterEach 🤔 

Can you please explain what's going on here? I'm a bit confused

test/unit/models/task.test.js
```diff
@@ -0,0 +1,141 @@
...
+
+    describe('When user exists and status is passed', function () {
+      it('Should return blank array if user has no task with that status', async function () {
+        const allTasks = await tasks.fetchUserTasks(otherUser.username)
+        expect(allTasks).to.eql([])
+      })
+
+      it('Should return all completed task if completed status is passed', async function () {
+        const isCompleted = (task) => {
+          return task.status === 'completed'
```
swarajpure, line 85: better make of existing constants than using strings for task status

swarajpure (changes requested):
Some suggestions and some doubts, please check!

test/unit/models/task.test.js
```diff
@@ -0,0 +1,141 @@
...
+            status: 'completed',
+            participants: [
+              appOwner.username
+            ]
+          }]
+
+        await tasks.updateTask(taskData[0])
+        const allTasks = await tasks.fetchUserTasks(appOwner.username)
+        const task1 = allTasks[0]
+        if (task1.type === 'group') {
```
sebuelias, line 43: we have constants for tasks in constants/task.js -> TASK_STATUS and TASK_TYPE
Better we use them for the task type checks and task status checks

Issue #368, opened 2021-05-31
Update tasks/:username route to return active task.


bhk31:
This API is needed for issue. https://github.com/Real-Dev-Squad/website-members/issues/146

ankurnarkhede:
@bhk31 , do we have a different API to get inactive tasks?
What is the purpose of this API, can you describe it in description?

bhk31:
> @bhk31 , do we have a different API to get inactive tasks?

@ankurnarkhede No we don't have any such API

>What is the purpose of this API, can you describe it in description?

Updated the description

ankurnarkhede:
@bhk31 , yeah, but what about the inactive tasks?
In the user's profile, we will also need to show all the tasks, active as well as inactive, right?
The PR is currently just sending out active tasks. We can have any query param stating the filter required.

bhk31:
> @bhk31 , yeah, but what about the inactive tasks?
> In the user's profile, we will also need to show all the tasks, active as well as inactive, right?
> The PR is currently just sending out active tasks. We can have any query param stating the filter required.

This has been requested by @sumitd94. Previously it was returning all task irrespective of the status.

Please check https://discord.com/channels/673083527624916993/735063616168001536/848896848706142248

sumitd94:
@bhk31 Ideally the scenario should be, you will be getting a query param `?status=active`
Based on the status you have to return the tasks.

Make the API in such a way, if I pass `?status=inactive` it should return all the inactive tasks, and so on.

Change the API and make sure it does not break the system.

PR #369, opened 2021-05-31
Fetch active tasks for user 

bhk31:


ankurnarkhede (approved):
LGTM ✅

ankurnarkhede (changes requested):
Pending comment on https://github.com/Real-Dev-Squad/website-backend/issues/368

sumitd94 (dismissed):
Looks good to me

controllers/tasks.js
```diff
@@ -52,8 +52,13 @@ const fetchTasks = async (req, res) => {
  */
 const getUserTasks = async (req, res) => {
   try {
+    const statuses = req.query.status
     const { username } = req.params
-    const allTasks = await tasks.fetchUserTasks(username)
+    let allTasks = []
+
+    if (statuses) {
+      allTasks = await tasks.fetchUserTasks(username, statuses.split(','))
```
sumitd94, line 60: why are we splitting? we are not going to have the request that will look something like this - `tasks/:username?status=active,inactive`
this is invalid.
so you can do this

```javascript
await tasks.fetchUserTasks(username, [status])
```

Also rename `statuses` to `status`

controllers/tasks.js
```diff
@@ -52,8 +52,13 @@ const fetchTasks = async (req, res) => {
  */
 const getUserTasks = async (req, res) => {
   try {
+    const statuses = req.query.status
     const { username } = req.params
-    const allTasks = await tasks.fetchUserTasks(username)
+    let allTasks = []
+
+    if (statuses) {
```
sumitd94, line 59: If no status, then I think we should be calling
```javascript
const allTasks = await tasks.fetchUserTasks(username)
```

sumitd94 (changes requested):
One small change and then it's good to go.

controllers/tasks.js
```diff
@@ -52,8 +52,15 @@ const fetchTasks = async (req, res) => {
  */
 const getUserTasks = async (req, res) => {
   try {
+    const status = req.query.status
     const { username } = req.params
-    const allTasks = await tasks.fetchUserTasks(username)
+    let allTasks = []
+
+    if (status) {
+      allTasks = await tasks.fetchUserTasks(username, status)
```
sumitd94, line 60: status needs to go in as an array
`[status]`

sumitd94 (changes requested):

controllers/tasks.js
bhk31: Ohh yes

sumitd94 (dismissed):
Thanks for the changes!

controllers/tasks.js
```diff
@@ -52,8 +52,15 @@ const fetchTasks = async (req, res) => {
  */
 const getUserTasks = async (req, res) => {
   try {
+    const status = req.query.status
```
swarajpure, line 55: Can we use destructuring here?

test/integration/tasks.test.js
```diff
@@ -261,7 +260,7 @@ describe('Tasks', function () {
     it('Should return 200 when username is valid', function (done) {
       chai
         .request(app)
-        .get(`/tasks/${appOwner.username}`)
+        .get(`/tasks/${appOwner.username}?status=active`)
```
swarajpure, line 263: Can we also have a test for the case when status is not sent?

swarajpure (changes requested):
Suggesting some changes, please check!

test/integration/tasks.test.js
```diff
@@ -284,7 +283,7 @@ describe('Tasks', function () {
     it('Should return 404 when username is invalid', function (done) {
```
ankurnarkhede, line 283: Also add a test case with status `inactive`, `invalidstring`, blank, etc

controllers/tasks.js
```diff
@@ -52,8 +52,15 @@ const fetchTasks = async (req, res) => {
...
     const { username } = req.params
-    const allTasks = await tasks.fetchUserTasks(username)
+    let allTasks = []
+
+    if (status) {
+      allTasks = await tasks.fetchUserTasks(username, [status])
```
ankurnarkhede, line 60: Simply pass status and have a default in the function itself.

ankurnarkhede (changes requested):

controllers/tasks.js
ankurnarkhede: Needs to be added to API defintion.

ankurnarkhede (changes requested):

test/integration/tasks.test.js
bhk31: @ankurnarkhede  Is it ok if I add these test under [this](https://github.com/Real-Dev-Squad/website-backend/issues/381) task as under this task I have to write almost all test cases again, which will revamp whole code of tasks integration test.

test/integration/tasks.test.js
bhk31: @swarajpure Is it ok if I add these test under [this](https://github.com/Real-Dev-Squad/website-backend/issues/381) task as under this task I have to write almost all test cases again, which will revamp whole code of tasks integration test.

bhk31: @swarajpure Will do this change under [this](https://github.com/Real-Dev-Squad/website-backend/issues/381) task.

test/integration/tasks.test.js
bhk31: @ankurnarkhede Will do this change under [this](https://github.com/Real-Dev-Squad/website-backend/issues/381) task.

controllers/tasks.js
```diff
@@ -52,8 +53,15 @@ const fetchTasks = async (req, res) => {
...
+    const { status } = req.query
     const { username } = req.params
-    const allTasks = await tasks.fetchUserTasks(username)
+    let allTasks = []
+
+    if (!Object.values(TASK_STATUS).includes(status)) {
+      return res.boom.notFound('Status not found!')
+    }
+
+    allTasks = await tasks.fetchUserTasks(username, status ? [status] : [])
```
swarajpure, line 64: What if I send `?status=active+completed` in query params? I suppose it should be [...status] instead of [status], what do you think? How will we handle multiple statuses?

bhk31: I thought of handling it like this ? https://github.com/Real-Dev-Squad/website-backend/pull/369#discussion_r650407769

swarajpure: Okay. That's our assumption for now. Let's go ahead with that then 👍🏻

swarajpure: Wait, if it's going to just one status at a time, why are we even wrapping it inside a list? It should simply be a string, right?

bhk31: > why are we even wrapping it inside a list? It should simply be a string, right?
didn't get you with this ?

> That's our assumption for now. Let's go ahead with that then 👍🏻

Didn't get should we go ahead with `?status=active+completed` or `?status=active,completed` or should not support multiple status. I have implemented earlier with this `?status=active,completed` but @sumitd94 said its invalid. So i reverted the code. Let me know with which approach we should move ahead.

swarajpure (approved):
LGTM!

PR #370, opened 2021-06-02
Refactor code in tasks model

bhk31:
Refactored the code in tasks model.

- Created a function `buildTasks` which will build an array of tasks which we want to return as response. This code is repeated at multiple places so created a function for it in `utils/tasks`.
- Renamed the variable `tasksSnapshot` to `groupTasksSnapshot` which is more relevant and logical as it stores tasks which are of type group.
- Renamed the variable `assigneeSnapshot` to `featureTasksSnapshot` which is more relevant and logical as it stores tasks which are of type feature.


bhk31:
The test which is failed is fixed under this PR https://github.com/Real-Dev-Squad/website-backend/pull/369/files

swarajpure (approved):
LGTM! Let @sumitd94 have a look once
Once the PR for fixing the failing test gets merged, we can merge this one

sumitd94 (approved):
Looks good to me!

swarajpure:
I re-ran the checks twice already but still the checks are not passing 🥺 
We can't merge the PR if the checks aren't passing. 

Once the failing-checks-PR gets merged, we can finally have this merged 🎉

Issue #384, opened 2021-06-13
Create a model to retrieve users of RDS with roles of member


harshith-venkatesh:
With latest changes in the member data model, new query to be generated to fetch all the users of RDS with roles of member.|

![image](https://user-images.githubusercontent.com/44545218/121796311-9af4ce80-cc35-11eb-8587-6130493d2bcc.png)

PR #385, opened 2021-06-13
Create QueryModel to fetch users of RDS with roles of member

harshith-venkatesh:


harshith-venkatesh:
closes #384

harshith-venkatesh:
closes #341

models/members.js
```diff
@@ -98,8 +98,37 @@ const deleteIsMemberProperty = async () => {
   }
 }
 
+/**
+ * Fetches the data about our members with role member:true
+ * @return {Promise<userModel|Array>}
+ */
+
+const fetchOnlyMembers = async () => {
```
ankurnarkhede, line 106: How is this function different from the `fetchMembers` function defined already?
Can we reuse it?

ankurnarkhede (changes requested):

models/members.js
harshith-venkatesh: @ankurnarkhede , with the function `fetchMembers` we would be receiving the details of all the users along with their roles defined.
Upon scaling, we would rather be doing filter operation for millions of data which could be handled with db call while fetching required details right thus reducing computational time.
Open to suggestion

ankurnarkhede:
In that case, the names of the functions needs to be changed. If the
function fetchMembers meant to fetch all users, not sure why the name is
fetchMembers and not fetchUsers. 🤔 Can you check if the functions using
the fetchMembers function actually need all users or just members, and
modify the function names accordingly? There doesn't seem any difference in
fetchMembers and fetchOnlyMembers for the reader 😅.

models/members.js
harshith-venkatesh: We do have `fetchUsers` in users model due to which I went with different way of naming convention...
Shall I proceed with changing naming convention of `fetchMembers` to `fetchMembersWithAllRoles` and `fetchOnlyMembers` to `fetchMembersWithMemberRole` ?

sumitd94: just go with `fetchUsers`

ankurnarkhede:
If the functions consuming them just require members, we can simply have
one function of fetchMembers and use it elsewhere.
Else, you can better create a function like fetchUsersWithRole(role=<role>)
and use it across the app.

harshith-venkatesh:
@ankurnarkhede @sumitd94 changes are updated please review

constants/users.js
```diff
@@ -1,2 +1,8 @@
 const userStatusEnum = ['ooo', 'idle', 'active']
-module.exports = userStatusEnum
+const ROLES = {
+  APPOWNER: 'app_owner',
+  DEFAULT: 'default',
+  MEMBER: 'member',
+  SUPERUSER: 'super_user'
+}
```
swarajpure, line 7: Are we going to have `admin` role?

harshith-venkatesh: These were the available roles that I had added
If we have documented about having admin role
I will add it accordingly @swarajpure

sumitd94 (dismissed):
Looks good to me!

controllers/members.js
```diff
@@ -10,7 +11,7 @@ const tasks = require('../models/tasks')
 
 const getMembers = async (req, res) => {
   try {
-    const allMembers = await fetchMembers()
+    const allMembers = await fetchUsers()
```
bhk31, line 14: What do you think `allMembers` is better or `users` or `allUsers` ?

controllers/members.js
```diff
@@ -31,9 +32,9 @@ const getMembers = async (req, res) => {
 
 const getIdleMembers = async (req, res) => {
   try {
-    const allMembers = await fetchMembers()
+    const onlyMembers = await fetchUsersWithRole(ROLES.MEMBER)
```
bhk31, line 35: `idleMembers` would be better nomenclature according to function name

controllers/members.js
harshith-venkatesh: I would go with `allMembers` here as we are actually assigning isMember value true or false property to it.

bhk31: We won't need to change API contracts and UI, just need to change variable name 

   ```
const allUsers = await fetchUsers()
     return res.json({
      message: allUsers.length ? 'Members returned successfully!' : 'No member found',
      members: allUsers
    })
```

harshith-venkatesh: I did understand the approach you had meant, I was not convinced enough with the change in variable names from allMembers to allUsers so far.

harshith-venkatesh: @bhk31 , I might be wrong here and would definitely stand to be corrected

bhk31: @harshith-venkatesh We can go ahead with either of the way. I don't have any problem with your approach. We can have opinion of @swarajpure @sumitd94 @ankurnarkhede  

@swarajpure @sumitd94 @ankurnarkhede Let us know your opinion.

harshith-venkatesh: It is actually a simple PR that is getting extended by a long time leading to prod issue of null data being observed.
I would rather make the changes to get this PR resolved as I am feeling bad looking at the idle member page being in that state

sumitd94 (approved):

Issue #366, opened 2021-05-31
Update task constants


whyDontI:
Add following statuses in [task constants](https://github.com/Real-Dev-Squad/website-backend/blob/develop/constants/tasks.js)
`active | assigned | blocked | completed | unAssigned`

Check [this](https://github.com/Real-Dev-Squad/website-api-contracts/issues/55) issue for more context

ritikrana4:
I will like to work on this.

swarajpure:
> I will like to work on this.

Assigned to you!

PR #389, opened 2021-06-17
Added a TASK_STATUS  - unassigned to constants

ritikrana4:
Updated the TASK_STATUS constant with a new value "UNASSIGNED: unAssigned".

ISSUE: Update task constants #366

@swarajpure 
@whyDontI

### DIFF
--- a/constants/tasks.js
+++ b/constants/tasks.js
@@ -8,7 +8,8 @@ const TASK_STATUS = {
   ACTIVE: 'active',
   ASSIGNED: 'assigned',
   BLOCKED: 'blocked',
-  COMPLETED: 'completed'
+  COMPLETED: 'completed',
+  UNASSIGNED: 'unAssigned'
 }
 
 module.exports = { TASK_TYPE, TASK_STATUS }
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -33,8 +33,8 @@ const getIdleMembers = async (req, res) => {
   try {
     const allMembers = await fetchMembers()
     const taskParticipants = await tasks.fetchActiveTaskMembers()
-    const idleMembers = allMembers?.filter(({ id }) => !taskParticipants.has(id))
-    const idleMemberUserNames = idleMembers?.map((member) => member.username)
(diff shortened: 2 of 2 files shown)


controllers/members.js
```diff
@@ -33,8 +33,8 @@ const getIdleMembers = async (req, res) => {
   try {
     const allMembers = await fetchMembers()
     const taskParticipants = await tasks.fetchActiveTaskMembers()
-    const idleMembers = allMembers?.filter(({ id }) => !taskParticipants.has(id))
-    const idleMemberUserNames = idleMembers?.map((member) => member.username)
+    const idleMembers = allMembers.filter(({ id }) => !taskParticipants.has(id))
+    const idleMemberUserNames = idleMembers.map((member) => member.username)
```
swarajpure, line 37: Why was this removed? We need those checks in place
It's not by mistake, it's called optional chaining: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

You maybe having older version of Node, that's why it didn't work in your machine, try updating node in your machine

constants/tasks.js
```diff
@@ -8,7 +8,8 @@ const TASK_STATUS = {
   ACTIVE: 'active',
   ASSIGNED: 'assigned',
   BLOCKED: 'blocked',
-  COMPLETED: 'completed'
+  COMPLETED: 'completed',
+  UNASSIGNED: 'unAssigned'
```
swarajpure, line 12: Cool!

swarajpure (changes requested):
Please revert the change of removing optional chaining operators, we can then merge this PR 🙂

controllers/members.js
ritikrana4: ok. Did not know about that. 👍

ritikrana4:
Reverted the changes and has added optional chaining. 
@swarajpure Please review.

swarajpure (approved):
Thanks for the change!

You might want to join our Discord server where we (Real Dev Squad) have all the technical and non-technical discussions. You can get more tasks, increase your knowledge and upskill yourself 😉 

Join discord server here: https://discord.gg/a3hBt2PZ

PR #402, opened 2021-06-28
Removed generateUrl method

prakashchoudhary07:


swarajpure (approved):
Thanks for the fix! ⚡

PR #406, opened 2021-06-30
Add a system for feature rollout for all RDS services

Cartmanishere:
The system supports rolling out a feature in 3 different ways:

1. Roles Based Rollout
    - Enable feature for all users having certain roles.
2. Percentage Rollout
    - Enable feature for a certain percentage of users.
3. Toggle Rollout
    - Turn feature on or off for everyone

Screenshot 2022-01-15 at 11 43 28 AM


models/featureFlags.js
```diff
@@ -1,13 +1,14 @@
...
+const { defaultConfig } = require('../constants/featureFlag')
 
 /**
  * Fetch all tasks
- *
+ * @param getOwnerInfo: Fetch the owner info for all the feature flags if true.
  * @return {Promise<featureFlags|Array>}
  */
-const fetchFeatureFlag = async () => {
+const fetchFeatureFlag = async (getOwnerInfo = true) => {
```
ankushdharkar, line 11: Why do we want to have this option? What are the benefits of getting the `ownerInfo` of who set the flag?

controllers/featureFlags.js
```diff
@@ -82,9 +84,33 @@ const deleteFeatureFlag = async (req, res) => {
...
 
+/**
+ * Get featureFlag config.
+ *
+ * This returns a list of feature flags and whether they are enabled or not.
+ * Feature flag being enabled depends on rollout config.
+ */
+const getConfig = async (req, res) => {
+  try {
+    const featureFlags = await featureFlagQuery.fetchFeatureFlag(false)
```
ankushdharkar, line 95: Should the function be plural? `fetchFeatureFlags`, like the LHS variable name?

middlewares/authenticate.js
```diff
@@ -83,13 +84,31 @@ module.exports = async (req, res, next) => {
...
+/**
+ * Check if authentication is present. If the request contains user auth, then add user details to the req object.
+ * If user is not logged in, just call the next handler.
+ * @param {Object} req
+ * @param {Object} res
+ * @param {function} next
+ */
+const maybeAuthenticate = async (req, res, next) => {
+  authenticate(req, res, next, false)
+}
```
ankushdharkar, line 109: Are we sure this is doing a conditional check as described in the comments above?

middlewares/authenticate.js
```diff
@@ -83,13 +84,31 @@ module.exports = async (req, res, next) => {
...
-        return res.boom.unauthorized('Unauthenticated User')
       }
-    } else {
-      return res.boom.unauthorized('Unauthenticated User')
     }
+
+    // If auth is not strictly required, just call the next handler.
+    if (!strict) {
+      return next()
+    }
```
ankushdharkar, line 94: I feel this doesn't belong in this function. The `authenticate` function should ideally have the singular responsibility to just authenticate.
Whether to call it or not, should be the responsibility of the callsite.

test/integration/featureFlag.test.js
```diff
@@ -120,6 +118,8 @@ describe('FeatureFlag', function () {
         })
     })
 
+    // TODO: Add test cases to update all the possible updates to the featureFlag config
```
ankushdharkar, line 121: Please add the ticket here for reference to track

utils/rollout.js
```diff
@@ -0,0 +1,111 @@
+const crypto = require('crypto')
+
+const generateUUID = () => {
+  return crypto.randomBytes(8).toString('hex')
```
ankushdharkar, line 4: Does this guarantee time-based uniqueness?

ankushdharkar: Or what about uniqueness in a distributed environment?

utils/rollout.js
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * 3 types of rollout configs are supported (in priority order):
+ *   a. Role based rollout:
+ *     - A list of roles is provided. If the user belongs to any of the roles, then feature is enabled.
+ *       Note: If userData is provided but doesn't match role, we don't fallback to other rollout but disable the feature.
+ *     - If userData is missing, it will fallback to the next type of rollout config.
+ *   b. Percentage rollout:
+ *     - Rollout to a certain percentage of users.
+ *     - User details are used to make this rollout monotonic.
+ *     - If usrData is missing, then the rollout is done randomly.
```
ankushdharkar, line 68: Nit: Typo

utils/rollout.js
```diff
@@ -0,0 +1,111 @@
...
+const hashCode = (str) => {
+  let hash = 0
+  for (let i = 0; i < str.length; i++) {
+    hash = Math.imul(31, hash) + str.charCodeAt(i) | 0
+  }
+  return Math.abs(hash)
+}
+
+/**
+ * Check if the user has any one of the roles that are allowed for this feature.
```
ankushdharkar, line 24: Would it make sense to make both roles the same datatype?

utils/rollout.js
```diff
@@ -0,0 +1,111 @@
...
+const featureFlagRollout = (featureFlag, userData) => {
+  if (!featureFlag.config) {
+    throw Error('Feature flag config is absent')
+  }
+
+  const { roleBased, percentage, enabled } = featureFlag.config
+
+  // Check if roleBased rollout can be used
+  if (roleBased && roleBased.active && userData) {
+    return roleBasedRollout(roleBased.roles || [], userData.roles || {})
```
ankushdharkar, line 98: Should the function manage missing values?

utils/rollout.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // Check if percentage rollout can be used
+  if (percentage && percentage.active) {
```
ankushdharkar, line 102: Nit: [Optional Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)?

ankushdharkar (changes requested):

controllers/featureFlags.js
Cartmanishere: Yes

middlewares/authenticate.js
Cartmanishere: `authenticate` isn't just a function, it's a middleware. 

I can refactor the core auth logic into a separate function and call that optionally in the middleware based on flag.

middlewares/authenticate.js
Cartmanishere: I tested it manually.

I'll write a test 😄

models/featureFlags.js
Cartmanishere: The same function can be called for doing the CRUD of feature flags.

On the internal dashboard for feature flags we can show who created the flag along with it's current status.

Ideally, I also want to track who updated as well. But beyond the scope of this PR. It can be added later.

test/integration/featureFlag.test.js
Cartmanishere: Ok

utils/rollout.js
Cartmanishere: Neither of those is a requirement for this usecase. We don't need rigorous randomness.

That said, `randomBytes` is described as `cryptographically strong` so I believe it must be safe in those scenarios as well, though I did not find a way to confirm it. 

Refer: https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

utils/rollout.js
Cartmanishere: I feel like `array` is more suitable.

But the `user.roles` schema is a map in database (for reasons I cannot recall). 

It requires DB migration and changes in `authorization` middleware as well so I didn't bother with it.

utils/rollout.js
Cartmanishere: 👍

utils/rollout.js
Cartmanishere: This is done for defensive reasons.

Sometimes our DB data is not exactly as we expect. Some entries are not migrated and fields might be absent. Hence the required defaults.

Other option is throw exceptions but I think a WARN log should be fine as this is not a critical function.

utils/rollout.js
Cartmanishere: 👍

controllers/featureFlags.js
Cartmanishere: Done

middlewares/authenticate.js
Cartmanishere: Done

utils/rollout.js
Cartmanishere: Done

utils/rollout.js
Cartmanishere: Done

middlewares/authenticate.js
vinayak-trivedi: Is maybeAuthenticate a good naming convention?

controllers/featureFlags.js
```diff
@@ -1,4 +1,6 @@
 const featureFlagQuery = require('../models/featureFlags')
+const logger = require('../utils/logger')
```
swarajpure, line 2: I thought we had logger declared globally for this project?

controllers/featureFlags.js
```diff
@@ -82,9 +84,33 @@ const deleteFeatureFlag = async (req, res) => {
...
+ *
+ * This returns a list of feature flags and whether they are enabled or not.
+ * Feature flag being enabled depends on rollout config.
+ */
+const getConfig = async (req, res) => {
+  try {
+    const featureFlags = await featureFlagQuery.fetchFeatureFlags(false)
+    return res.json({
+      data: featureFlags.map((featureFlag) => {
+        return {
```
swarajpure, line 98: Since we're already using arrow function here and not doing anything before return, can we simply use () => () to return it? I'm talking about omitting the `return` keyword

middlewares/authenticate.js
```diff
@@ -26,6 +26,61 @@ const checkRestricted = async (req, res, next) => {
...
+ * @param res
+ * @returns {Promise<{res, isTokenRefreshed: boolean, req}>}
+ */
+const refreshToken = async (req, res) => {
+  const refreshTtl = config.get('userToken.refreshTtl')
+  const token = req.cookies[config.get('userToken.cookieName')]
+  const { userId, iat } = authService.decodeAuthToken(token)
+  const newToken = authService.generateAuthToken({ userId })
+  const rdsUiUrl = new URL(config.get('services.rdsUi.baseUrl'))
+  let isTokenRefreshed = false
```
swarajpure, line 64: This should be on the top right? This variable being after `newToken` sort of made me confused

middlewares/authenticate.js
swarajpure: Yes we should probably rename this

middlewares/validators/featureFlags.js
```diff
@@ -25,8 +19,16 @@ const updateFeatureFlags = async (req, res, next) => {
   const schema = joi.object().strict().keys({
     title: joi.string().optional(),
     config: joi.object({
-      enabled: joi.boolean().required()
-    }).required()
+      roleBased: joi.object({
```
swarajpure, line 22: Can we use `object.strict` here?

models/featureFlags.js
```diff
@@ -80,17 +87,21 @@ const addFeatureFlags = async (featureFlag, username) => {
 const updateFeatureFlags = async (featureFlag, featureFlagId) => {
   try {
     const doc = await featureFlagModel.doc(featureFlagId).get()
-    if (!doc.data()) {
+    if (!doc.data() || Object.keys(featureFlag).length === 0) {
```
swarajpure, line 90: What if featureFlag itself is `null` or `undefined`?

controllers/featureFlags.js
akshay1502: Done. How do you catch such things? 🤔

middlewares/authenticate.js
akshay1502: @swarajpure @vinayak-trivedi could you suggest a better name for it.

Issue #316, opened 2021-04-11
Get members badges - badges/:username


sumitd94:
We already have an API that returns all the badges.
[here](https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/badges) is the API Contract that returns all the badges.

We need to write a new API, that will return the badges of the individual member

nilisha-jais:
I want to work on it

PR #407, opened 2021-07-21
Get member badges API

Vividh25:
(#316)
1. Added a new endpoint `/badges/:username` which returns the badge title and badge description of the badges held by a member.
2. Updated Swagger with the docs of the new API.
3. Added a simple test for the above endpoint.  

The local firestore database: 
rds-database

Request and Response: 
apicall

Swagger API Docs: 
swagger-api


controllers/badge.js
```diff
@@ -20,6 +20,20 @@ const getBadges = async (req, res) => {
   }
 }
 
+const getUserBadges = async (req, res) => {
+  try {
+    const allUserBadges = await badgeQuery.fetchUserBadges(req.params.username)
+    return res.json({
```
swarajpure, line 26: Need to handle the case when the username given is invalid

models/badges.js
```diff
@@ -27,6 +27,35 @@ const fetchBadges = async ({
...
+    const snapshot = await badgeModel.get()
+    const allBadges = []
+    snapshot.forEach((doc) => {
+      allBadges.push(doc.data())
+    })
+    const userBadges = []
+    allBadges.forEach((badge) => {
+      badge.users.forEach((user) => {
+        if (user === username) {
+          userBadges.push({ title: badge.title, description: badge.description })
```
swarajpure, line 47: So much of stuff can be avoided
You can use `array-contains` or similar query provided by firestore 😉

models/badges.js
```diff
@@ -27,6 +27,35 @@ const fetchBadges = async ({
...
+  try {
+    const snapshot = await badgeModel.get()
+    const allBadges = []
+    snapshot.forEach((doc) => {
+      allBadges.push(doc.data())
+    })
+    const userBadges = []
+    allBadges.forEach((badge) => {
+      badge.users.forEach((user) => {
+        if (user === username) {
```
swarajpure, line 46: Are we not going to use userId? What if someone changes their username? This API won't work for them
Check functions defined in utils/users for the conversion

routes/badges.js
```diff
@@ -34,4 +34,37 @@ const badge = require('../controllers/badge.js')
...
+ *              schema:
+ *                type: object
+ *                properties:
+ *                  message:
+ *                    type: string
+ *                    example: User badges returned successfully!
+ *                  userBadges:
+ *                    type: array
+ *                    items:
+ *                      $ref: '#/components/schemas/userBadges'
```
swarajpure, line 59: Need to add 404 for the case when username provided doesn't exist

swarajpure (changes requested):
Suggesting some changes, please check!

controllers/badge.js
```diff
@@ -20,6 +20,20 @@ const getBadges = async (req, res) => {
...
+      message: 'User badges returned successfully!',
+      userBadges: allUserBadges
```
harshith-venkatesh, line 28: You can add a check for userbadge length and can print similar messages

test/integration/userBadges.test.js
```diff
@@ -0,0 +1,26 @@
...
+chai.use(chaiHttp)
+
+describe('User badges', function () {
+  describe('GET /badges/:username', function () {
+    it('Should get the list of user badges', function (done) {
+      chai
+        .request(app)
+        .get('/badges/ankush')
+        .end((err, res) => {
+          if (err) { return done() }
```
harshith-venkatesh, line 16: Test returns done() for all errors here... Shouldn't we be handling 404 , 401?

Vividh25: Yes I'll add that.

controllers/badge.js
Vividh25: Yes I'll add that.

routes/badges.js
Vividh25: Yes will do that.

models/badges.js
Vividh25: The requirements specified that we use the username of the member 😅

models/badges.js
Vividh25: Ahh yes, I'll make the changes.

controllers/badge.js
Vividh25: Okay, I'll make the changes.

models/badges.js
harshith-venkatesh: It would be better to go with unique userid in my opinion, we can discuss on that

harshith-venkatesh: It would be better to go with unique userid in my opinion, we can discuss on that

harshith-venkatesh (changes requested):

models/badges.js
```diff
@@ -27,6 +27,35 @@ const fetchBadges = async ({
...
+ */
+
+const fetchUserBadges = async (username) => {
+  try {
+    const snapshot = await badgeModel.get()
+    const allBadges = []
+    snapshot.forEach((doc) => {
+      allBadges.push(doc.data())
+    })
+    const userBadges = []
```
sumitd94, line 43: Do we really need to first insert all data into `allBadges` array and then loop through it and then push it into `userBadges`?

On line no 41, we have the badge data - `doc.data()`
Why not directly check if the userid matches and there itself push it into `userBadges`.

Does this make sense? we can avoid the extra for loop call.

Vividh25: Yes it makes sense, I'll make the required changes.

controllers/badge.js
```diff
@@ -1,4 +1,5 @@
 const badgeQuery = require('../models/badges')
+const logger = require('../utils/logger')
```
swarajpure, line 2: We don't need to import logger. It is added to global object I suppose

controllers/badge.js
```diff
@@ -20,6 +21,29 @@ const getBadges = async (req, res) => {
...
 }
 
+const getUserBadges = async (req, res) => {
+  try {
+    const result = await badgeQuery.fetchUserBadges(req.params.username)
+    if (result.userExists) {
+      if (result.userBadges.length === 0) {
+        return res.json({
+          message: 'This user does not have any badges'
+        })
```
swarajpure, line 31: This code is looking necessarily nested and complex. Can we simply return an empty array (which will be implicit, we don't have to manually do anything there) and custom message, in case of 0 badges?
We can use ternary operator, so we'll be having better readability. What do you say? You can check how we're doing that in other controllers

models/badges.js
```diff
@@ -27,6 +28,34 @@ const fetchBadges = async ({
...
+      const snapshot = await badgeModel.get()
+      const userBadges = []
+      snapshot.forEach((item) => {
+        if (item.data().users.includes(userID)) {
+          userBadges.push({ title: item.data().title, description: item.data().description })
+        }
+      })
+      return { userExists: true, userBadges: userBadges }
+    }
+    return { userExists: false, userBadges: [] }
```
swarajpure, line 51: Can we simply have a default value for `userExists` and `userBadges` declared at let at the top block? Then modify it based on condition and return the values? So we'd have just 1 return statement.

routes/badges.js
```diff
@@ -34,4 +34,37 @@ const badge = require('../controllers/badge.js')
...
+/**
+ * @swagger
+ * /badges/:username:
+ *   get:
+ *     summary: Get all the badges of a particular user.
+ *
+ *     tags:
+ *       - Badges
+ *     responses:
+ *       200:
```
swarajpure, line 46: 404 case not mentioned here

test/fixtures/userBadges/userBadges.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * User's badges response
+ * Multiple responses can be added to the object if required
+ *
+ * @return {Object}
+ */
+
+const userFound = {
+  userExists: true,
+  userBadges: [{ title: 'badgeTitle', description: 'badgeDes' }]
```
swarajpure, line 10: We can use `badgeDescription` as whole. `badgeDes` might be confusing: design? descending?

test/integration/userBadges.test.js
```diff
@@ -0,0 +1,62 @@
...
+          expect(res).to.have.status(200)
+          expect(res.body).to.be.a('object')
+          expect(res.body.message).to.equal('User badges returned successfully!')
+          expect(res.body.userBadges).to.be.a('array')
+
+          return done()
+        })
+    })
+    it('Should return a not found message if the user in not found', function (done) {
```
swarajpure, line 33: `user **in** not found`

swarajpure (changes requested):
Great job in there!
Some suggestions for code cleanup. Please check!

controllers/badge.js
Vividh25: Got it!

controllers/badge.js
Vividh25: Yes let's go with the ternary operator, I'll make the changes.

models/badges.js
Vividh25: Okay makes sense, will do it.

routes/badges.js
Vividh25: Yes will add that

test/fixtures/userBadges/userBadges.js
Vividh25: Got it, will change it

test/integration/userBadges.test.js
Vividh25: Opps! will correct that.

models/badges.js
```diff
@@ -27,6 +28,35 @@ const fetchBadges = async ({
...
+      userExists = true
+      const userID = result.user.id
+      const snapshot = await badgeModel.get()
+      snapshot.forEach((item) => {
+        if (item.data().users.includes(userID)) {
+          userBadges.push({ title: item.data().title, description: item.data().description })
+        }
+      })
+    }
+    return { userExists: userExists, userBadges: userBadges }
```
swarajpure, line 52: We can use shorthand here?
```JS
{ userExists, userBadges }
```

models/badges.js
```diff
@@ -27,6 +28,35 @@ const fetchBadges = async ({
...
+  try {
+    const userBadges = []
+    let userExists = false
+    const result = await fetchUser({ username })
+    if (result.userExists) {
+      userExists = true
+      const userID = result.user.id
+      const snapshot = await badgeModel.get()
+      snapshot.forEach((item) => {
+        if (item.data().users.includes(userID)) {
```
swarajpure, line 47: Better to use optional chaining here? What if some item doesn't have data property? What if some data property doesn't have users? Could happen, right? API will break in that case

controllers/badge.js
```diff
@@ -20,6 +20,19 @@ const getBadges = async (req, res) => {
   }
 }
 
+const getUserBadges = async (req, res) => {
+  try {
+    const result = await badgeQuery.fetchUserBadges(req.params.username)
+    return result.userExists
+      ? res.json({ message: 'User badges returned successfully!', userBadges: result.userBadges })
+      : res.boom.notFound('The user does not exist')
```
swarajpure, line 28: `No badges found`?

swarajpure (changes requested):

models/badges.js
Vividh25: Yes will add it.

models/badges.js
Vividh25: Okay will add that

controllers/badge.js
Vividh25: Yes I added that, but I have used a mix of normal `if-else` and ternary operator. I have made a custom response message whose value will be dependent upon whether the `userBadges` array is empty or not. This reduced the number of return statements in the same `if` block.

swarajpure (dismissed):
Looks good to me.
Congrats on your first PR in backend 🎉 

Let @sumitd94 have a look once and then proceed for merging this one

models/badges.js
```diff
@@ -27,6 +28,35 @@ const fetchBadges = async ({
...
+    const result = await fetchUser({ username })
+    if (result.userExists) {
+      userExists = true
+      const userID = result.user.id
+      const snapshot = await badgeModel.get()
+      snapshot.forEach((item) => {
+        if (item.data()?.users?.includes(userID)) {
+          userBadges.push({ title: item.data().title, description: item.data().description })
+        }
+      })
```
sumitd94, line 50: We can improve this section of code for better readability. Do the following

- Add a line break before and after this `forEach` section
- on line no 48, this

```javascript
const { title, description } = item.data();
```

Then use this `title` & `description` like this - 
```javascript
userBadges.push({ title, description });
```

This is shorthand for object. If the key & value are same, we can use it the way I wrote. This internally means

```javascript
userBadges.push({ title:title, description:description });
```

sumitd94 (changes requested):
One small change and after that we can merge!

models/badges.js
Vividh25: Okay got it, will do it.

Vividh25: Regarding the line break, you mean before the `forEach` and after the `if` block right? It shows an indentation error if I directly add a line break after `forEach`.

models/badges.js
```diff
@@ -27,6 +28,38 @@ const fetchBadges = async ({
...
+      const userID = result.user.id
+      const snapshot = await badgeModel.get()
+
+      snapshot.forEach((item) => {
+        if (item.data()?.users?.includes(userID)) {
+          const { title, description } = item.data()
+          userBadges.push({ title, description })
+        }
+      })
+    }
```
sumitd94, line 53: Looks so clean ❤️

sumitd94 (approved):
Great work 🔥

PR #408, opened 2021-07-23
change self task api ready

akshay1502:


controllers/tasks.js
```diff
@@ -123,10 +123,28 @@ const updateTask = async (req, res) => {
   }
 }
 
+const updateTaskStatus = async (req, res) => {
```
swarajpure, line 126: JSDoc missing for the function

controllers/tasks.js
```diff
@@ -123,10 +123,28 @@ const updateTask = async (req, res) => {
...
+  try {
+    const taskId = req.params.id
+    const { username } = req.userData
```
swarajpure, line 129: We're not using username anywhere i guess

controllers/tasks.js
```diff
@@ -123,10 +123,28 @@ const updateTask = async (req, res) => {
...
+    const task = await tasks.fetchTask(taskId)
+
+    if (task.taskIdnotfound) return res.boom.notFound("Task doesn't exist")
+    if (username !== task.taskData.assignee) return res.boom.forbidden('This task is not assigned to you')
+
+    await tasks.updateTaskStatus(req.body, taskId)
+    return res.json({ message: 'Task updated successfully!' })
```
swarajpure, line 136: Wasn't this supposed to return 204? As of now it's returning 200

models/tasks.js
```diff
@@ -93,6 +108,7 @@ const fetchTask = async (taskId) => {
   try {
     const task = await tasksModel.doc(taskId).get()
     const taskData = task.data()
+    if (!taskData) return { taskIdnotfound: true }
```
swarajpure, line 111: `taskNotFound` would be better variable name i guess

routes/tasks.js
```diff
@@ -155,4 +155,51 @@ router.patch('/:id', authenticate, authorizeUser('appOwner'), updateTask, tasks.
...
+ *           schema:
+ *             $ref: '#/components/schemas/errors/notFound'
+ *     500:
+ *       description: badImplementation
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/components/schemas/errors/badImplementation'
+ */
+router.patch('/self/:id', authenticate, updateTask, tasks.updateTaskStatus)
```
swarajpure, line 203: We can't use the same validator `updateTask`
We need a separate validator maybe `updateSelfTask` and it should allow only status

swarajpure (changes requested):
Great job in there!
Hope you had good learning experience in there!

There are a LOT of unnecessary files, please delete all of them
Leaving some comments as well, please check them!

Thank you! 🥳

controllers/tasks.js
```diff
@@ -123,10 +123,33 @@ const updateTask = async (req, res) => {
...
+/**
+ * Updates self task status
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const updateTaskStatus = async (req, res) => {
+  try {
+    const taskId = req.params.id
+    const { id } = req.userData
+    const task = await tasks.fetchSelfTask(taskId, id)
```
swarajpure, line 135: `id` while being passed as function argument doesn't clearly tell which `id` it is
Lets rename it while it is being destructured to `userId` and then pass it ahead, would make much more sense then
https://wesbos.com/destructuring-renaming

middlewares/validators/tasks.js
```diff
@@ -68,7 +68,21 @@ const updateTask = async (req, res, next) => {
...
+  })
+  try {
+    await schema.validateAsync(req.body)
+    next()
+  } catch (error) {
+    logger.error(`Error validating updateSelfTask payload : ${error}`)
+    res.boom.badRequest(error.details[0].message)
+  }
+}
+
```
swarajpure, line 83: Perfect

models/tasks.js
```diff
@@ -94,6 +116,25 @@ const fetchTask = async (taskId) => {
...
+ * @param taskId { string }: taskId which will be used to fetch the task
+ * @param id { string }: id to check task is assigned to self or not
+ * @return {Promsie<taskData|Object>}
+ */
+const fetchSelfTask = async (taskId, id) => {
+  try {
+    const task = await tasksModel.doc(taskId).get()
+    const taskData = task.data()
+    if (!taskData) return { taskNotFound: true }
+    if (id !== taskData.assignee) return { notAssignedToYou: true }
```
swarajpure, line 130: Again, what's `id`? Let's call it `userId`

routes/tasks.js
```diff
@@ -155,4 +155,51 @@ router.patch('/:id', authenticate, authorizeUser('appOwner'), updateTask, tasks.
...
 router.get('/:username', tasks.getUserTasks)
 
+/**
+ * @swagger
+ * /tasks/self/:id:
+ *  patch:
+ *   summary: used to update self task status
+ *   tags:
+ *     - Tasks
+ *   requestBody:
```
swarajpure, line 165: are you sure this is `requestBody`?

models/tasks.js
```diff
@@ -35,6 +35,28 @@ const updateTask = async (taskData, taskId = null) => {
...
+    const task = await tasksModel.doc(taskId).get()
+    await tasksModel.doc(taskId).set({
+      ...task.data(),
+      ...taskData
+    })
+    return { taskId }
+  } catch (err) {
+    logger.error('Error in updating task status', err)
+    throw err
+  }
```
swarajpure, line 57: Tomorrow if we wish to allow them them to change task percentage, are we going to create a new function `updateTaskProgress`? We create a generalised function and then we decide what is to be passed in that function. We already have such function `updateTask`, let's use that only.

middlewares/validators/tasks.js
```diff
@@ -68,7 +68,21 @@ const updateTask = async (req, res, next) => {
   }
 }
 
+const updateSelfTask = async (req, res, next) => {
+  const schema = joi.object().keys({
+    status: joi.string().required()
+  })
+  try {
```
swarajpure, line 75: Okay I remembered just now that we also need to allow `percentCompleted` to be changed. It has to be a `number`. Please incorporate this small change as well.

swarajpure (changes requested):
Requesting some changes, please check.

Congrats on your first PR in backend 🥳

controllers/tasks.js
```diff
@@ -123,10 +123,33 @@ const updateTask = async (req, res) => {
...
+ * @param res {Object} - Express response object
+ */
+const updateTaskStatus = async (req, res) => {
+  try {
+    const taskId = req.params.id
+    const { id: userId } = req.userData
+    const task = await tasks.fetchSelfTask(taskId, userId)
+
+    if (task.taskIdNotFound) return res.boom.notFound("Task doesn't exist")
+    if (task.notAssignedToYou) return res.boom.forbidden('This task is not assigned to you')
```
swarajpure, line 138: Inconsistency in using quotes. Let's just stick to single quotes. You can escape the apostrophe, check how it's done in other routes ;)
Hint: you'll find a slash ;)

swarajpure (changes requested):
Some very minor things remaining now
Post those changes, this can be merged

routes/tasks.js
akshay1502: I checked in other routes, for post or patch it's requestBody. Am I missing some points here?

models/tasks.js
```diff
@@ -94,6 +94,25 @@ const fetchTask = async (taskId) => {
...
+ * @param taskId { string }: taskId which will be used to fetch the task
+ * @param id { string }: id to check task is assigned to self or not
+ * @return {Promsie<taskData|Object>}
+ */
+const fetchSelfTask = async (taskId, userId) => {
+  try {
+    const task = await tasksModel.doc(taskId).get()
+    const taskData = task.data()
+    if (!taskData) return { taskNotFound: true }
+    if (userId !== taskData.assignee) return { notAssignedToYou: true }
```
swarajpure, line 108: Using optional chaining here would make sense. What if a task doesn't have `assignee`?

routes/tasks.js
swarajpure: Cool then

swarajpure (changes requested):

models/tasks.js
swarajpure: I got a bit confused here. Sorry for the confusion.

swarajpure (approved):
Looks good to me!

Congratulations on your first PR in backend! 🔥

PR #410, opened 2021-07-25
Fixed validators to strictly follow the schema as it is converting string to number

lakshayman:


swarajpure (approved):
Looks good to me. Thanks!

PR #412, opened 2021-08-02
Populate participants/challenges

akshay1502:
if participants array contains more than 10 elements then code is using while loop to make a query with 10 elements of an array 
and iterate over the array till the end.

### DIFF
--- a/controllers/challenge.js
+++ b/controllers/challenge.js
@@ -59,7 +59,8 @@ const subscribeToChallenge = async (req, res) => {
     const subscribeUser = await challengeQuery.subscribeUserToChallenge(userId, challengeId)
     if (subscribeUser) {
       return res.status(200).json({
-        message: 'User has subscribed to challenge'
+        challenge_id: challengeId,
+        is_user_subscribed: 1
       })
     } else {
       return res.boom.notFound('User cannot be subscribed to challenge')
--- a/controllers/challenge.js
+++ b/controllers/challenge.js
@@ -13,13 +13,13 @@ const sendChallengeResponse = async (req, res) => {
     if (req.method === 'GET') {
       const allChallenges = await challengeQuery.fetchChallenges()
       const challengesWithParticipants = await getParticipantsofChallenges(allChallenges)
-      if (challengesWithParticipants) {
+      if (challengesWithParticipants.length > 0) {
         return res.status(200).json({
           message: 'Challenges returned successfully!',
           challenges: challengesWithParticipants
         })
       } else {
-        return res.boom.notFound('No challenges found')
+        return res.status(204).send()
       }
     } else {
       if (req.method === 'POST') {
--- a/controllers/challenge.js
+++ b/controllers/challenge.js
@@ -13,14 +13,10 @@ const sendChallengeResponse = async (req, res) => {
     if (req.method === 'GET') {
       const allChallenges = await challengeQuery.fetchChallenges()
       const challengesWithParticipants = await getParticipantsofChallenges(allChallenges)
-      if (challengesWithParticipants.length > 0) {
-        return res.status(200).json({
-          message: 'Challenges returned successfully!',
-          challenges: challengesWithParticipants
-        })
-      } else {
-        return res.status(204).send()
-      }
+      return res.status(200).json({
+        message: challengesWithParticipants.length ? 'Challenges returned successfully!' : 'No Challenges found',
+        challenges: challengesWithParticipants
+      })
     } else {
       if (req.method === 'POST') {
         const challengeAdded = await challengeQuery.postChallenge(req.body)
--- a/controllers/challenge.js
+++ b/controllers/challenge.js
@@ -12,17 +12,11 @@ const sendChallengeResponse = async (req, res) => {
   try {
     if (req.method === 'GET') {
       const allChallenges = await challengeQuery.fetchChallenges()
-      const updatedChallengesData = await Promise.all(allChallenges.map(async (challenge) => {
-        const participants = await challengeQuery.fetchParticipantsData(challenge.participants)
-        return {
-          ...challenge,
-          participants
-        }
-      }))
-      if (updatedChallengesData) {
+      const challengesWithParticipants = await getParticipantsofChallenges(allChallenges)
+      if (challengesWithParticipants) {
         return res.status(200).json({
           message: 'Challenges returned successfully!',
-          challenges: updatedChallengesData
+          challenges: challengesWithParticipants
         })
       } else {
         return res.boom.notFound('No challenges found')
@@ -45,6 +39,20 @@ const sendChallengeResponse = async (req, res) => {
     logger.error(`Error while retriving challenges ${err}`)
     return res.boom.serverUnavailable(ERROR_MESSAGE)
   }
+}
+
+/**
+ * @param {Array} allChallenges
+ * @returns {Promise<participants|Array>}
+ */
+const getParticipantsofChallenges = async (allChallenges) => {
+  return await Promise.all(allChallenges.map(async (challenge) => {
+    const participants = await challengeQuery.fetchParticipantsData(challenge.participants)
+    return {
+      ...challenge,
+      participants
+    }
+  }))
 }
 
 /**
--- a/controllers/challenge.js
+++ b/controllers/challenge.js
@@ -12,8 +12,8 @@ const sendChallengeResponse = async (req, res) => {
   try {
     if (req.method === 'GET') {
       const allChallenges = await challengeQuery.fetchChallenges()
-      const challengesWithParticipants = await getParticipantsofChallenges(allChallenges)
-      return res.status(200).json({
+      const challengesWithParticipants = await Promise.all(await getParticipantsofChallenges(allChallenges))
+      return res.json({
         message: challengesWithParticipants.length ? 'Challenges returned successfully!' : 'No Challenges found',
         challenges: challengesWithParticipants
       })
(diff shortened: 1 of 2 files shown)


models/challenges.js
```diff
@@ -35,6 +36,33 @@ const fetchChallenges = async () => {
...
+ * Fetch the <user object> from participants array
+ * @param {Array} participants
+ * @returns {Promise<challengesModel|Array>}
+ */
+const fetchParticipants = async (participants) => {
+  try {
+    const fetchedparticipants = []
+    while (participants.length) {
+      const currentParticipants = participants.splice(0, 10)
+      const getParticipants = await userModel.where(admin.firestore.FieldPath.documentId(), 'in', currentParticipants).get()
```
swarajpure, line 49: what's `admin.firestore.FieldPath.documentId()` doing?

models/challenges.js
```diff
@@ -66,10 +94,10 @@ const postChallenge = async (challengeData) => {
 const subscribeUserToChallenge = async (userId, challengeId) => {
   try {
     const getUser = await userModel.doc(userId).get()
-    const user = getUser.data().github_display_name
+    const user = getUser.data()
     if (user) {
       const challengeRef = await challengesModel.doc(challengeId)
-      await challengeRef.update({ participants: Firestore.FieldValue.arrayUnion({ name: user }) })
+      await challengeRef.update({ participants: Firestore.FieldValue.arrayUnion({ id: userId }) })
```
swarajpure, line 100: What are you trying to do here?

models/challenges.js
```diff
@@ -5,6 +5,7 @@
 
 const Firestore = require('@google-cloud/firestore')
 const firestore = require('../utils/firestore')
+const admin = require('firebase-admin')
```
swarajpure, line 8: why?

controllers/challenge.js
```diff
@@ -12,10 +12,17 @@ const sendChallengeResponse = async (req, res) => {
   try {
     if (req.method === 'GET') {
       const allChallenges = await challengeQuery.fetchChallenges()
-      if (allChallenges.length > 0) {
+      const newChallenges = await Promise.all(allChallenges.map(async (challenge) => {
+        const getParticipants = await challengeQuery.fetchParticipants(challenge.participants)
```
swarajpure, line 16: If there are 1,000 participants
You're to going to make this loop 1,000 times
Internally each loop is going to 1000/10 = 100 times

You'll be running it 1,000,00 times?? 

Isn't that highly inefficient?

swarajpure (changes requested):
Needs discussion on this one.

I'm not able to understand what you're trying to achieve and the way you're doing it, seems too inefficient

models/challenges.js
akshay1502: as .doc() directly gives the document, but I wanted to search for document which resembles from any of the id from array, I used this.

models/challenges.js
akshay1502: as participants should store userIds and not username, so replacing user with id.

models/challenges.js
```diff
@@ -5,6 +5,7 @@
 
 const Firestore = require('@google-cloud/firestore')
 const firestore = require('../utils/firestore')
+const userQuery = require('./users')
```
swarajpure, line 8: Since we're using only `fetchUser` out of this file, can we just import that? Hint: destructuring

models/challenges.js
```diff
@@ -35,6 +36,27 @@ const fetchChallenges = async () => {
...
 
+/**
+ * Fetch the <user object> from participants array
+ * @param {Array} participants
+ * @returns {Promise<challengesModel|Array>}
+ */
+const fetchParticipants = async (participants) => {
+  try {
+    const fetchedparticipants = await Promise.all(participants.map(async (userId) => {
+      const { user } = await userQuery.fetchUser({ userId: userId })
```
swarajpure, line 47: `{ userId: userId }` can be shortened to just `{ userId }`, right?

models/challenges.js
swarajpure: How is this getting stored?
Objects inside the array?

controllers/challenge.js
swarajpure, line 16: A lot is happening here, maybe you could use one more variable to keep things clear?

models/challenges.js
```diff
@@ -35,6 +36,27 @@ const fetchChallenges = async () => {
   }
 }
 
+/**
+ * Fetch the <user object> from participants array
+ * @param {Array} participants
+ * @returns {Promise<challengesModel|Array>}
+ */
+const fetchParticipants = async (participants) => {
```
swarajpure, line 44: how about `fetchParticipantsData`?

controllers/challenge.js
swarajpure, line 16: Why not just call this `getParticipants` as `participants`?

swarajpure (changes requested):

controllers/challenge.js
```diff
@@ -12,10 +12,17 @@ const sendChallengeResponse = async (req, res) => {
...
     if (req.method === 'GET') {
       const allChallenges = await challengeQuery.fetchChallenges()
-      if (allChallenges.length > 0) {
+      const updatedChallengesData = await Promise.all(allChallenges.map(async (challenge) => {
+        const participants = await challengeQuery.fetchParticipantsData(challenge.participants)
+        return {
+          ...challenge,
+          participants
+        }
+      }))
```
swarajpure, line 21: A lot seems to be happening here, could please separate out `allChallenges.map` part?

swarajpure: Wait, why is the function name `updatedChallengesData` inside `GET`?

models/challenges.js
```diff
@@ -35,6 +36,27 @@ const fetchChallenges = async () => {
...
+const fetchParticipantsData = async (participants) => {
+  try {
+    const fetchedparticipants = await Promise.all(participants.map(async (userId) => {
+      const { user } = await fetchUser({ userId })
+      return {
+        ...user,
+        phone: undefined,
+        email: undefined
+      }
+    }))
```
swarajpure, line 53: Again, can we seperate out the map part?

swarajpure (changes requested):
Requesting some changes for readability reasons

controllers/challenge.js
```diff
@@ -12,10 +12,11 @@ const sendChallengeResponse = async (req, res) => {
   try {
     if (req.method === 'GET') {
       const allChallenges = await challengeQuery.fetchChallenges()
-      if (allChallenges.length > 0) {
+      const challengesWithParticipants = await getParticipantsofChallenges(allChallenges)
+      if (challengesWithParticipants) {
```
swarajpure, line 16: If there are no challenges, that isn't the user's fault, right? 4xx are user's fault. We should be showing 200 only if challenges are not found with a message that no challenges found

controllers/challenge.js
```diff
@@ -40,6 +41,20 @@ const sendChallengeResponse = async (req, res) => {
   }
 }
 
+/**
+ * @param {Array} allChallenges
+ * @returns {Promise<participants|Array>}
+ */
+const getParticipantsofChallenges = async (allChallenges) => {
+  return await Promise.all(allChallenges.map(async (challenge) => {
```
swarajpure, line 49: What about the case when allChallenges is [] (empty)?

akshay1502: I checked when we pass [] empty array it returns a 200 status with the message.

controllers/challenge.js
```diff
@@ -12,14 +12,11 @@ const sendChallengeResponse = async (req, res) => {
...
-      if (allChallenges.length > 0) {
-        return res.status(200).json({
-          message: 'Challenges returned successfully!',
-          challenges: allChallenges
-        })
-      } else {
-        return res.boom.notFound('No challenges found')
-      }
+      const challengesWithParticipants = await getParticipantsofChallenges(allChallenges)
+      return res.status(200).json({
```
swarajpure, line 16: No need of `.status(200)` it's by default 200 only

controllers/challenge.js
swarajpure: cool

models/challenges.js
```diff
@@ -35,6 +36,36 @@ const fetchChallenges = async () => {
...
+ */
+const fetchParticipantsData = async (participants) => {
+  try {
+    const fetchedparticipants = getParticipantsData(participants)
+    return fetchedparticipants
+  } catch (err) {
+    logger.error('Failed to get participated users', err)
+    throw err
+  }
+}
```
swarajpure, line 52: Do we really need this function?

models/challenges.js
```diff
@@ -35,6 +36,36 @@ const fetchChallenges = async () => {
...
+ */
+const getParticipantsData = async (participants) => {
+  return await Promise.all(participants.map(async (userId) => {
+    const { user } = await fetchUser({ userId })
+    return {
+      ...user,
+      phone: undefined,
+      email: undefined
+    }
+  }))
```
swarajpure, line 67: I was saying it could be like `Promise.all(array of promises returns from another function)`

models/challenges.js
```diff
@@ -35,6 +36,29 @@ const fetchChallenges = async () => {
...
+  try {
+    const promises = participants.map(async (userId) => {
+      const { user } = await fetchUser({ userId })
+      return {
+        ...user,
+        phone: undefined,
+        email: undefined
+      }
+    })
+    const fetchedparticipants = await Promise.all(promises)
```
swarajpure, line 54: Sweet!

controllers/challenge.js
```diff
@@ -12,14 +12,11 @@ const sendChallengeResponse = async (req, res) => {
...
       const allChallenges = await challengeQuery.fetchChallenges()
-      if (allChallenges.length > 0) {
-        return res.status(200).json({
-          message: 'Challenges returned successfully!',
-          challenges: allChallenges
-        })
-      } else {
-        return res.boom.notFound('No challenges found')
-      }
+      const challengesWithParticipants = await Promise.all(await getParticipantsofChallenges(allChallenges))
```
swarajpure, line 15: How about moving the thing inside the bracket to a variable and using that variable here?

swarajpure:
```JS
const promiseArray = await getParticipantsofChallenges(allChallenges)
const challengesWithParticipants = await Promise.all(promiseArray)
```

swarajpure (approved):
Looks good to me!

PR #416, opened 2021-08-10
test for Patch /tasks/self/:id

akshay1502:


test/integration/tasks.test.js
```diff
@@ -293,4 +293,54 @@ describe('Tasks', function () {
...
+    it('Should return Forbidden error if task is not assigned to self', async function () {
+      const { user: { id, username } } = await userModel.fetchUser({ username: 'akshay' })
+
+      const res = await chai
+        .request(app)
+        .patch(`/tasks/self/${taskId1}`)
+        .set('cookie', `${cookieName}=${authService.generateAuthToken({ userId: id })}`)
+
+      const { taskData: { assignee } } = await tasks.fetchTask(taskId1)
+      if (username !== assignee) {
```
swarajpure, line 340: Why is this line here? 😅

test/integration/tasks.test.js
```diff
@@ -293,4 +293,51 @@ describe('Tasks', function () {
...
+        .end((err, res) => {
+          if (err) { return done(err) }
+          expect(res).to.have.status(404)
+          expect(res.body.message).to.equal('Task doesn\'t exist')
+          return done()
+        })
+    })
+
+    it('Should return Forbidden error if task is not assigned to self', async function () {
+      const { user: { id } } = await userModel.fetchUser({ username: 'akshay' })
```
swarajpure, line 332: you'll first need to add this user in DB right?

akshay1502: I don't think we need to add user in DB.

test/integration/tasks.test.js
```diff
@@ -293,4 +294,55 @@ describe('Tasks', function () {
...
+          expect(res.body.message).to.equal('Task doesn\'t exist')
+          return done()
+        })
+    })
+
+    it('Should return Forbidden error if task is not assigned to self', async function () {
+      const { userId } = await userModel.addOrUpdate({
+        github_id: 'akshay1502',
+        username: 'akshay'
+      })
```
swarajpure, line 336: Better to use some user from the fixtures here?
Also please have a look at line 26, the function and the way to use the function

test/integration/tasks.test.js
```diff
@@ -293,4 +294,55 @@ describe('Tasks', function () {
...
+
+      const res = await chai
+        .request(app)
+        .patch(`/tasks/self/${taskId1}`)
+        .set('cookie', `${cookieName}=${authService.generateAuthToken({ userId })}`)
```
swarajpure, line 341: Please let's keep it a bit clean, we could compute the jwt somewhere else and use it from a variable?

test/integration/tasks.test.js
```diff
@@ -293,4 +294,53 @@ describe('Tasks', function () {
...
+      const jwt = authService.generateAuthToken({ userId })
+
+      const res = await chai
+        .request(app)
+        .patch(`/tasks/self/${taskId1}`)
+        .set('cookie', `${cookieName}=${jwt}`)
+
+      expect(res).to.have.status(403)
+      expect(res.body.message).to.equal('This task is not assigned to you')
+      await cleanDb()
```
swarajpure, line 343: Can we have this function call inside `after` block  in the TOP level describe block?

test/integration/tasks.test.js
```diff
@@ -293,4 +294,53 @@ describe('Tasks', function () {
         })
     })
   })
+
+  describe('PATCH /self/:id', function () {
```
swarajpure, line 298: Please add a test for no cookie case, as well

swarajpure (approved):
Hope you learnt about testing 🎉 
Great job in here! 🔥

PR #418, opened 2021-08-13
Adding a check for node version 14 or above

### LINKED ISSUE #417: Allow only node version 14 and above
Currently if someone uses node version 12, since node v12 doesn't support optional chaining, project breaks for them.
We can add a check on project level that we'll only support node v14 and above

rohan09-raj:
Closes #417


package.json
```diff
@@ -10,8 +14,10 @@
     "validate-setup": "node scripts/validateSetup.js",
     "test-integration": "sh scripts/tests/testIntegration.sh",
     "test-unit": "sh scripts/tests/testUnit.sh",
-    "test": "npm run lint && npm run test-integration && npm run test-unit",
-    "generate-api-schema": "node utils/generateAPISchema.js"
+    "test": "npm run lint && npm run test-integration && npm run test-unit && bundle exec rake cucumber && bundle exec rake rspec && bundle exec rake jasmine:ci",
```
harshith-venkatesh, line 17: Is cucumber.js needs to be installed for the execution of this script?

harshith-venkatesh: How does rspec and jasmine:ci helps here?

rohan09-raj: I am not too much familiar with the code written. I saw some of the code working same as I wanted that's why I just copy pasted from that site.
I will surely read about all the terms I have used there.

rohan09-raj: Also, I followed the steps as listed there, and there was no instruction of installing cucumber.js. So, I guess it is not required to install cucumber.js

prakashchoudhary07: I think node version checking should be done while validating the setup

I think ` bundle exec rake cucumber && bundle exec rake rspec && bundle exec rake jasmine:ci",` is not required, we are not using jasmine or cucumber or rspc?

package.json
```diff
@@ -10,8 +14,10 @@
...
+    "generate-api-schema": "node utils/generateAPISchema.js",
+    "requirements-check": "node check_node_version.js",
+    "postinstall": ":.\node_modules/bower/bin/bower install && npm run requirements-check"
```
prakashchoudhary07, line 20: Why are you using bower here?

prakashchoudhary07 (changes requested):

package.json
prakashchoudhary07, line 20: Why are you using bower here?

package.json
rohan09-raj: https://medium.com/@faith__ngetich/locking-down-a-project-to-a-specific-node-version-using-nvmrc-and-or-engines-e5fd19144245

@prakashchoudhary07  I referred to this site for completing this task and yes, I removed that line you have suggested as it was an unnecessary test.

package.json
rohan09-raj: I guess it's used here because after checking the node version, if it's greater than -v14 then we have to install the required dependencies and packages, so it does that only.

rohan09-raj: and for the above issue, all that cucumber jasmine or rspc, I removed that test from the code and committed again

rohan09-raj: @harshith-venkatesh @prakashchoudhary07 Please review it again, I have made the changes suggested

prakashchoudhary07: Please recheck, we are not using bower as a package manager here

prakashchoudhary07 (changes requested):

package.json
rohan09-raj: Yes, we are using npm right ?? It is a silly mistake.

rohan09-raj: @prakashchoudhary07 Changed the package manager from bower to npm. Please review

check_node_version.js
```diff
@@ -0,0 +1,6 @@
+const semver = require('semver')
+const { engines } = require('./package')
+const version = engines.node
+if (!semver.satisfies(process.version, version)) {
+  throw new Error(`The current node version${process.version} does not satisfy the required version ${version} .`)
+}
```
swarajpure, line 6: Please move this file to `scripts` folder and keep the filename in camelCase instead of snake_case

check_node_version.js
```diff
@@ -0,0 +1,6 @@
+const semver = require('semver')
+const { engines } = require('./package')
+const version = engines.node
+if (!semver.satisfies(process.version, version)) {
+  throw new Error(`The current node version${process.version} does not satisfy the required version ${version} .`)
```
swarajpure, line 5: NIT: what if we add a space between `node version` and `${process.version}`

swarajpure: NIT: what if we remove the space before the period/dot?

package.json
```diff
@@ -11,7 +15,9 @@
     "test-integration": "sh scripts/tests/testIntegration.sh",
     "test-unit": "sh scripts/tests/testUnit.sh",
     "test": "npm run lint && npm run test-integration && npm run test-unit",
-    "generate-api-schema": "node utils/generateAPISchema.js"
+    "generate-api-schema": "node utils/generateAPISchema.js",
+    "requirements-check": "node check_node_version.js",
+    "postinstall": ":.\node_modules/npm/bin/npm install && npm run requirements-check"
```
swarajpure, line 20: What's `:.\node_modules/npm/bin/npm install` doing? 🤔

swarajpure (changes requested):

package.json
harshith-venkatesh: Are we trying to access npm post build through this script?

rohan09-raj: @swarajpure @harshith-venkatesh Reverted all the changes and added a new code for the check as I encountered with much simpler way to add a node version check, Please review.

swarajpure (commented):
Since you haven't made any changes in packages, the package-lock file shouldn't change. Please delete your `node_modules` folder and package-lock and do npm install once after deleting this two things, then commit the new package lock

rohan09-raj:
I have created a new pull request for this issue because there were big file size issues and thought that this and the previous changes can cause problems in the future but later realized that this is something obvious because the when running npm install the node_modules folder was created and this was the reason of the increase in the size of local setup folder of this project.

Also, this enhancement issue has been merged into develop branch from the new pull request created

PR #419, opened 2021-08-19
Feature/node  v14 check

### LINKED ISSUE #417: Allow only node version 14 and above

rohan09-raj:
Closes #417


swarajpure (approved):
Looks good to me. Approving from my side.

Good job in here! 💯 

Let @harshith-venkatesh have a look once and then it can be merged.

harshith-venkatesh (approved):

rohan09-raj:
Thank you @swarajpure @harshith-venkatesh

Issue #426, opened 2021-10-14
Feature request: archive member. 


MehulKChaudhari:
`Super User` can archive RDS members who are listed in the new members section.

PR #427, opened 2021-10-14
Feat: Archive users/members from the members page

MehulKChaudhari:
PR for #426


models/members.js
```diff
@@ -148,8 +148,31 @@ const fetchUsersWithRole = async (role) => {
...
+ * @param userId { String }: User id of user to be modified
+ * @return { Object }: whether moveToMember was successful or not and whether user is already a member or not
+ */
+
+const addArchiveRoleToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user?.roles && user?.roles?.archived_member) return { isArchived: true, addArchiveRole: false }
+    const roles = user.roles ? { ...user.roles, archived_member: true } : { archived_member: true }
```
ankushdharkar, line 162: Nit: camelCase the key name?

MehulKChaudhari: Yes, Done. Thank you!

test/integration/members.test.js
```diff
@@ -185,4 +188,81 @@ describe('Members', function () {
...
+            expect(res).to.have.status(400)
+            expect(res.body).to.be.a('object')
+            expect(res.body.message).to.equal('User is already archived')
+
+            return done()
+          })
+      })
+    })
+
+    it('Should return 401 if user is not a super_user', function (done) {
```
akshay1502, line 249: Here, it should return 403 logically, right?

routes/members.js
```diff
@@ -1,6 +1,6 @@
 const express = require('express')
 const router = express.Router()
-const { getMembers, getIdleMembers, migrateUserRoles, deleteIsMember, moveToMembers } = require('../controllers/members')
+const { getMembers, getIdleMembers, migrateUserRoles, deleteIsMember, moveToMembers, archiveMembers } = require('../controllers/members')
```
akshay1502, line 3: We use every function from controllers. Let's not destructure it. Naming it as `member` would be great, and then using it as `members.___`. You can see it in other routes files for reference.

MehulKChaudhari: Yes, good point, I followed what others did, and if I changed it now we have to change a lot of code.

test/integration/members.test.js
MehulKChaudhari: Yes it should return `403` but whoever coded the middleware to check to the user has return status 401, So I have followed else we have to change it everywhere where we have used this middleware.

routes/members.js
akshay1502: we only have to make changes in the routes/member.js file. You can even use `find and replace` for speeding it up. And better to do changes now, in the future we might have to change 2x of code.

akshay1502: We just have to make changes in routes/member.js. You can use `find and replace` if you want to speed it up. And better to do changes now, in the future we might have to change 2x of code.

MehulKChaudhari: Yeah makes sense. Done. Thank you for pointing this out.

test/integration/members.test.js
akshay1502: @swarajpure could we do anything here with return response?

routes/members.js
akshay1502: @MehulKChaudhari do the same here the way you did for controllers/member, please. I think you forgot to do it here.

MehulKChaudhari: Done. Thank you

test/integration/members.test.js
swarajpure: Yes we need to make that change in the middleware itself, change should be in seperate PR

models/members.js
```diff
@@ -148,8 +148,31 @@ const fetchUsersWithRole = async (role) => {
...
+ * changes the role of a new user to member
+ * @param userId { String }: User id of user to be modified
+ * @return { Object }: whether moveToMember was successful or not and whether user is already a member or not
+ */
+
+const addArchiveRoleToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user?.roles && user?.roles?.archivedMember) return { isArchived: true, addArchiveRole: false }
```
sumitd94, line 161:
```suggestion
    if user?.roles?.archivedMember) return { isArchived: true, addArchiveRole: false }
```

test/integration/members.test.js
sumitd94: yes, please create a ticket and tackle it in a different PR

models/members.js
MehulKChaudhari: Done. Thank you.!

models/members.js
```diff
@@ -148,8 +148,31 @@ const fetchUsersWithRole = async (role) => {
...
+ * changes the role of a new user to member
+ * @param userId { String }: User id of user to be modified
+ * @return { Object }: whether moveToMember was successful or not and whether user is already a member or not
+ */
+
+const addArchiveRoleToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user?.roles?.archivedMember) return { isArchived: true, addArchiveRole: false }
```
akshay1502, line 161: Do we need `addArchiveRole` to be returned? Because we are not using it in the controller.

controllers/members.js
```diff
@@ -114,7 +114,33 @@ const deleteIsMember = async (req, res) => {
   }
 }
 
+/**
+ * Achives old member from new members list.
```
akshay1502, line 118: Typo *Archives

models/members.js
MehulKChaudhari: Oh yes, we don't need it. I added for better understanding. Is there any downside of returning extra value?

akshay1502: For personal use it's ok. But in production unnecessary code is not good.

MehulKChaudhari: Yeah makes sense. Removed it.

models/members.js
```diff
@@ -148,8 +148,31 @@ const fetchUsersWithRole = async (role) => {
...
+const addArchiveRoleToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user?.roles?.archivedMember) return { isArchived: true }
+    const roles = user.roles ? { ...user.roles, archivedMember: true } : { archivedMember: true }
+    await userModel.doc(userId).update({
+      roles
+    })
+    return { isArchived: false, addArchiveRole: true }
```
shubham-y, line 166: https://github.com/Real-Dev-Squad/website-backend/pull/427#discussion_r743883464
`addArchiveRole` can also be removed from here

shubham-y (changes requested):

models/members.js
MehulKChaudhari: Yes, Done

akshay1502 (dismissed):

public/apiSchema.json
```diff
@@ -1 +1,4 @@
...
...ts/schemas/errors/badImplementation"}}}}}}},"/wallet/":{"get":{"summary":"Gets the user wallet details","tags":["wallet"],"responses":{"200":{"description":"Return wallet","content":{"application/json":{"schema":{"$ref":"#/components/schemas/wallet"}}}},"401":{"description":"unAuthorized","content":{"application/json":{"schema":{"$ref":"#/components/schemas/errors/unAuthorized"}}}},"500":{"description":"badImplementation","content":{"application/json":{"schema":{"$ref":"#/components/schemas/errors/badImplementation"}}}}}}},"/wallet/:username":{"get":{"summary":"Gets the user wallet details for a username, if you are an authorized superuser","tags":["wallet"],"responses":{"200":{"description":"Return wallet","content":{"application/json":{"schema":{"$ref":"#/components/schemas/wallet"}}}},"401":{"description":"unAuthorized","content":{"application/json":{"schema":{"$ref":"#/components/schemas/errors/unAuthorized"}}}},"403":{"description":"forbidden","content":{"application/json":{"schema":{"$ref":"#/components/schemas/errors/forbidden"}}}},"500":{"description":"badImplementation","content":{"application/json":{"schema":{"$ref":"#/components/schemas/errors/badImplementation"}}}}}}}}}
```
shubham-y, line 4: This line is repeated in the file. @MehulKChaudhari  Please check.

MehulKChaudhari: We are not really using schema now ig.

MehulKChaudhari: Fixing it. Thank you

shubham-y (approved):

controllers/members.js
```diff
@@ -114,7 +114,33 @@ const deleteIsMember = async (req, res) => {
...
+*
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const archiveMembers = async (req, res) => {
+  try {
+    const { username } = req.params
+    const user = await fetchUser({ username })
+    if (user.userExists) {
```
swarajpure, line 128: Better to use optional chaining here?

models/members.js
```diff
@@ -148,8 +148,31 @@ const fetchUsersWithRole = async (role) => {
...
+ * changes the role of a new user to member
+ * @param userId { String }: User id of user to be modified
+ * @return { Object }: whether moveToMember was successful or not and whether user is already a member or not
+ */
+
+const addArchiveRoleToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    if (user?.roles?.archivedMember) return { isArchived: true }
```
swarajpure, line 161: Since we're having archivedMember: true in both the cases, what if we seperate out that as a variable so we reduce duplication?

test/integration/members.test.js
swarajpure, line 249: Let's not keep snake case (super_user) content in middle of an English statement, please keep it with normal spacing

MehulKChaudhari: yes done.

models/members.js
MehulKChaudhari: yes, we can do that.

controllers/members.js
MehulKChaudhari: Done.

swarajpure (approved):
Looks good to me
Great job 🔥

models/members.js
swarajpure: Where's the updated code?

swarajpure (changes requested):

models/members.js
```diff
@@ -148,8 +148,32 @@ const fetchUsersWithRole = async (role) => {
...
+ * @return { Object }: whether moveToMember was successful or not and whether user is already a member or not
+ */
+
+const addArchiveRoleToMembers = async (userId) => {
+  try {
+    const userDoc = await userModel.doc(userId).get()
+    const user = userDoc.data()
+    const archivedMember = true
+    if (user?.roles?.archivedMember) return { isArchived: true }
+    const roles = user.roles ? { ...user.roles, archivedMember } : { archivedMember }
```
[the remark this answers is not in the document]
MehulKChaudhari: yes makes sense. We need a object of arrays. so yes

swarajpure (approved):
Great job in here 🔥 

Thanks for the tests ⚡

PR #440, opened 2021-11-05
tests for challenges api

### LINKED ISSUE #431: write tests for challenges API

Ishikabaid:
closes #431

### DIFF
--- a/test/integration/challenges.test.js
+++ b/test/integration/challenges.test.js
@@ -29,6 +29,7 @@ describe('Challenges', function () {
       chai
         .request(app)
         .get('/challenges')
+        .set('cookie', `${cookieName}=${jwt}`)
         .end((err, res) => {
           if (err) {
             return done(err)
@@ -45,11 +46,13 @@ describe('Challenges', function () {
       chai
         .request(app)
         .get('/challenges')
+        .set('cookie', `${cookieName}=${jwt}`)
         .end((err, res) => {
           if (err) {
             return done()
           }
-          expect(res).to.have.status(404)
+          expect(res).to.have.status(200)
+          expect(res.body.challenges).to.eql([])
           expect(res.body.message).to.be.equal('No Challenges found')
 
           return done()
@@ -63,7 +66,12 @@ describe('Challenges', function () {
         .request(app)
         .post('/challenges')
         .set('cookie', `${cookieName}=${jwt}`)
-        .send()
+        .send({
+          title: 'Test challenge-create',
+          level: 'easy',
+          start_date: 123,
+          end_date: 456
+        })
         .end((err, res) => {
           if (err) {
             return done(err)
@@ -82,6 +90,7 @@ describe('Challenges', function () {
       chai
         .request(app)
         .post('/challenges/subscribe')
+        .set('cookie', `${cookieName}=${jwt}`)
         .end((err, res) => {
           if (err) {
             return done(err)
--- a/test/integration/challenges.test.js
+++ b/test/integration/challenges.test.js
@@ -1,0 +1,111 @@
+const chai = require('chai')
+const { expect } = chai
+const chaiHttp = require('chai-http')
+
+const app = require('../../server')
+const authService = require('../../services/authService')
+const addUser = require('../utils/addUser')
+const cleanDb = require('../utils/cleanDb')
+
+const config = require('config')
+const cookieName = config.get('userToken.cookieName')
+
+chai.use(chaiHttp)
+
+describe('Challenges', function () {
+  let jwt
+
+  beforeEach(async function () {
+    const userId = await addUser()
+    jwt = authService.generateAuthToken({ userId })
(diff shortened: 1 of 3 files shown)


test/integration/challenges.test.js
```diff
@@ -0,0 +1,111 @@
...
+          }
+          expect(res).to.have.status(200)
+          expect(res.body).to.be.a('object')
+          expect(res.body.message).to.be.equal('Challenges returned successfully!')
+          expect(res.body.challenges).to.be.a('array')
+
+          return done()
+        })
+    })
+    it('Should return a not found message if the challenges is not found', function (done) {
```
swarajpure, line 44: Sure about this? Is the code written this way? 

If yes, it shouldn't be. 4xx errors are for client's mistake. If there's no challenge in DB, that isn't user's fault.

test/integration/challenges.test.js
```diff
@@ -0,0 +1,111 @@
...
+        })
+    })
+  })
+
+  describe('POST /subscribe', function () {
+    it('Should subscribe user to a challenge', function (done) {
+      chai
+        .request(app)
+        .post('/challenges/subscribe')
+        .end((err, res) => {
```
swarajpure, line 85: Do we not mention which challenge are we subscribing to? 🤔

test/integration/challenges.test.js
Ishikabaid: Yes, I've checked routes for challenges and it had 404 only.

test/integration/challenges.test.js
Ishikabaid: Yes, it should be `/challenges/subscribe/challengeId`

test/integration/challenges.test.js
akshay1502: @Ishikabaid we don't return 404 for 'No challenges'. You can see [here](https://github.com/Real-Dev-Squad/website-backend/blob/5ec05e86695a2e717030156e91f779f45c611adf/controllers/challenge.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L17). We check whether `challenges` variable is null or not, and according to that we return the msg and challenges.

test/integration/challenges.test.js
akshay1502: No. Api is right `challenges/subscribe`. We pass the `challengeId` in request body.

test/integration/challenges.test.js
```diff
@@ -0,0 +1,111 @@
...
+    })
+  })
+
+  describe('POST /challenges', function () {
+    it('Should create a new challenge', function (done) {
+      chai
+        .request(app)
+        .post('/challenges')
+        .set('cookie', `${cookieName}=${jwt}`)
+        .send()
```
akshay1502, line 66: We need to pass the requestBody for `/challenges` it can not be empty. You can see similar example [here](https://github.com/Real-Dev-Squad/website-backend/blob/5ec05e86695a2e717030156e91f779f45c611adf/test/integration/tasks.test.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L85)

Ishikabaid: okay

test/integration/challenges.test.js
```diff
@@ -0,0 +1,120 @@
...
+        .get('/challenges')
+        .set('cookie', `${cookieName}=${jwt}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err)
+          }
+          expect(res).to.have.status(200)
+          expect(res.body).to.be.a('object')
+          expect(res.body.message).to.be.equal('Challenges returned successfully!')
+          expect(res.body.challenges).to.be.a('array')
```
akshay1502, line 40: @Ishikabaid you need to add a dummy challenge here since DB is empty. You can refer [this](https://github.com/Real-Dev-Squad/website-backend/blob/1fbda36c6b1cc764a2b099d1431d5160c681b882/test/integration/tasks.test.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L198)

akshay1502, line 40: Also check if the array length is greater than 1 or not since we are checking for length in controller.

ankurnarkhede, line 40: Also assert on the individual values in case of 200 response here.

ankurnarkhede (changes requested):
Add a test for unauthorised too.

Issue #463, opened 2021-12-28
<Identity Service> Create /users/identityURL endpoint


lakshayman:
Create `/users/identityURL` route which serves a PATCH request having a required username param.
## Body
```
{
   identityURL: '<identityURL>'
}
```

It will set the identityURL in user's object.

** Response :  **
```
{
    message: 'updated identity URL!!'
    identityURL: <new identityURL>
}
```

For Reference: [here](https://github.com/Real-Dev-Squad/identity-service/issues/18)

vinayak-trivedi:
Hey, I will do it

PR #481, opened 2022-01-03
created the endpoint for identityURL

vinayak-trivedi:
closes https://github.com/Real-Dev-Squad/website-backend/issues/463


controllers/users.js
```diff
@@ -142,12 +142,26 @@ const postUserPicture = async (req, res) => {
     return res.boom.badImplementation('An internal server error occurred')
   }
 }
-
+const identityURL = async (req, res) => {
+  try {
+    const userId = req.userData.id
+    await userQuery.addOrUpdate(req.body, userId)
```
swarajpure, line 148: What should be the schema/structure for `req.body`?

harshith-venkatesh, line 148: What kind of validations will be done for URL?
Correct me if I am missing something here and sharing schema/ structure here would be helpful 👍🏻

vinayak-trivedi: {
   identityURL: "https://abc.com"
}
This is req.body

test/integration/users.test.js
```diff
@@ -227,6 +227,25 @@ describe('Users', function () {
...
+    })
+  })
+  describe('PATCH /users/identityURL', function () {
+    it('Should update the identityURL', function (done) {
+      chai
+        .request(app)
+        .patch('/users/identityURL')
+        .set('cookie', `${cookieName}=${jwt}`)
+        .send({
+          identityURL: 'identityURL'
```
swarajpure, line 241: This shouldn't be allowed right? We should allow only valid links. Tomorrow someone might send `joker` here, is that a valid request body? Please see if we can use this: https://joi.dev/api/?v=17.5.0#stringurioptions

lakshayman (dismissed):
Looks good to me

controllers/users.js
```diff
@@ -142,12 +142,26 @@ const postUserPicture = async (req, res) => {
...
+    const { user } = await userQuery.fetchUser({ userId })
```
ankurnarkhede, line 149: Not getting an error from `userQuery.addOrUpdate` itself means the value has been updated. We dont need to get the values from DB again here.

ankurnarkhede (changes requested):

ankur-lt (approved):

PR #482, opened 2022-01-05
add chaincode generation endpoint with integration test

pallabez:


routes/users.js
```diff
@@ -265,4 +265,6 @@ router.get('/:username', users.getUser)
 // upload.single('profile') -> multer inmemory storage of file for type multipart/form-data
 router.post('/picture', authenticate, upload.single('profile'), users.postUserPicture)
 
+router.patch('/chainCode', authenticate, users.updateChainCode)
```
swarajpure, line 268: What's the reason of going ahead with patch request here?

pallabez: We are updating the chain code. There might already be an existing chain code in the user object so we can't use POST here.

pallabez: The endpoint is used for generating a new chain code for that user, regardless of if the person have a pre existing chain code or not.

controllers/users.js
```diff
@@ -143,11 +144,33 @@ const postUserPicture = async (req, res) => {
   }
 }
 
+/**
+ * Post user profile picture
```
harshith-venkatesh, line 148: Please add comment for `updateChainCode`

routes/users.js
harshith-venkatesh: `The endpoint is used for generating a new chain code for that user` this is bit contradicting the above statement here

harshith-venkatesh (changes requested):

controllers/users.js
pallabez: Right. Thanks for pointing it out.

controllers/users.js
```diff
@@ -145,19 +145,19 @@ const postUserPicture = async (req, res) => {
...
  * @param req {Object} - Express request object
  * @param res {Object} - Express response object
  */
-const updateChainCode = async (req, res) => {
+const updateChaincode = async (req, res) => {
   try {
     const key = uuid.v4()
     const { id: userId } = req.userData
-    await userQuery.addOrUpdate({ chainCode: key }, userId)
+    await userQuery.addOrUpdate({ chaincode: key }, userId)
```
lakshayman, line 157: @pallabez can you remove this line as it is not needed
We discussed in last call

lakshayman (changes requested):

controllers/users.js
pallabez: got it. 👍

Issue #473, opened 2022-01-01
<identity service> Create /users/chainCode endpoint


lakshayman:
We have to generate a distinct chain code for the user every time this endpoint hits. User's cookies will be sent with a call.

Generate a chain code and send it back.

For Reference: https://github.com/Real-Dev-Squad/identity-service/issues/18

PR #486, opened 2022-01-08
add chaincode endpoint with integration test

pallabez:
closes https://github.com/Real-Dev-Squad/website-backend/issues/473

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -158,7 +158,7 @@ const updateChaincode = async (req, res) => {
       chaincode: key
     })
   } catch (error) {
-    logger.error(`Error while updating user: ${error}`)
+    logger.error(`Error while updating user's chaincode: ${error}`)
     return res.boom.serverUnavailable('Something went wrong please contact admin')
   }
 }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -1,6 +1,7 @@
 
 const userQuery = require('../models/users')
 const imageService = require('../services/imageService')
+const uuid = require('uuid')
 /**
  * Fetches the data about our users
  *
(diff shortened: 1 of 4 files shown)


controllers/users.js
```diff
@@ -143,11 +144,31 @@ const postUserPicture = async (req, res) => {
...
+ */
+const updateChaincode = async (req, res) => {
+  try {
+    const key = uuid.v4()
+    return res.json({
+      message: 'Generated New Chaincode!!',
+      chaincode: key
+    })
+  } catch (error) {
+    logger.error(`Error while updating user: ${error}`)
```
lakshayman, line 161:
```suggestion
    logger.error(`Error while updating user's chaincode: ${error}`)
```

lakshayman (dismissed):
Else Looks fine to me

lakshayman (approved):
Great! Good Work

shubham-y:
@pallabez Are we updating the newly generated chain code to the user document?

pallabez:
> @pallabez Are we updating the newly generated chain code to the user document?

No, we are going to store the newly generated chain code in a different document that hasn't been created yet.

shubham-y:
> > @pallabez Are we updating the newly generated chain code to the user document?
> 
> No, we are going to store the newly generated chain code in a different document that hasn't been created yet.
So for now, shouldn't we create a GET chain code endpoint as we are updating anything. What do you think?

pallabez:
> > > @pallabez Are we updating the newly generated chain code to the user document?
> > 
> > 
> > No, we are going to store the newly generated chain code in a different document that hasn't been created yet.
> 
> So for now, shouldn't we create a GET chain code endpoint as we are updating anything. What do you think?

Seems reasonable. I'll ask Lakshay about it.

lakshayman (approved):
Looks Good!

controllers/users.js
```diff
@@ -143,11 +144,31 @@ const postUserPicture = async (req, res) => {
...
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const getChaincode = (req, res) => {
+  try {
+    const key = uuid.v4()
+    return res.json({
+      message: 'Generated New Chaincode!!',
+      chaincode: key
```
ankurnarkhede, line 158: How are we associating the code with a specific user? This controller is just generating a UUID v4, that we could have generated on the UI itself if we dont want to associate it with a user... 🤔

ankurnarkhede (changes requested):

controllers/users.js
ankushdharkar: Good catch @ankurnarkhede. We should store the chaincode in firestore and read the document id as a key (helps prevent potential collisions also).

@pallabez Can you add that as a comment please, so the next person (maybe you?) can remember to make the appropriate changes. Attach the ticket link in the comments please.

pallabez: Got it 👍

PR #488, opened 2022-01-11
new task status will be returned

akshay1502:
Updating task status to the new task workflow while returning.


Pavangbhat (changes requested):
Are we updating tasks with new task workflow status ? i.e
**active** → **IN_PROGRESS**, **unAssigned** → **AVAILABLE**

And also please note that the task is considered as active if it's in **BLOCKED**, **IN_PROGRESS**, **SMOKE_TESTING**

for reference ⬇️ 
image

prakashchoudhary07 (approved):
looks good to me :+1:

models/tasks.js
```diff
@@ -86,8 +90,11 @@ const fetchActiveTaskMembers = async () => {
 const fetchTask = async (taskId) => {
   try {
     const task = await tasksModel.doc(taskId).get()
-    const taskData = task.data()
-    return { taskData: await fromFirestoreData(taskData) }
+    const taskData = await fromFirestoreData(task.data())
+    if (taskData?.status) {
+      taskData.status = TASK_STATUS[taskData.status]
```
Pavangbhat, line 95: Currently, we have task status as **pending** and **High Priority** please those cases as well

Pavangbhat (changes requested):
Hi, Akshay great work  🎉 
There are a few changes needed please work on it.

Also, make sure the following scenarios are handled 

- [ ] we can hit **/tasks/username?status=active** where we need to return tasks that are in either of the status **[active, blocked, completed, pending, IN_PROGRESS, BLOCKED,  SMOKE_TESTING, COMPLETED]**

- [ ] idle members are those who don't have a task which is either in [active, blocked, pending, IN_PROGRESS, BLOCKED,  SMOKE_TESTING]

Pavangbhat:
This PR is continued with #514

PR #507, opened 2022-01-22
Fix create and update tasks API for invalid username

### LINKED ISSUE #445: API breaks if participants of tasks object has unknown username
The participants array stores array of **user_ids** where **user_ids** are generated by **username**(by the backend) and if the **username** does not exist then particular **user_id** would be false hence breaking the API

Below structure breaks the API for **/tasks** :point_down: 

**participants : [
"user_id",
 false,
]**

shubham-y:
Closes #445


MehulKChaudhari (dismissed):
Thanks for the tests Shubham. I see no changes here.

controllers/tasks.js
```diff
@@ -16,6 +16,10 @@ const addNewTask = async (req, res) => {
     };
     const task = await tasks.updateTask(body);
 
+    if (task.userNotFound) {
+      logger.error("Error while creating new task: Incorrect username passed");
+      return res.boom.badRequest("Unable to add task");
```
Pavangbhat, line 21: **Suggestion:** I think we should have a better message here maybe like user not found

controllers/tasks.js
```diff
@@ -116,7 +120,11 @@ const updateTask = async (req, res) => {
       return res.boom.notFound("Task not found");
     }
 
-    await tasks.updateTask(req.body, req.params.id);
+    const updateTaskResult = await tasks.updateTask(req.body, req.params.id);
+    if (updateTaskResult.userNotFound) {
+      logger.error(`Error while updating taskId ${req.params.id}: Incorrect username passed`);
+      return res.boom.badRequest("Unable to update task");
```
Pavangbhat, line 126: **Suggestion:** I think we should have a better message here

models/tasks.js
```diff
@@ -14,6 +14,9 @@ const { TASK_TYPE, TASK_STATUS } = require("../constants/tasks");
 const updateTask = async (taskData, taskId = null) => {
   try {
     taskData = await toFirestoreData(taskData);
+    if (taskData.userNotFound) {
```
Pavangbhat, line 17: Will we be able to update tasks if there is no **assignee**?

Pavangbhat (changes requested):
Hi @shubham-y Nice Work
I have left few comments please address them

utils/users.js
```diff
@@ -71,9 +71,11 @@ const getParticipantUserIds = async (participantArray) => {
 
     const promises = participantArray.map(async (participant) => {
       const participantUserId = await getUserId(participant.trim());
+      if (!participantUserId) return false;
       return participantUserId;
     });
-    const participantUserIds = await Promise.all(promises);
+    let participantUserIds = await Promise.all(promises);
+    participantUserIds = participantUserIds.filter(Boolean); // remove invalid users
```
Pavangbhat, line 78: Learned a new shortcut here 😃

models/tasks.js
shubham-y: Yes absolutely, there is a check in `toFireStoreData` function
https://github.com/Real-Dev-Squad/website-backend/blob/e4bb6ca1e299d15baf8d91db245e82335b077acf/utils/tasks.js#L37

controllers/tasks.js
shubham-y: Thanks for the suggestion. The message is updated to `User not found`

controllers/tasks.js
shubham-y: Thanks for the suggestion. The message is updated to `User not found`

Pavangbhat (dismissed):
Thankyou @shubham-y 
Everything Looks Good to me 🎉

models/tasks.js
ankushdharkar, line 17: Should this be here? Should the model be silently returning, instead of updating the Firestore?

ankushdharkar (changes requested):

models/tasks.js
shubham-y: Thanks for pointing it out. Moved this check to `controller/tasks.js`

test/integration/tasks.test.js
```diff
@@ -112,6 +113,67 @@ describe("Tasks", function () {
...
+          title: "Test task - Create",
+          type: "group",
+          endsOn: 123,
+          startedOn: 456,
+          status: "completed",
+          percentCompleted: 10,
+          completionAward: { [DINERO]: 3, [NEELAM]: 300 },
+          lossRate: { [DINERO]: 1 },
+          assignee: appOwner.username,
+          participants: ["dummyUser"],
```
akshay1502, line 158: It should give an error for an invalid participant name, right?

test/integration/tasks.test.js
```diff
@@ -112,6 +113,67 @@ describe("Tasks", function () {
...
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(400);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("User not found");
+          return done();
+        });
+    });
+    it("Should not return invalid username in participants", function (done) {
```
akshay1502, line 143: why aren't we checking for valid usernames in participants array?

utils/users.js
```diff
@@ -71,9 +71,11 @@ const getParticipantUserIds = async (participantArray) => {
 
     const promises = participantArray.map(async (participant) => {
```
akshay1502, line 72: What if we use filter over here directly, since getUserId will return an id(positive value) or `false`?

controllers/tasks.js
```diff
@@ -144,10 +153,24 @@ const updateTaskStatus = async (req, res) => {
...
+    if (task.notAssignedToYou) {
+      logger.error(`Error while updating taskId ${req.params.id}: Forbidden user`);
+      return res.boom.forbidden("This task is not assigned to you");
+    }
+    if (task.taskData.status === "VERIFIED") {
+      logger.error(`Error while updating taskId ${req.params.id}: Already verified task`);
       return res.boom.forbidden("Status cannot be updated. Please contact admin.");
+    }
+
+    const taskData = await toFirestoreData(req.body);
```
akshay1502, line 169: @shubham-y can we have a separate function for lines 169 to 173, if it's going to be used frequently in the future, as of now we are using it 3 times. What do you think?

test/integration/tasks.test.js
shubham-y: As per my discussion with @ankushdharkar, we decided to filter out invalid usernames if passed in the participants array and throw a 400 Bad Request response if invalid username is passed as an assignee

test/integration/tasks.test.js
shubham-y: Refer - https://github.com/Real-Dev-Squad/website-backend/pull/507#discussion_r856157739

test/integration/tasks.test.js
akshay1502: okay.

controllers/tasks.js
shubham-y: I think separating the logic for logging and returning a response in a different function other than the controller function is not recommended

utils/users.js
shubham-y: But the async function returns a Promise rather than a Boolean value. So we cannot directly replace map to filter

akshay1502 (dismissed):

controllers/tasks.js
```diff
@@ -144,10 +153,24 @@ const updateTaskStatus = async (req, res) => {
...
-    if (task.taskData.status === "VERIFIED")
+    if (task.taskNotFound) {
+      logger.error(`Error while updating taskId ${req.params.id}: Invalid taskId passed`);
+      return res.boom.notFound("Task doesn't exist");
+    }
+    if (task.notAssignedToYou) {
+      logger.error(`Error while updating taskId ${req.params.id}: Forbidden user`);
+      return res.boom.forbidden("This task is not assigned to you");
+    }
+    if (task.taskData.status === "VERIFIED") {
```
ankushdharkar, line 164: Let's have this from a constants file please

ankushdharkar (changes requested):

controllers/tasks.js
shubham-y: Thanks for pointing it out. Refactored all the places where task status was not imported from constants file

prakashchoudhary07 (approved):
LGTM :+1:

Issue #464, opened 2021-12-28
[ New Task Workflow ] New Task workflow changes 


Pavangbhat:
**Details :-**  https://github.com/Real-Dev-Squad/website-backend/issues/460

**Task :-**
Task status should be in the upper case
- [x] Update task status list in models and handle of side effects ( can be a sub issues)

akshay1502:
I would like to work on this.

Pavangbhat:
Assigned to @akshay1502

Issue #465, opened 2021-12-28
[ New Task Workflow ] Create a endpoint to update task's status based on ETA


Pavangbhat:
Needs Clarification 

**Details :-** https://github.com/Real-Dev-Squad/website-backend/issues/460

**Tasks :-**

- [x] create an endpoint that fires a function that loops through all tasks whose status is **ASSIGNED | IN_PROGRESS** and if **endsOn** > **currentTime** SET task’s status back to **AVAILABLE**, **assignee** field to be cleared and log the member and task details (For future use)

PR #514, opened 2022-01-27
Feat/new task workflow

akshay1502:
closes: #464 and #465 
added endpoint for getting overdue tasks as available.


constants/tasks.js
```diff
@@ -5,11 +5,23 @@ const TASK_TYPE = {
...
+  SMOKE_TESTING: 'SMOKE_TESTING',
+  NEEDS_REVIEW: 'NEEDS_REVIEW',
+  IN_REVIEW: 'IN_REVIEW',
+  APPROVED: 'APPROVED',
+  MERGED: 'MERGED',
+  SANITY_CHECK: 'SANITY_CHECK',
+  REGRESSION_CHECK: 'REGRESSION_CHECK',
+  RELEASED: 'RELEASED',
+  VERIFIED: 'VERIFIED'
 }
```
Pavangbhat, line 25: What can we do for the tasks that are currently set as **High Priority**  also the tasks that might not fit in this list in that case status will be undefined

Pavangbhat (changes requested):
Re:

- [x]  we can hit /tasks/username?status=active where we need to return tasks that are in either of the status [active, blocked, completed, pending, IN_PROGRESS, BLOCKED, SMOKE_TESTING, COMPLETED]

- [ ] idle members are those who don't have a task which is either in [active, blocked, pending, IN_PROGRESS, BLOCKED, SMOKE_TESTING]

constants/tasks.js
akshay1502: I am handling the `high priority` status in new commit.

akshay1502: I checked the current /task api and it has only 4 status values which we are handling already.

akshay1502:
> Re:
> 
>     * [x]   we can hit /tasks/username?status=active where we need to return tasks that are in either of the status [active, blocked, completed, pending, IN_PROGRESS, BLOCKED, SMOKE_TESTING, COMPLETED]
> 
>     * [ ]  idle members are those who don't have a task which is either in [active, blocked, pending, IN_PROGRESS, BLOCKED, SMOKE_TESTING]

@Pavangbhat do we have to show completed tasks in `?status=active` ?

models/tasks.js
```diff
@@ -174,9 +191,32 @@ const fetchUserActiveAndBlockedTasks = async (username) => {
...
+/**
+ * Fetch all overdue tasks
+ * @param overdueTasks <Array>: tasks which are overdue
+ * @return {Promsie<Array>}
+ */
+const overdueTasks = async (overDueTasks) => {
+  try {
+    const newAvailableTasks = await Promise.all(overDueTasks.map(async (task) => {
+      let { status, assignee, id } = task
+      status = TASK_STATUS.UNASSIGNED
```
Pavangbhat, line 206: Since we are moving with a new task workflow let's make this AVAILABLE

controllers/tasks.js
```diff
@@ -145,11 +153,33 @@ const updateTaskStatus = async (req, res) => {
...
+ * Fetches all the overdue tasks
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const overdueTasks = async (req, res) => {
+  try {
+    const allTasks = await tasks.fetchTasks()
+    const now = Math.floor(Date.now() / 1000)
+    const overDueTasks = allTasks.filter(task => (task.status === 'ASSIGNED' || task.status === 'IN_PROGRESS') && task.endsOn < now)
```
Pavangbhat, line 166: We also need to consider pending here

Pavangbhat (changes requested):
Needs some tiny changes 
Everything else looks good to me 🎉

models/tasks.js
```diff
@@ -86,8 +91,11 @@ const fetchActiveTaskMembers = async () => {
 const fetchTask = async (taskId) => {
   try {
     const task = await tasksModel.doc(taskId).get()
-    const taskData = task.data()
-    return { taskData: await fromFirestoreData(taskData) }
+    const taskData = await fromFirestoreData(task.data())
+    if (taskData?.status) {
+      taskData.status = TASK_STATUS[taskData.status.toUpperCase()]
```
Pavangbhat, line 96: **Enhancement**: Let's have the default case as **IN_PROGRESS** everywhere because superusers can modify task status from the browser, while doing that if he/she makes a typo then task status will be undefined right?

Pavangbhat (changes requested):

constants/tasks.js
```diff
@@ -5,11 +5,27 @@ const TASK_TYPE = {
...
 const TASK_STATUS = {
-  ACTIVE: 'active',
-  ASSIGNED: 'assigned',
-  BLOCKED: 'blocked',
-  COMPLETED: 'completed',
-  UNASSIGNED: 'unAssigned'
+  // old task workflow
+  ACTIVE: 'IN_PROGRESS',
+  'HIGH PRIORITY': 'HIGH PRIORITY',
+  PENDING: 'PENDING',
```
Pavangbhat, line 11: **Pending** should be in progress here, right?

Pavangbhat (changes requested):

models/tasks.js
akshay1502: done.

models/tasks.js
akshay1502: having `IN_PROGRESS` doesn't feel right here? Since it's a typo, then we should treat it as a typo and should have a different name for it, by which anyone can identify that status is wrong here.

controllers/tasks.js
swarajpure, line 166: can we use constants here?

controllers/tasks.js
```diff
@@ -53,15 +53,23 @@ const fetchTasks = async (req, res) => {
...
-    if (status && !Object.values(TASK_STATUS).includes(status)) {
+    if (status && !Object.keys(TASK_STATUS).includes(status.toUpperCase())) {
       return res.boom.notFound('Status not found!')
     }
 
-    allTasks = await tasks.fetchUserTasks(username, status ? [status] : [])
+    if (status) {
+      if (status === 'active') {
+        status = ['active', 'blocked', 'pending', 'IN_PROGRESS', 'BLOCKED', 'SMOKE_TESTING']
+      } else {
```
swarajpure, line 67: can we use constants here?

models/tasks.js
```diff
@@ -61,7 +65,8 @@ const fetchTasks = async () => {
 
 const fetchActiveTaskMembers = async () => {
   try {
-    const tasksSnapshot = await tasksModel.where('type', '==', TASK_TYPE.FEATURE).where('status', '==', TASK_STATUS.ACTIVE).get()
+    const status = ['active', 'blocked', 'pending', 'IN_PROGRESS', 'BLOCKED', 'SMOKE_TESTING']
+    const tasksSnapshot = await tasksModel.where('type', '==', TASK_TYPE.FEATURE).where('status', 'in', status).get()
```
swarajpure, line 69: can we use constants here?

models/tasks.js
```diff
@@ -156,15 +166,22 @@ const fetchUserTasks = async (username, statuses = []) => {
...
     throw err
   }
 }
 
 const fetchUserActiveAndBlockedTasks = async (username) => {
-  return await fetchUserTasks(username, ['active', 'pending', 'blocked'])
+  return await fetchUserTasks(username,
+    ['active', 'pending', 'blocked', // old task workflow
+      'IN_PROGRESS', 'BLOCKED', 'SMOKE_TESTING'] // new task workflow
+  )
```
swarajpure, line 184: can we use constants here?

controllers/tasks.js
```diff
@@ -53,15 +53,23 @@ const fetchTasks = async (req, res) => {
...
 
-    if (status && !Object.values(TASK_STATUS).includes(status)) {
+    if (status && !Object.keys(TASK_STATUS).includes(status.toUpperCase())) {
       return res.boom.notFound('Status not found!')
     }
 
-    allTasks = await tasks.fetchUserTasks(username, status ? [status] : [])
+    if (status) {
+      if (status === 'active') {
+        status = ['active', 'blocked', 'pending', 'IN_PROGRESS', 'BLOCKED', 'SMOKE_TESTING']
```
swarajpure, line 66: Can we have this array made into a variable and then calling this variable here, that will help us understand what this array is about

models/tasks.js
```diff
@@ -174,9 +191,32 @@ const fetchUserActiveAndBlockedTasks = async (username) => {
  */
 
 const fetchUserCompletedTasks = async (username) => {
-  return await fetchUserTasks(username, ['completed'])
+  return await fetchUserTasks(username, ['completed', 'COMPLETED'])
```
swarajpure, line 194: can we use constants here?

controllers/tasks.js
Pavangbhat: Hi @swarajpure 
If we make them constants there is going to be collisions of the names because of this 👇 
[constants/tasks.js](https://github.com/Real-Dev-Squad/website-backend/blob/f2aa4b36123c09d465372e89f8ce8f167eed1c61/constants/tasks.js)

We can have a separate list of constants starting with something like **OLD_ACTIVE** but it might create confusions 
So what do you think we should do here ( Also one point in time old status will longer be in use and we might need to remove all of it)

Pavangbhat (commented):
Hi @swarajpure 
Let me know what do you think about this 👇 
https://github.com/Real-Dev-Squad/website-backend/pull/514/files#r799712960

Pavangbhat (approved):

Pavangbhat (dismissed):
Everything looks good to me 🎉

test/integration/tasks.test.js
```diff
@@ -137,7 +138,9 @@ describe("Tasks", function () {
 
   describe("GET /tasks/self", function () {
     it("Should get all the active and blocked tasks of the user", function (done) {
```
ankurnarkhede, line 140: Need test cases for:
- tasks with different status
- Overdue tasks

ankurnarkhede (changes requested):
Changes seems fine.
Need integration tests and unit tests as mentioned in comment.

constants/tasks.js
```diff
@@ -5,11 +5,34 @@ const TASK_TYPE = {
...
+  "HIGH PRIORITY": "HIGH PRIORITY",
+  PENDING: "IN_PROGRESS",
+  ASSIGNED: "ASSIGNED",
+  BLOCKED: "BLOCKED",
+  COMPLETED: "COMPLETED",
+
+  // new task workflow
+  IN_PROGRESS: "IN_PROGRESS",
+  UNASSIGNED: "AVAILABLE",
+  AVAILABLE: "AVAILABLE",
```
swarajpure, line 19: Two keys with same value?

akshay1502: Yes, `unassigned` is from old task values, and `available` is from new status values.

controllers/tasks.js
```diff
@@ -53,15 +55,22 @@ const fetchTasks = async (req, res) => {
...
       return res.boom.notFound("Status not found!");
     }
 
-    allTasks = await tasks.fetchUserTasks(username, status ? [status] : []);
+    if (status) {
+      if (status === OLD_ACTIVE) {
+        status = [OLD_ACTIVE, OLD_BLOCKED, OLD_PENDING, IN_PROGRESS, BLOCKED, SMOKE_TESTING];
+      } else {
+        status = [status];
+      }
```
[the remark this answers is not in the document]
akshay1502: it will give 404 as `status not found`(line63 no.)

minalchaudhari14:
Please assigned this task to me.

models/tasks.js
```diff
@@ -154,15 +165,29 @@ const fetchUserTasks = async (username, statuses = []) => {
 
     const promises = tasks.map(async (task) => fromFirestoreData(task));
     const updatedTasks = await Promise.all(promises);
-    return updatedTasks;
+    const taskList = updatedTasks.map((task) => {
+      task.status = TASK_STATUS[task.status.toUpperCase()];
```
palakgupta2712, line 169: what if that task.status is not present in the `TASK_STATUS` ?

constants/tasks.js
```diff
@@ -5,11 +5,34 @@ const TASK_TYPE = {
...
 
 const TASK_STATUS = {
-  ACTIVE: "active",
-  ASSIGNED: "assigned",
-  BLOCKED: "blocked",
-  COMPLETED: "completed",
-  UNASSIGNED: "unAssigned",
+  // old task workflow
+  ACTIVE: "IN_PROGRESS",
+  "HIGH PRIORITY": "HIGH PRIORITY",
```
palakgupta2712, line 10: You can remove quotes from the key.
https://github.com/airbnb/javascript#objects--quoted-props

test/integration/tasks.test.js
```diff
@@ -370,8 +377,34 @@ describe("Tasks", function () {
...
+        const { status, startedOn, endsOn, assignee } = task.unassignedTask;
+        expect(status).to.equal("AVAILABLE");
+        expect(assignee).to.equal(null);
+        expect(startedOn).to.equal(null);
+        expect(endsOn).to.equal(null);
+      });
+    });
+
+    it("Should return [] if no overdue task", async function () {
+      await tasks.updateTask(tasksData[2]);
```
palakgupta2712, line 402: What in the case if the logged in user is not a superUser? `${cookieName}=${jwt}` ?

test/integration/tasks.test.js
```diff
@@ -370,8 +377,34 @@ describe("Tasks", function () {
...
+    it("Should return all the overdue Tasks", async function () {
+      await tasks.updateTask(tasksData[0]);
+      await tasks.updateTask(tasksData[1]);
+      const res = await chai.request(app).get("/tasks/overdue").set("cookie", `${cookieName}=${superUserJwt}`);
+
+      expect(res).to.have.status(200);
+      expect(res.body.newAvailableTasks).to.be.a("array");
+      res.body.newAvailableTasks.forEach((task) => {
+        const { status, startedOn, endsOn, assignee } = task.unassignedTask;
+        expect(status).to.equal("AVAILABLE");
```
palakgupta2712, line 394: For readability and future updates/changes, try to use the same convention throughout the code.
Either use string or constant as already doing on line no 179.

test/integration/tasks.test.js
```diff
@@ -169,7 +176,7 @@ describe("Tasks", function () {
           }
           expect(res).to.have.status(200);
           expect(res.body).to.be.a("array");
-          expect(res.body[0].status).to.equal("completed");
+          expect(res.body[0].status).to.equal(TASK_STATUS.COMPLETED);
```
palakgupta2712, line 179: Why aren't we destructuring `TASK_STATUS.COMPLETED` beforehand as already doing for other task status?

models/tasks.js
akshay1502: We are showing the value which is already present as status(will make the change in next commit). Like here,https://github.com/Real-Dev-Squad/website-backend/blob/ccc712575ba3dfb7e35a0f9fb92659d2fcbccce1/models/tasks.js#L119

constants/tasks.js
palakgupta2712: Okay. Got it.

test/integration/tasks.test.js
akshay1502: 👍 will destructure it.

test/integration/tasks.test.js
akshay1502: okay.

Pavangbhat (approved):

Pavangbhat:
All changes have been done

Issue #520, opened 2022-02-05
PATCH /users/:userId endpoint


lakshayman:
Sets the user's latest profileDiff approval key to `APPROVED`.

Merge the new data of the user from profileDiff collection into users collection

and log the changes in logs collection

Request you are getting
{
  'first_name': string,
  'last_name': string,
  'yoe': number,
  'company': string,
  'designation': string,
  'github_id': string,
  'linkedin_id': string,
  'twitter_id': string,
  'instagram_id': string,
  'website': string,
 }

For more info refer here: https://github.com/Real-Dev-Squad/identity-service/issues/18

vikasosmium:
API Contract is not updated for this. @pallabez

PR #526, opened 2022-02-11
add PATCH /users/:userId endpoint and logging endpoint

pallabez:
Closes
- #520

### PATCH /users/:username endpoint
- Merge the profileDiffs to user document if this API is called. Only usable by super users.
- Set profileDiffs approval field from "PENDING" to "APPROVED"
- Logs the profile data change in logs of type "PROFILE_DIFF_APPROVED"

#### Structure of the log - 
```
{
  type: "PROFILE_DIFF_APPROVED",
  timestamp: <timestamp>,
  meta: {
    approvedBy: <super_userId>,
    userId: <userId>,
  }
  body: {
    profileDiffId: <profileDiffId>,
    message: ""
  }
}
````

### GET /logs/:type endpoint
- Fetch logs with the given query fields

#### Example 
- `localhost:3000/logs/ProfileChange?meta.username=pallab&meta.approvedBy=ankush`
- `localhost:3000/logs/ProfileChange?meta.username=lakshay`

### CODE REVIEW

File: models/logs.js [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Adds log
+ *
+ * @param type { String }: Type of the log
+ * @param data { String }: Body of the log
+ */
+const add = async (type, data) => {
+  try {
+    const log = {
+      type,
```
ankurnarkhede, line 14: How are we going to query the logs specific to any entity whenever required? These logs seems like those cant be queried.

ankurnarkhede (changes requested):

routes/users.js
```diff
@@ -267,4 +268,6 @@ router.post("/picture", authenticate, upload.single("profile"), users.postUserPi
 
 router.patch("/identityURL", authenticate, userValidator.updateIdentityURL, users.identityURL);
 
+router.patch("/:username", authenticate, authorizeUser("superUser"), users.updateUser);
```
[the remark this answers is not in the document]
pallabez: It's to make sure that their profile data is moderated. Superuser get's list of all profile changes(limited) and depending on the data it can be rejected or approved.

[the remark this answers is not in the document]
pallabez: This api is called after the profile changes is approved.

models/profileDiffs.js
```diff
@@ -0,0 +1,74 @@
+const firestore = require("../utils/firestore");
+const profileDiffsModel = firestore.collection("profileDiffs");
+
+/**
+ * Fetches the data about our users
+ * @param username { String }: Username of the user to fetch data of
+ * @return {Promise<profileDiffsModel>}
+ */
+const fetchProfileDiffsData = async (username) => {
```
pallabez, line 9: Not using this in this endpoint but is used in other endpoint

pallabez:
Changes requested by Ankush: 
- Change username to userId (Username might be changeable in future)
- Remove restriction of query-able fields being inside `meta`
- Remove oldProfileData and newProfileData from logs and only store profileDiffId

controllers/users.js
```diff
@@ -146,6 +149,40 @@ const postUserPicture = async (req, res) => {
...
+/**
+ * Updates the user data
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const updateUser = async (req, res) => {
+  try {
+    const userId = req.params.userId;
+    const { id: profileDiffId, message, ...profileDiff } = req.body;
```
harshith-venkatesh, line 161: Why do we need to update id here?
And could we look into more variable names instead of profileDiffId  ---> updatedUserId / profileDetails instead of profileDiff?

controllers/users.js
```diff
@@ -146,6 +149,40 @@ const postUserPicture = async (req, res) => {
...
+
+    const user = await userQuery.fetchUser({ userId });
+    if (!user.userExists) return res.boom.notFound("User doesn't exist");
+
+    const profileResponse = await profileDiffsQuery.update({ approval: "APPROVED" }, profileDiffId);
```
harshith-venkatesh, line 166: Can we look into creating constants for strings?
Like `"Approved"`

models/profileDiffs.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+/**
+ * Update profileDiff
+ *
+ * @param profileDiffData { Object }: Data to be added
+ * @param profileId { String }: Id of the profileDiff
+ */
+
+const update = async (profileDiffData, profileId) => {
```
harshith-venkatesh, line 28: `update` is a very common word... Could we be more specific here please

harshith-venkatesh (changes requested):

controllers/users.js
pallabez: The API is part of Identity Service. The variable names are taken with account of those only. 

Profile Diff are profile changes that user requested. These changes are stored in `profileDiff` model and only updated in user document when a `super_user` approves it. A user themselves can't update their data via this API.

We are passing userId as params because `req.userData` contains the data of super_user who called it.

controllers/users.js
pallabez: Ok. Will make the changes.

models/profileDiffs.js
pallabez: Is `updateProfileDiff` ok?

harshith-venkatesh (approved):
LGTM

vikasosmium:
API Contract is not updated for this. @pallabez

Issue #527, opened 2022-02-11
Update {API}/tasks/self to return all the tasks assigned to a member


Pavangbhat:
Currently, **{API}/tasks/self** endpoint returns a few sets of tasks based on some set of statutes.
Tasks:-
- [x] Endpoint should return all the tasks assigned to a member sorted by **startedOn** field the latest one should be in the top
- [x] Check if **/tasks/self** endpoint's handler function being used some where else

Aryan-9211:
I want to work on this issue.

PR #532, opened 2022-02-15
return tasks according to startedOn field

Aryan-9211:
PR for issue #527

### DIFF
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -126,7 +126,7 @@ const fetchSelfTask = async (taskId, userId) => {
  * @return {Promise<tasks|Array>}
  */
 
-const fetchUserTasks = async (username, statuses = []) => {
+const fetchUserTasks = async (username, statuses = [], field, order) => {
   try {
     const userId = await userUtils.getUserId(username);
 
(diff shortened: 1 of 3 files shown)


Pavangbhat (changes requested):
Some checks are failing please have a look

models/tasks.js
```diff
@@ -161,8 +161,9 @@ const fetchUserTasks = async (username, statuses = []) => {
   }
 };
 
-const fetchUserActiveAndBlockedTasks = async (username) => {
-  return await fetchUserTasks(username, ["active", "pending", "blocked"]);
+const fetchAllSortedUserTasks = async (username) => {
+  const selftasks = await fetchUserTasks(username);
+  return await selftasks.orderBy("endsOn");
```
akshay1502, line 166: Do we have `orderBy` in javascript?🤔

models/tasks.js
```diff
@@ -161,8 +161,14 @@ const fetchUserTasks = async (username, statuses = []) => {
...
-const fetchUserActiveAndBlockedTasks = async (username) => {
-  return await fetchUserTasks(username, ["active", "pending", "blocked"]);
+/**
+ *
+ * @param username { string } : username which will be used to fetch all self tasks
+ * @returns {Promise<tasks>|Array}
+ */
+const fetchSelfTasks = async (username) => {
+  const allTasks = await fetchUserTasks(username);
+  return allTasks.sort((a, b) => a.endsOn - b.endsOn);
```
Pavangbhat, line 171: We should show the nearest deadline first

Pavangbhat (changes requested):
Nice Work
There is one change needed

models/tasks.js
shubham-y, line 171: Any reasons to not use Firestore `orderBy()` function for sorting?

Pavangbhat: Lets go with **orderBy()**

akshay1502: @shubham-y @Pavangbhat `orderBy()` is a query provided by firestore, since we are using already written function `fetchUserTasks` here, To use orderBy we need to use it in the fetchuserTasks function which I don't think is a good idea since we are using the function at other places too. So I am continuing with sort method.

Pavangbhat: Can we have a second parameter here and based on it we can use **orderBy**

akshay1502: done.

models/tasks.js
```diff
@@ -126,7 +126,7 @@ const fetchSelfTask = async (taskId, userId) => {
  * @return {Promise<tasks|Array>}
  */
 
-const fetchUserTasks = async (username, statuses = []) => {
+const fetchUserTasks = async (username, statuses = [], field, order) => {
```
Pavangbhat, line 129: **Suggestion:** we can place statues as the last argument so that it can be skipped at line 190

Pavangbhat (approved):
I have left one suggestion please have a look.
Everything looks good to me. 
Great work @akshay1502 🎉

models/tasks.js
akshay1502: majority of the time we would be using the status parameter and field, order less frequently. So let's have status on the 2nd.

Issue #501, opened 2022-01-16
Check current coverage and Files to ignore


ankushdharkar:
# Unit Testing
![image](https://user-images.githubusercontent.com/63191499/154795077-b312ea1e-45ae-4597-b6a4-151388329d7a.png)

## Files and Folders to ignore

- [ ] website-backend 
![image](https://user-images.githubusercontent.com/63191499/154795228-98bebb4e-ed48-464e-aefd-546e65378776.png)
- [ ] website-backend/config
![image](https://user-images.githubusercontent.com/63191499/154795257-35826bf0-f10b-46c4-baa2-fa5d8950d169.png)
- [ ] website-backend/constants
![image](https://user-images.githubusercontent.com/63191499/154795326-9d9cba5b-4ef0-443c-88e9-91e7d0b3c38d.png)
- [X] website-backend/controllers
![image](https://user-images.githubusercontent.com/63191499/154795363-9acf512f-ecb2-4020-9c5a-38c5e24b873c.png)
- [X] website-backend/docs
![image](https://user-images.githubusercontent.com/63191499/154795485-340ff6bd-59c4-40ae-9513-f5fa27bcf1aa.png)
- [ ] website-backend/middlewares
![image](https://user-images.githubusercontent.com/63191499/154795507-b855c9e4-52c6-4458-a1bf-6484c16e1482.png)
- [ ] website-backend/middlewares/validators
![image](https://user-images.githubusercontent.com/63191499/154795517-331754fc-b9d3-4489-a797-73be9c49351f.png)
- [X] website-backend/mockdata
![image](https://user-images.githubusercontent.com/63191499/154795560-01debaf9-aa5e-439b-bedd-a2ac54b23b7d.png)
- [ ] website-backend/models
![image](https://user-images.githubusercontent.com/63191499/154795663-856aab81-6b1d-4290-80fb-86196f51a04a.png)
- [ ] website-backend/routes
![image](https://user-images.githubusercontent.com/63191499/154795672-426486c8-5fee-4b7e-84fd-06d172088119.png)
- [ ] website-backend/scripts
![image](https://user-images.githubusercontent.com/63191499/154795689-f23313fa-43ff-4309-a4ed-cc1beee87304.png)
- [ ] website-backend/services
![image](https://user-images.githubusercontent.com/63191499/154795710-432a798b-4f1b-4996-ad39-1a9244332328.png)
- [ ] website-backend/utils
[...]

PR #534, opened 2022-02-23
ignore unnecessary files in test coverage report

pallabez:
Closes 
- https://github.com/Real-Dev-Squad/website-backend/issues/501

# After

### Integration Test Report
image
image
image

### Unit Test Report

image
image

### DIFF
--- a/scripts/tests/testIntegration.sh
+++ b/scripts/tests/testIntegration.sh
@@ -4,4 +4,4 @@
 export NODE_ENV='test'
 
 echo 'Start firestore emulator and run integration tests:'
-firebase emulators:exec 'nyc --n=controllers/** mocha test/integration/**'
+firebase emulators:exec 'nyc mocha test/integration/**'
--- a/scripts/tests/testIntegration.sh
+++ b/scripts/tests/testIntegration.sh
@@ -4,4 +4,4 @@
 export NODE_ENV='test'
 
 echo 'Start firestore emulator and run integration tests:'
-firebase emulators:exec 'nyc mocha test/integration/**'
+firebase emulators:exec 'nyc --n=controllers/** mocha test/integration/**'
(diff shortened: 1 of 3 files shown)


scripts/tests/testUnit.sh
```diff
@@ -4,4 +4,4 @@
 export NODE_ENV='test'
 
 echo 'Start firestore emulator and run unit tests:'
-firebase emulators:exec 'nyc mocha test/unit/**'
+firebase emulators:exec 'nyc --n=models --n=middlewares/**.js --n=services --n=utils mocha test/unit/**'
```
pallabez, line 7: Tried using array here to include multiple folders but couldn't get syntax right.

scripts/tests/testIntegration.sh
```diff
@@ -4,4 +4,4 @@
 export NODE_ENV='test'
 
 echo 'Start firestore emulator and run integration tests:'
-firebase emulators:exec 'nyc mocha test/integration/**'
+firebase emulators:exec 'nyc --n=controllers/** mocha test/integration/**'
```
ankurnarkhede, line 7: The tests internally also cover all other folders. Is there any flag to ignore few folders, rather than include?

ankurnarkhede (changes requested):

scripts/tests/testIntegration.sh
pallabez: There is --x or --exclude to ignore but there is a lot of folders to ignore.

ankur-lt: Updated here: https://github.com/Real-Dev-Squad/website-backend/issues/501#issuecomment-1048772568

pallabez:
Merged in 
- https://github.com/Real-Dev-Squad/website-backend/pull/543

Issue #539, opened 2022-02-24
Improve test coverage to 50%


pallabez:
# Files that needs more test coverage
Priority in decreasing order

# Unit Test

## website-backend/models
- [x] badges.js 
- [x] stocks.js
- [x] members.js
- [x] challenges.js
- [x] auctions.js
- [x] tasks.js 
- [x] featureFlags.js  
- [x] wallets.js

![image](https://user-images.githubusercontent.com/63191499/155532055-ca340634-e008-484f-bbf3-463e49d4900d.png)

## website-backend/utils
![image](https://user-images.githubusercontent.com/63191499/155531845-2c72549d-49d7-496d-9c28-4218e6c7a81f.png)

## website-backend/services
![image](https://user-images.githubusercontent.com/63191499/155531893-23c96e81-f503-486b-9a54-a58a79dc8db8.png)

# Integration Test

Already Covered

pallabez:
Working on website-backend/models/auctions.js

pallabez:
Working on `website-backend/models/wallets.js`

siddhesh494:
Working on `website-backend/models/members.js`

siddhesh494:
Working on `website-backend/models/badges.js`

PR #543, opened 2022-03-03
add unit tests for models/auction.js

pallabez:
- Merge after #534

- Fixes https://github.com/Real-Dev-Squad/website-backend/issues/539
## New coverage 
image

### DIFF
--- a/test/fixtures/currencies/currencies.js
+++ b/test/fixtures/currencies/currencies.js
@@ -7,18 +7,16 @@ const { DINERO, NEELAM } = require("../../../constants/wallets");
  * @return {Object}
  */
 
-module.exports = () => {
-  return [
-    {
-      [DINERO]: 1000,
-      [NEELAM]: 2,
-    },
-    {
-      [DINERO]: 2000,
-      [NEELAM]: 0,
-    },
-    {
-      [DINERO]: 1000,
-    },
-  ];
-};
+module.exports = [
+  {
+    [DINERO]: 1000,
+    [NEELAM]: 2,
+  },
+  {
+    [DINERO]: 2000,
+    [NEELAM]: 0,
+  },
+  {
+    [DINERO]: 1000,
+  },
+];
--- a/test/fixtures/currencies/currencies.js
+++ b/test/fixtures/currencies/currencies.js
@@ -7,9 +7,18 @@ const { DINERO, NEELAM } = require("../../../constants/wallets");
  * @return {Object}
  */
 
-const currencies = {
-  [DINERO]: 1000,
-  [NEELAM]: 2,
+module.exports = () => {
+  return [
+    {
+      [DINERO]: 1000,
+      [NEELAM]: 2,
+    },
+    {
+      [DINERO]: 2000,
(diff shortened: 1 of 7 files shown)


test/unit/models/auctions.test.js
```diff
@@ -0,0 +1,113 @@
...
+    seller = await addUser();
+    bidder = await addUser(user[1]);
+    auctionId = await auctions.createNewAuction({ seller, initialPrice, itemType, endTime, quantity });
+  });
+  afterEach(async function () {
+    await cleanDb();
+    sinon.restore();
+  });
+
+  describe("createNewAuction", function () {
```
swarajpure, line 35: Can we go ahead with simple English words with spaces here? Rather than camelCase

test/unit/models/auctions.test.js
```diff
@@ -0,0 +1,113 @@
...
+    it("Should create a new auction", async function () {
```
swarajpure, line 36: Should we also have a testcase where you're not able to successfully create an auction? Where you get 4xx as response?

test/unit/models/auctions.test.js
```diff
@@ -0,0 +1,113 @@
...
+      expect(response.end_time).to.be.equal(endTime);
+      expect(response.quantity).to.be.equal(quantity);
+      expect(response.seller).to.be.equal(user[0].username);
+      expect(response.bidders_and_bids).to.be.a("array").with.lengthOf(0);
+      expect(response.highest_bidder).to.be.equal(undefined);
+    });
+    it("Should return not found", async function () {
+      const response = await auctions.fetchAuctionById("undefined");
+
+      expect(response).to.be.equal(false);
```
swarajpure, line 65: Instead of passing `undefined`, can we pass some other string like `test123`? Someone might think of this an `undefined` keyword of JS and not any random string

test/unit/models/auctions.test.js
pallabez: Shouldn't it be the function name as done in other unit tests?

test/unit/models/auctions.test.js
pallabez: Got it :+1: . The intentional was to bring the test coverage above 50% so didn't wrote all the cases.

test/unit/models/auctions.test.js
pallabez: Ok :+1:

test/fixtures/currencies/currencies.js
```diff
@@ -7,9 +7,18 @@ const { DINERO, NEELAM } = require("../../../constants/wallets");
...
+      [NEELAM]: 2,
+    },
+    {
+      [DINERO]: 2000,
+      [NEELAM]: 0,
+    },
+    {
+      [DINERO]: 1000,
+    },
+  ];
```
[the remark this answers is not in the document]
pallabez: Not really. But most fixtures follow this style so I'm doing it as well.

swarajpure (approved):
Great job in here! 🔥

PR #548, opened 2022-03-07
adds unit test for models/members.js

siddhesh494:
#539 
Adds unit tests for models/members.js file
### New Coverage
image


test/unit/models/members.test.js
```diff
@@ -0,0 +1,119 @@
...
+    await cleanDb();
+  });
+
+  describe("fetchUsers", function () {
+    it("should return all the users", async function () {
+      const result = await members.fetchUsers();
+      const user = result[0];
+
+      expect(result).to.be.a("array");
+      expect(user.first_name).to.be.a("string");
```
pallabez, line 25: Looks good but as we are comparing fields directly, so we don't need to compare the field type anymore.

Also, the result looks something like this: 
image

We can use `Object.key` with `map` function to compare all fields easily as done in `models/users.js`

pallabez:
LGTM :+1:

prakashchoudhary07:
Closing this PR, as `/members` route is being deprecated

PR #549, opened 2022-03-09
add unit test for models/wallets.js

pallabez:
Help Fixes
- #539

## New coverage

image

### DIFF
--- a/test/fixtures/currencies/currencies.js
+++ b/test/fixtures/currencies/currencies.js
@@ -7,18 +7,16 @@ const { DINERO, NEELAM } = require("../../../constants/wallets");
  * @return {Object}
  */
 
-module.exports = () => {
-  return [
-    {
-      [DINERO]: 1000,
-      [NEELAM]: 2,
-    },
-    {
-      [DINERO]: 2000,
-      [NEELAM]: 0,
-    },
-    {
-      [DINERO]: 1000,
-    },
-  ];
-};
+module.exports = [
+  {
+    [DINERO]: 1000,
+    [NEELAM]: 2,
+  },
+  {
+    [DINERO]: 2000,
+    [NEELAM]: 0,
+  },
+  {
+    [DINERO]: 1000,
+  },
+];
--- a/test/fixtures/currencies/currencies.js
+++ b/test/fixtures/currencies/currencies.js
@@ -7,9 +7,18 @@ const { DINERO, NEELAM } = require("../../../constants/wallets");
  * @return {Object}
  */
 
-const currencies = {
-  [DINERO]: 1000,
-  [NEELAM]: 2,
+module.exports = () => {
+  return [
+    {
+      [DINERO]: 1000,
+      [NEELAM]: 2,
+    },
+    {
+      [DINERO]: 2000,
+      [NEELAM]: 0,
+    },
+    {
+      [DINERO]: 1000,
+    },
+  ];
 };
-
-module.exports = currencies;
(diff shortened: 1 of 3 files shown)


siddhesh494:
LGTM 👍

test/fixtures/currencies/currencies.js
```diff
@@ -7,9 +7,18 @@ const { DINERO, NEELAM } = require("../../../constants/wallets");
  * @return {Object}
  */
 
-const currencies = {
-  [DINERO]: 1000,
-  [NEELAM]: 2,
+module.exports = () => {
+  return [
+    {
+      [DINERO]: 1000,
```
MehulKChaudhari, line 13: I did not understand this syntax why it is `[DINERO]` and not `"DINERO"`?

MehulKChaudhari: Also, why did you export a function from here for any specific reason? Like you can just export an object it will be easy to consume as well. you need not to call it when using,

test/integration/auction.test.js
```diff
@@ -14,7 +14,7 @@ const { createWallet } = require("../../models/wallets");
 const userData = require("../fixtures/user/user")();
 const { auctionData, auctionKeys, auctionWithIdKeys } = require("../fixtures/auctions/auctions");
 const { initial_price: initialPrice, item_type: itemType, end_time: endTime, quantity } = auctionData;
-const currenciesData = require("../fixtures/currencies/currencies");
+const currenciesData = require("../fixtures/currencies/currencies")()[0];
```
MehulKChaudhari, line 17: This is a little complex to understand. You can define the `[0]` data to another variable on another line?

something like `const currencyData = currenciesData()[0]`

Also I think this could have been done without making `currenciesData` a function

MehulKChaudhari (commented):
Tests look good :fire: Thanks Pallab.

test/fixtures/currencies/currencies.js
pallabez: image

DINERO is a variable, that's why it's like that. 

For second question. Yes, I was planning on doing that but most fixtures follow that pattern so continuing with function export. Should I change that?

test/integration/auction.test.js
pallabez: 1) Will do that. :+1: 

2) Mentioned above

test/integration/auction.test.js
```diff
@@ -14,7 +14,7 @@ const { createWallet } = require("../../models/wallets");
 const userData = require("../fixtures/user/user")();
 const { auctionData, auctionKeys, auctionWithIdKeys } = require("../fixtures/auctions/auctions");
 const { initial_price: initialPrice, item_type: itemType, end_time: endTime, quantity } = auctionData;
-const currenciesData = require("../fixtures/currencies/currencies");
+const currenciesData = require("../fixtures/currencies/currencies")[0];
```
MehulKChaudhari, line 17: You can destructure the currenciesData on the next line it will be more readable. 

like this: 
`const currencyData = currenciesData[0];` or `[const currencyData] = currenciesData`

pallabez: Made the changes, is that ok?

pallabez:
Currencies fixture will be handled in the ticket after this merge
- https://github.com/Real-Dev-Squad/website-backend/issues/585

(suggested by Ankush)

MehulKChaudhari (approved):

PR #550, opened 2022-03-10
adds unit tests for models/badges.js

siddhesh494:
- #539  

### New Coverage

image


test/unit/models/members.test.js
```diff
@@ -0,0 +1,123 @@
+const chai = require("chai");
```
pallabez, line 1: Any reason why we are pushing members test as well?

siddhesh494: ohh sorry for that I by mistakely created branch from tests/members so that why it is showing the changes from that branch also, is there something i can do to solve this problem?

test/unit/models/badges.test.js
```diff
@@ -0,0 +1,53 @@
...
+      expect(response.userExists).to.be.equal(true);
+      expect(response.userBadges).of.length.greaterThan(0);
+    });
+
+    it("should return empty badges if username does not exist", async function () {
+      const response = await badges.fetchUserBadges("undefined");
+
+      expect(response).to.be.a("object");
+      expect(response.userExists).to.be.equal(false);
+      expect(response.userBadges).of.length(0);
```
pallabez, line 50: Can we compare the userBadges as well?
image

pallabez, line 50: Can we compare the userBadges as well?
image

test/unit/models/badges.test.js
```diff
@@ -0,0 +1,53 @@
...
+  let userID;
+  beforeEach(async function () {
+    userID = await addUser();
+    await addBadges({ userID });
+  });
+  afterEach(async function () {
+    await cleanDb();
+  });
+
+  describe("getbadges", function () {
```
pallabez, line 21: We should be writing the function exact name here. `getbadges` -> `fetchBadges`

pallabez, line 21: We should be writing the function exact name here. `getbadges` -> `fetchBadges`

prakashchoudhary07:
Closing this, as its out of sync and new tests will be written later

PR #551, opened 2022-03-20
add unit tests for models/challenges and add time utils 

pallabez:
Helps Fixes 
- https://github.com/Real-Dev-Squad/website-backend/issues/539

## Todo 
- Fix data model of challenges

## New test coverage
image

### DIFF
--- a/models/challenges.js
+++ b/models/challenges.js
@@ -6,7 +6,6 @@
 const Firestore = require("@google-cloud/firestore");
 const firestore = require("../utils/firestore");
 const { fetchUser } = require("./users");
-const { getTimeInSecondAfter } = require("../utils/time");
 
 const challengesModel = firestore.collection("challenges");
 const userModel = firestore.collection("users");
@@ -67,13 +66,13 @@ const fetchParticipantsData = async (participants) => {
 
 const postChallenge = async (challengeData) => {
   try {
-    const { start_at: startAt, end_at: endAt } = challengeData;
-    const startat = new Firestore.Timestamp(getTimeInSecondAfter({ timestamp: startAt }), 0);
-    const endat = new Firestore.Timestamp(getTimeInSecondAfter({ timestamp: endAt }), 0);
+    const { start_date: startDate, end_date: endDate } = challengeData;
+    const startdate = new Firestore.Timestamp(startDate, 0);
+    const enddate = new Firestore.Timestamp(endDate, 0);
     const challengeRef = await challengesModel.add({
       ...challengeData,
-      start_at: startat,
-      end_at: endat,
+      start_date: startdate,
+      end_date: enddate,
       participants: [],
       is_active: true,
     });
--- a/models/challenges.js
+++ b/models/challenges.js
@@ -6,6 +6,7 @@
 const Firestore = require("@google-cloud/firestore");
 const firestore = require("../utils/firestore");
 const { fetchUser } = require("./users");
+const { getTimeInSecondAfter } = require("../utils/time");
 
 const challengesModel = firestore.collection("challenges");
 const userModel = firestore.collection("users");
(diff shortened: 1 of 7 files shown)


test/fixtures/challenges/challenges.js
```diff
@@ -0,0 +1,10 @@
+module.exports = () => {
+  return [
+    {
+      title: "Sherlock and Anagrams",
+      level: "Easy",
+      start_date: parseInt(Date.now() / 1000),
+      end_date: parseInt(Date.now() / 1000) + 60 * 60 * 24 * 10, // After 10 days
```
shubham-y, line 7: **Suggestion**: Add a constant variable to store seconds
Reference - https://github.com/ryanmcdermott/clean-code-javascript#use-searchable-names

test/unit/models/challenges.test.js
```diff
@@ -0,0 +1,86 @@
+/* eslint-disable security/detect-object-injection */
```
shubham-y, line 1: Please check if we can remove this

test/unit/models/challenges.test.js
```diff
@@ -0,0 +1,86 @@
...
+const challengeDataArray = require("../../fixtures/challenges/challenges")();
+const challengeData = challengeDataArray[0];
+
+describe("Challenges", function () {
+  afterEach(async function () {
+    await cleanDb();
+  });
+  describe("postChallenge", function () {
+    it("should add the challenge", async function () {
+      const challengeId = await challengeQuery.postChallenge(challengeData);
```
shubham-y, line 23: Can we move it to `beforeEach` function as it's being used in multiple it blocks?

shubham-y (changes requested):
Good work
Requesting some changes

test/fixtures/challenges/challenges.js
```diff
@@ -0,0 +1,12 @@
...
+
+module.exports = () => {
+  return [
+    {
+      title: "Sherlock and Anagrams",
+      level: "Easy",
+      start_date: parseInt(Date.now() / 1000),
+      end_date: parseInt(Date.now() / 1000) + convertToSeconds.DAYS_10, // After 10 days
+    },
+  ];
```
ankushdharkar, line 11: Instead of doing mathematical calculations here, I am wondering if we can import a util function for our Date needs?

```js
getTimeNow()
getTimeAfter({ days: 10})
```

Also, why are we passing timestamp in a "<>_date" key?

test/unit/models/challenges.test.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+    it("should return a empty array", async function () {
+      await cleanDb();
+      const response = await challengeQuery.fetchChallenges();
+
+      expect(response).to.be.a("array").with.lengthOf(0);
+    });
+  });
+
+  describe("subscribeUserToChallenge", function () {
```
shubham-y, line 60: We should add a test passing invalid userId to `subscribeUserToChallenge()` and `fetchParticipantsData()`. What do you think?

shubham-y: Also, pass a challenge fixture that already has a participant in order to test multiple cases

pallabez: Done! :+1:

test/fixtures/challenges/challenges.js
pallabez: Done :+1:

test/fixtures/time/time.js
```diff
@@ -0,0 +1,55 @@
+const minutesToMilliseconds = [
```
shubham-y, line 1: Suggestion to please use capital letters and underscores for naming constant variables
Reference - https://javascript.info/variables#uppercase-constants

test/unit/utils/time.test.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable security/detect-object-injection */
```
shubham-y, line 1: Please check if we can remove this

models/challenges.js
```diff
@@ -66,13 +67,13 @@ const fetchParticipantsData = async (participants) => {
...
+    const endat = new Firestore.Timestamp(getTimeInSecondAfter({ timestamp: endAt }), 0);
     const challengeRef = await challengesModel.add({
       ...challengeData,
-      start_date: startdate,
-      end_date: enddate,
+      start_at: startat,
+      end_at: endat,
       participants: [],
       is_active: true,
     });
```
shubham-y, line 79: I think it will be better to raise a different PR for changing the keys of challenges object because it might cause issue in frontend. What do you think?

pallabez: Yeah, you are right. Will do that.

test/fixtures/time/time.js
pallabez: No function in fixtures follows it. I believe it's not really a constant. Is it still suggested?

test/unit/utils/time.test.js
pallabez: Yes. Thanks for pointing that out.

PR #553, opened 2022-03-22
add unit tests for utils/multer

pallabez:
Help Fixes
- https://github.com/Real-Dev-Squad/website-backend/issues/539

## New test coverage 
image

### DIFF
--- a/test/unit/utils/multer.test.js
+++ b/test/unit/utils/multer.test.js
@@ -3,8 +3,12 @@ const { expect } = chai;
 const sinon = require("sinon");
 
 const cleanDb = require("../../utils/cleanDb");
-
 const multer = require("../../../utils/multer");
+
+const errorMessage = require("../../../constants/errorMessages");
+const multerConstant = require("../../../constants/multer");
+const MB_1 = multerConstant.FILE_SIZE_1MB;
+const profileFileSize = multerConstant.PROFILE_FILE_SIZE;
 
 describe("multer", function () {
   afterEach(async function () {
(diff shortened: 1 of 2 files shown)


test/unit/utils/multer.test.js
```diff
@@ -0,0 +1,64 @@
...
+    await cleanDb();
+    sinon.restore();
+  });
+  describe("fileFilterImagesOnly", function () {
+    const callBack = (err, success) => {
+      return { err, success };
+    };
+    const fileType = [{ mimetype: "image/png" }, { mimetype: "test" }];
+    it("should pass without any error", async function () {
+      const response = await multer.fileFilterImagesOnly({}, fileType[0], callBack);
```
pallabez, line 20: Instead of callback, we can here too use sinon spy.

test/unit/utils/multer.test.js
```diff
@@ -0,0 +1,64 @@
...
+      { code: "TYPE_UNSUPPORTED_FILE" },
+      { code: "TEST" },
+    ];
+    beforeEach(async function () {
+      spy = sinon.spy();
+    });
+
+    it("should call the entityTooLarge error callback", function () {
+      multer.multerErrorHandling(errCode[0], {}, { boom: { entityTooLarge: spy } }, {});
+      expect(spy.callCount).to.be.equal(1);
```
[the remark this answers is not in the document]
prakashchoudhary07: Can we check for something else also? Like the response

test/unit/utils/multer.test.js
prakashchoudhary07: You can, but I think this is fine, What improvements do we get if Sinon is used here?

prakashchoudhary07 (changes requested):

test/unit/utils/multer.test.js
pallabez: Ok. It will be more consistent, that's the only benefit.

prakashchoudhary07 (changes requested):
Nice work :tada: :tada:

prakashchoudhary07 (dismissed):

PR #554, opened 2022-03-23
Made /users/verify

lakshayman:


controllers/users.js
```diff
@@ -1,6 +1,34 @@
...
+
+const verifyUser = async (req, res) => {
+  try {
+    const userId = req.userData.id;
+    if (!req.userData?.identityURL) {
+      return res.boom.serverUnavailable("IdentityURL is Missing");
+    }
+    await userQuery.addOrUpdate({ identityStatus: "PENDING" }, userId);
+  } catch (error) {
+    logger.error(`Error while fetching all users: ${error}`);
```
shubham-y, line 14: The error message seems incorrect. Please check

test/integration/users.test.js
```diff
@@ -354,4 +354,36 @@ describe("Users", function () {
...
+        .end((err, res) => {
+          if (err) {
+            return done();
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Your request has been queued successfully");
+          return done();
+        });
+    });
```
shubham-y, line 372: Can we also add assertions to check if user `identityStatus` is updated to `PENDING`?

shubham-y (changes requested):
Good work
Requesting some changes

controllers/users.js
```diff
@@ -1,6 +1,34 @@
...
+  const fetchData = {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      username: req.userData.username,
+    }),
+  };
+  fetch(process.env.IDENTITY_SERVICE_URL, fetchData);
```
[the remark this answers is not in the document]
shubham-y: Also, is there a reason why the already existing [fetch](https://github.com/Real-Dev-Squad/website-backend/blob/develop/utils/fetch.js) util function is not being used instead of adding a new package?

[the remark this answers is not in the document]
lakshayman: Fetch call is outside the try-catch on purpose because we don't care about the response of that call.

Node has no existing fetch

test/integration/users.test.js
lakshayman: that will be automatically checked if there is no error

PR #556, opened 2022-03-25
add unit tests for models/featureFlag

pallabez:
## New Test Coverage

image

We are using fetching and storing by `username` and not by `userId`, which will cause problem in future if username changes.


pallabez:
This endpoint is getting deprecated. New endpoint :
- https://github.com/Real-Dev-Squad/website-backend/issues/563

PR #562, opened 2022-04-10
Integration Test for api improvement

### LINKED ISSUE #545: We are having API `/arts/user/self` for getting all the arts that we created or own. But what if we want to see other people's arts as well, there should be `/arts/user/:username` instead of `/arts/user/self`

RahulGoyal-tech:
closes #545


ankushdharkar (changes requested):
@RahulGoyal-tech This PR seems to have too many changes outside of just tests. Please send only the tests.

Issue #564, opened 2022-04-12
POST /flag/add endpoint


pallabez:
# POST /flag/add 

## Parent Issue
- #563 

## Steps

Branch - `feature-flag/add`
- [x] Create flag.js file for route, controller and models
- [x] Set `/flag/add` authentication to `super_user` and add `/flag` route to `routes/index.js`
- [x] Create controller and model (function name - `addFlag` for both)

Example of object to be added 
```
title: dark-mode
enabled: true/flase
roles: [members, no_role, super_user],
users: {
  "pallab id": true,
  "rohit id": false,
},
```

-----

Branch - `feature-flag/add-tests-validator`
- [x] Write integration and unit test 
- [x] Add `joi` validation middle ware (All the fields are required/not optional)

RahulGoyal-tech:
Assign me this issue please!!

shubham-y:
Reopening the issue as PR is in progress

pallabez:
Now Handled in the repo 
- https://github.com/Real-Dev-Squad/feature-flag-service

Issue #563, opened 2022-04-11
API endpoint for Feature Flag


pallabez:
# Feature Flag API 

## Feature Flag Object 
```
title: dark-mode
enabled: true
roles: [members, no_role, super_user],
users: {
   "pcBpdEQ5ydRakHHSpY2p": true,
   "quEpdE34ydRakJSpY45p": false,
},
```

## Requests

|               Route                  |          Description           |           Authentication          |
| :--------------------------------: | :-----------------------------: | :---------------------------------: |
|      POST /flag/add       |  Adds the flag         | super_user   |
|      PATCH /flag/:flagId |  Updates the flag    | super_user   |
|      GET /flag                |  Fetches all flags    | super_user   |
|      GET /flag/:flagId     |  Fetches a flag        | user              |

## POST /flag/add
Add the feature flag

- **Params**\
    None
- **Query**\
    None
- **Body**
```
title: dark-mode
enabled: true
roles: [members, no_role, super_user],
users: {
   "pcBpdEQ5ydRakHHSpY2p": true,
   "quEpdE34ydRakJSpY45p": false,
},

```
- **Headers**\
    None
- **Cookie**\
    Super User Cookies
- **Success Response:**
- **Code:** 200
    - **Content:** 
        ```
        {
            message: 'Added feature flag successfully!',
            flagId: <id>
        }
        ```

## PATCH /flag/:flagId 
Update the feature flag

- **Params**\
    None
- **Query**\
   flagId
- **Body**
```
enabled: true
roles: [members, no_role, super_user],
users: {
   "pcBpdEQ5ydRakHHSpY2p": true,
   "quEpdE34ydRakJSpY45p": false,
},
```
- **Headers**\
    None
- **Cookie**\
    Super User Cookies
- **Success Response:**
- **Code:** 200
    - **Content:** 
        ```
        {
            message: 'Updated feature flag successfully!',
            flagId: <id>
        }
        ```
## GET /flag   
Gets all the feature flag

- **Params**\
    None
- **Query**\
   None
- **Body**
   None
- **Headers**\
    None
- **Cookie**\
    Super User Cookies
- **Success Response:**
- **Code:** 200
    - **Content:** 
        ```
        {
            message: 'Fetched all feature flag successfully!',
            flags: [
                {
                    flag_object1
                },
                {
                    flag_object2
                }
            ]
        }
        ```

pallabez:
Now Handled in the repo 
[...]

PR #570, opened 2022-04-16
Feature flag/add

RahulGoyal-tech:
flag/add endpoint created with tests and validators.

Resolves
Base Ticket : #564 
Parent Issue : #563

### DIFF
--- a/controllers/flag.js
+++ b/controllers/flag.js
@@ -1,7 +1,7 @@
 const flagQuery = require("../models/flag");
 
 /**
- * Adds art
+ * Controller function to add feature flag data to firestore
  *
  * @param req {Object} - Express request object
  * @param res {Object} - Express response object
--- a/controllers/flag.js
+++ b/controllers/flag.js
@@ -10,7 +10,7 @@ const addFlag = async (req, res) => {
   try {
     const flagId = await flagQuery.addFlag(req.body);
     return res.json({
-      message: "Add feature flag successfully!",
+      message: "Added feature flag successfully!",
       flagId: flagId,
     });
   } catch (err) {
--- a/controllers/flag.js
+++ b/controllers/flag.js
@@ -1,0 +1,24 @@
+const flagQuery = require("../models/flag");
+
+/**
+ * Adds art
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const addFlag = async (req, res) => {
+  try {
+    const flagId = await flagQuery.addFlag(req.body);
+    return res.json({
+      message: "Add feature flag successfully!",
+      flagId: flagId,
+    });
+  } catch (err) {
+    logger.error(`Error while adding featureFlag info: ${err}`);
+    return res.boom.badImplementation("Something went wrong please contact admin");
+  }
+};
+
+module.exports = {
+  addFlag,
+};
--- a/middlewares/validators/flag.js
+++ b/middlewares/validators/flag.js
@@ -1,6 +1,6 @@
 const joi = require("joi");
 
-const addFeatureFlagValidator = async (req, res, next) => {
+const addFeatureFlag = async (req, res, next) => {
   const schema = joi
     .object()
     .strict()
@@ -15,11 +15,11 @@ const addFeatureFlagValidator = async (req, res, next) => {
     await schema.validateAsync(req.body);
     next();
   } catch (error) {
-    logger.error(`Error validating createTask payload : ${error}`);
+    logger.error(`Error in validating Feature Flag : ${error}`);
     res.boom.badRequest(error.details[0].message);
   }
 };
 
 module.exports = {
-  addFeatureFlagValidator,
+  addFeatureFlag,
 };
--- a/middlewares/validators/flag.js
+++ b/middlewares/validators/flag.js
@@ -1,6 +1,6 @@
 const joi = require("joi");
 
-const addFeatureFlag = async (req, res, next) => {
+const addFeatureFlagValidator = async (req, res, next) => {
   const schema = joi
     .object()
     .strict()
@@ -21,5 +21,5 @@ const addFeatureFlag = async (req, res, next) => {
 };
 
 module.exports = {
-  addFeatureFlag,
+  addFeatureFlagValidator,
 };
--- a/middlewares/validators/flag.js
+++ b/middlewares/validators/flag.js
@@ -1,0 +1,25 @@
+const joi = require("joi");
+
+const addFeatureFlagValidator = async (req, res, next) => {
+  const schema = joi
+    .object()
+    .strict()
+    .keys({
+      title: joi.string().required(),
+      enabled: joi.boolean().required(),
+      roles: joi.array().items(joi.string()).required(),
+      users: joi.object().required(),
+    });
+
+  try {
+    await schema.validateAsync(req.body);
+    next();
+  } catch (error) {
+    logger.error(`Error validating createTask payload : ${error}`);
+    res.boom.badRequest(error.details[0].message);
+  }
+};
+
+module.exports = {
+  addFeatureFlagValidator,
+};
--- a/models/flag.js
+++ b/models/flag.js
@@ -2,7 +2,7 @@ const firestore = require("../utils/firestore");
 const flagModel = firestore.collection("featureFlags");
 
 /**
- *
+ * Model function to add flagData in featureFlags collection of linked firestore
  * @param flagData { Object }: flag data object to be stored in DB
  * @returns Flag Id: String
  */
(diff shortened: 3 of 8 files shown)


middlewares/validators/flag.js
```diff
@@ -0,0 +1,25 @@
...
+      enabled: joi.boolean().required(),
+      roles: joi.array().items(joi.string()).required(),
+      users: joi.object().required(),
+    });
+
+  try {
+    await schema.validateAsync(req.body);
+    next();
+  } catch (error) {
+    logger.error(`Error validating createTask payload : ${error}`);
```
pallabez, line 18: It will be `Error in validating Feature Flag` right?

test/unit/models/flag.test.js
```diff
@@ -0,0 +1,34 @@
...
+    flagId = await flagModel.addFlag(flagData);
+  });
+
+  afterEach(async function () {
+    await cleanDb();
+  });
+
+  describe("add feature flag", function () {
+    it("Should return flag Id", async function () {
+      const fetched = await flagFirestore.doc(flagId).get();
```
shubham-y, line 25: NIT: `flagSnapshot` will be a better variable name

models/flag.js
```diff
@@ -0,0 +1,21 @@
+const firestore = require("../utils/firestore");
+const flagModel = firestore.collection("featureFlags");
+
+/**
+ *
+ * @param flagData { Object }: flag data object to be stored in DB
+ * @returns Flag Id: String
+ */
```
shubham-y, line 8: Please add a description for `addFlag` function

shubham-y (commented):
Thanks for the tests 👍
Suggesting some changes

routes/flag.js
```diff
@@ -0,0 +1,10 @@
+const express = require("express");
+const router = express.Router();
+const authenticate = require("../middlewares/authenticate");
+const { authorizeUser } = require("../middlewares/authorization");
+const flagController = require("../controllers/flag");
+const { addFeatureFlagValidator } = require("../middlewares/validators/flag");
+
+router.post("/add", authenticate, authorizeUser("superUser"), addFeatureFlagValidator, flagController.addFlag);
```
pallabez, line 8: Can we use the global constant here, instead of "superUser"?

middlewares/validators/flag.js
RahulGoyal-tech: Yeah we can write that too....

controllers/flag.js
```diff
@@ -0,0 +1,24 @@
+const flagQuery = require("../models/flag");
+
+/**
+ * Adds art
```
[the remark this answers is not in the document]
RahulGoyal-tech: I do not get your point. Can you please elaborate a bit on this.

test/unit/models/flag.test.js
RahulGoyal-tech: will do. By the way what is `NIT`?

routes/flag.js
RahulGoyal-tech: Are you suggesting to do as follow :
const superUser = "superUser";
authorizeUser(superUser)

models/flag.js
RahulGoyal-tech: done

test/unit/models/flag.test.js
rajat-mehra05: > will do. By the way what is `NIT`?

NIT in this context means nit pick. Its generally used for small change requests and suggestions.

rajat-mehra05 (dismissed):
Looks good to me !

routes/flag.js
pallabez: Yes. Check `constants` > `role.js`

routes/flag.js
```diff
@@ -0,0 +1,11 @@
+const express = require("express");
+const router = express.Router();
+const authenticate = require("../middlewares/authenticate");
+const { authorizeUser } = require("../middlewares/authorization");
+const flagController = require("../controllers/flag");
+const { addFeatureFlag } = require("../middlewares/validators/flag");
+const superUser = "superUser";
```
pallabez, line 7: can you use this from `constants/roles`?

pallabez (approved):
LGTM :+1:

shubham-y (dismissed):

pallabez (dismissed):

controllers/flag.js
```diff
@@ -0,0 +1,24 @@
...
+ * Controller function to add feature flag data to firestore
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const addFlag = async (req, res) => {
+  try {
+    const flagId = await flagQuery.addFlag(req.body);
+    return res.json({
+      message: "Add feature flag successfully!",
```
ankushdharkar, line 13: Nit: "Added"

middlewares/validators/flag.js
```diff
@@ -0,0 +1,25 @@
+const joi = require("joi");
+
+const addFeatureFlag = async (req, res, next) => {
+  const schema = joi
+    .object()
+    .strict()
+    .keys({
+      title: joi.string().required(),
+      enabled: joi.boolean().required(),
+      roles: joi.array().items(joi.string()).required(),
```
ankushdharkar, line 10: Why is this required? I can see a case for us adding feature flag to individual usernames/ids, instead of using roles

routes/flag.js
```diff
@@ -0,0 +1,11 @@
+const express = require("express");
+const router = express.Router();
+const authenticate = require("../middlewares/authenticate");
+const { authorizeUser } = require("../middlewares/authorization");
+const flagController = require("../controllers/flag");
+const { addFeatureFlag } = require("../middlewares/validators/flag");
+const { SUPER_USER } = require("../constants/roles");
+
+router.post("/add", authenticate, authorizeUser(SUPER_USER), addFeatureFlag, flagController.addFlag);
```
ankushdharkar, line 9: I find the name of middleware`addFeatureFlag` a little misleading. Should it renamed to demonstrate that it validates the incoming request?

ankushdharkar: Like, maybe  'validateAddingFeatureFlagReq`?

test/integration/flag.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+describe("flags", function () {
+  let jwt;
+  let nonSuperUserJwt;
+
+  beforeEach(async function () {
+    const userId = await addUser(superUser);
+    jwt = authService.generateAuthToken({ userId });
+    const nonSuperUserId = await addUser(nonSuperUser);
+    nonSuperUserJwt = authService.generateAuthToken({ nonSuperUserId });
```
ankushdharkar, line 27: Nice!

routes/flag.js
ankushdharkar, line 9: HTTP paths may be better off with just nouns and no verbs. Here is a short read from [Google Cloud Blog](https://cloud.google.com/blog/products/api-management/restful-api-design-nouns-are-good-verbs-are-bad).

What are your views in having a POST to `/feature-flag`, instead of `/flag/add`?

test/unit/models/flag.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+  beforeEach(async function () {
+    flagId = await flagModel.addFlag(flagData);
+  });
+
+  afterEach(async function () {
+    await cleanDb();
+  });
+
+  describe("add feature flag", function () {
```
ankushdharkar, line 23: Is the module name correct? I feel we are not checking for just "add feature flag" in this scenario, but more of whether feature flags are working correctly or not.

ankushdharkar (changes requested):

routes/flag.js
RahulGoyal-tech: So i named this function as `addFeatureFlagValidator` earlier but changed to `addFeatureFlag` based upon a review. Is that name good enough or need to write some different name.

test/unit/models/flag.test.js
RahulGoyal-tech: But we are checking whether the process of adding feature flag was done correctly or not so it becomes a suitable name i guess. Do you have any other suggesstions.

routes/flag.js
pallabez: How about importing it as 
```js
const validator = require("../middlewares/validators/flag");
```
So it becomes `validator.addFeatureFlag`?

test/unit/models/flag.test.js
pallabez: In all other test cases, we are writing the function name here and not what it does. This makes it easier to find for what function the test cases are running. So ideally it should be `addFeatureFlag`. What do you think?

prakashchoudhary07:
Closing this, are we have a separate service for feature-flag service, https://github.com/Real-Dev-Squad/feature-flag-backend/ https://github.com/Real-Dev-Squad/feature-flag-service/

PR #574, opened 2022-04-18
fix logs fetch authorization

pallabez:
Default users were able to use the `/logs/:type` endpoint which were supposed to be available only to super users


lakshayman (approved):
Just an import change.
Looks good to me.

PR #584, opened 2022-04-26
Fix UpdateUser Function

### LINKED ISSUE #579: Fix updateUser Function
Currently, the updateUser function is getting data to update in req.body, now in req.body you will get only profileDiffId so fetch the data from profile and update that.

RahulGoyal-tech:
closes #579

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -198,8 +198,10 @@ const updateUser = async (req, res) => {
   try {
     const { id: profileDiffId, message } = req.body;
 
-    const { approval, timestamp, userId, ...profileDiff } = await profileDiffsQuery.fetchProfileDiff(profileDiffId);
-    if (!profileDiff) return res.boom.notFound("Profile Diff doesn't exist");
+    const profileDiffData = await profileDiffsQuery.fetchProfileDiff(profileDiffId);
+    if (!profileDiffData) return res.boom.notFound("Profile Diff doesn't exist");
+
+    const { approval, timestamp, userId, ...profileDiff } = profileDiffData;
 
     const user = await userQuery.fetchUser({ userId });
     if (!user.userExists) return res.boom.notFound("User doesn't exist");
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -199,16 +199,13 @@ const updateUser = async (req, res) => {
     const { id: profileDiffId, message } = req.body;
 
     const { approval, timestamp, userId, ...profileDiff } = await profileDiffsQuery.fetchProfileDiff(profileDiffId);
+    if (!profileDiff) return res.boom.notFound("Profile Diff doesn't exist");
 
     const user = await userQuery.fetchUser({ userId });
     if (!user.userExists) return res.boom.notFound("User doesn't exist");
 
-    const profileResponse = await profileDiffsQuery.updateProfileDiff(
-      { approval: profileDiffStatus.APPROVED },
-      profileDiffId
-    );
+    await profileDiffsQuery.updateProfileDiff({ approval: profileDiffStatus.APPROVED }, profileDiffId);
 
-    if (profileResponse.notFound) return res.boom.notFound("Profile Diff doesn't exist");
     await userQuery.addOrUpdate(profileDiff, userId);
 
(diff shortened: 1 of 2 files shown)


lakshayman (approved):
Looks good to me.

models/profileDiffs.js
```diff
@@ -24,6 +24,22 @@ const fetchProfileDiffs = async () => {
...
 
+/**
+ * Fetches the profileDiff data of the provided profileDiff Id
+ * @param profileDiffId profileDiffId of the diffs need to be fetched
+ * @returns profileDiff Data
+ */
+const fetchProfileDiff = async (profileDiffId) => {
+  try {
+    const Diff = await profileDiffsModel.doc(profileDiffId).get();
+    const profileDiffData = Diff.data();
```
pallabez, line 35:
```suggestion
    const profileDiff = await profileDiffsModel.doc(profileDiffId).get();
    const profileDiffData = profileDiff.data();
```

controllers/users.js
```diff
@@ -196,8 +196,9 @@ const postUserPicture = async (req, res) => {
  */
 const updateUser = async (req, res) => {
   try {
-    const userId = req.params.userId;
-    const { id: profileDiffId, message, ...profileDiff } = req.body;
+    const { id: profileDiffId, message } = req.body;
+
+    const { approval, timestamp, userId, ...profileDiff } = await profileDiffsQuery.fetchProfileDiff(profileDiffId);
```
pallabez, line 201: I believe we should handle error here when we get we get empty data and remove the error handler at line 211
```js
if (profileResponse.notFound) return res.boom.notFound("Profile Diff doesn't exist");
```

controllers/users.js
```diff
@@ -196,18 +196,16 @@ const postUserPicture = async (req, res) => {
...
+    if (!profileDiff) return res.boom.notFound("Profile Diff doesn't exist");
```
pallabez, line 202: Does this throw `notFound` error when we pass a random profileDiffId? Was it tested?

RahulGoyal-tech: This function doesn't check for that. It only checks if profileDiff is empty the throws an error.

pallabez: So can we do something like this? 
```js
    const profileDiffData = await profileDiffsQuery.fetchProfileDiff(profileDiffId);
    if (!profileDiffData) return res.boom.notFound("Profile Diff doesn't exist");
    
    const { approval, timestamp, userId, ...profileDiff } = profileDiffData;
```

RahulGoyal-tech: What will be the advantage of this change?

pallabez: Try passing a random profileDiffId and compare the result.

RahulGoyal-tech: Got it 👍

pallabez (approved):
Looks good to me :+1:

RahulGoyal-tech:
@lakshayman when will we be merging this PR.

controllers/users.js
```diff
@@ -196,18 +196,18 @@ const postUserPicture = async (req, res) => {
  */
 const updateUser = async (req, res) => {
   try {
-    const userId = req.params.userId;
-    const { id: profileDiffId, message, ...profileDiff } = req.body;
+    const { id: profileDiffId, message } = req.body;
+
+    const profileDiffData = await profileDiffsQuery.fetchProfileDiff(profileDiffId);
+    if (!profileDiffData) return res.boom.notFound("Profile Diff doesn't exist");
```
Neha, line 202: Move hardcoded content to constant file

controllers/users.js
```diff
@@ -196,18 +196,18 @@ const postUserPicture = async (req, res) => {
...
+
+    const { approval, timestamp, userId, ...profileDiff } = profileDiffData;
 
     const user = await userQuery.fetchUser({ userId });
     if (!user.userExists) return res.boom.notFound("User doesn't exist");
```
Neha, line 207: Personal view: Do not drop {} it is hard for folks to read. We use bundler and they take care of the optimization of the code.

Rule: Make it easy to read.

Issue #567, opened 2022-04-14
GET /flag endpoint


pallabez:
GET /flag endpoint

## Parent Issue
- #563 

## Steps

Branch - `feature-flag/fetch`
- [x] Checkout `feature-flag/add` and create a new branch `feature-flag/fetch` from there.
- [x] Create GET route `/flag` and set authentication to `super_user`
- [x] Create controller and model (function name - `fetchFlags` for both)

## res.body
```
message: "Flags returned successfully!",
flags: [
    {  
       flagId: <flagId>
       title: dark-mode
       enabled: true
       roles: [members, no_role, super_user],
       users: {
           "pcBpdEQ5ydRakHHSpY2p": true,
           "quEpdE34ydRakJSpY45p": false,
        },
    },
    {
         flag_object2...
    }
]
```
-----

Branch - `feature-flag/fetch-tests-validators`
- [ ] Write integration and unit test

pallabez:
Now Handled in the repo 
- https://github.com/Real-Dev-Squad/feature-flag-service

PR #586, opened 2022-04-28
Feature flag/fetch

akash2819:
Added api end point for feature flags


controllers/flag.js
```diff
@@ -0,0 +1,37 @@
...
+    logger.error(`Error while adding featureFlag info: ${err}`);
+    return res.boom.badImplementation("Something went wrong please contact admin");
+  }
+};
+const fetchFlags = async (req, res) => {
+  try {
+    const allFlags = await flagQuery.fetchFlags();
+    return res.json({
+      message: allFlags.length ? "Flags returned successfully!" : "No flag found",
+      flagQuery: allFlags,
```
pallabez, line 26:
```suggestion
    const flags = await flagQuery.fetchFlags();
    return res.json({
      message: flags.length ? "Flags returned successfully!" : "No flag found",
      flags
```

Can we keep it accurate to the res.body? 
- https://github.com/Real-Dev-Squad/website-backend/issues/567

models/flag.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+};
+
+const fetchFlags = async () => {
+  try {
+    const flag = await flagModel.get();
+    const flags = [];
+    flag.forEach((doc) => {
+      const flagdata = doc.data();
+      flag.push({
```
pallabez, line 25: It should be `flags.push` right?

akash2819: Yes ,  Sorry for that :(
Now resolved Please check

prakashchoudhary07:
Closing this, as feature flag service is being maintained in other project
https://github.com/Real-Dev-Squad/feature-flag-service/ https://github.com/Real-Dev-Squad/feature-flag-backend/

PR #592, opened 2022-05-03
done with all 3 apis

spverma536:


controllers/newsletter.js
```diff
@@ -0,0 +1,46 @@
...
+const addEmail = async (req, res) => {
+  try {
+    const a = await newsletters.subscribe(req.body);
+    return res.json({
+      message: "gotcha",
+      response: a,
+    });
+  } catch (err) {
+    logger.error("control failed");
+    return res.boom.badImplementation("not working");
```
pallabez, line 12: Can we be more descriptive about the error here?

controllers/newsletter.js
```diff
@@ -0,0 +1,46 @@
+const newsletters = require("../models/newsletter");
+
+const addEmail = async (req, res) => {
+  try {
+    const a = await newsletters.subscribe(req.body);
+    return res.json({
+      message: "gotcha",
+      response: a,
```
pallabez, line 8: What are we getting in a? Can we use better variable name here?

controllers/newsletter.js
spverma536: done

controllers/newsletter.js
spverma536: done

rohan09-raj:
@spverma536  Can you please link the issue with PR in the description ??
Closes #<ISSUE_NUMBER>
Also, can you please change the title of the PR to something relevant to the code, something like Newsletter API and add some description about the APIs to give more context to the reviewers.

controllers/newsletter.js
```diff
@@ -0,0 +1,48 @@
+const newsletters = require("../models/newsletter");
+
+const ERROR_MESSAGE = "Something went wrong. Please try again or contact admin";
+
+const addEmail = async (req, res) => {
+  try {
+    const response = await newsletters.subscribe(req.body);
+    return res.json({
+      message: "gotcha",
```
rohan09-raj, line 9: Can we please change the message to something like `Email added successfully` ?

controllers/newsletter.js
```diff
@@ -0,0 +1,48 @@
...
+    logger.error("Logging error from getMailingList controllers");
+    return res.boom.badImplementation(ERROR_MESSAGE);
+  }
+};
+
+const removeEmail = async (req, res) => {
+  try {
+    const response = await newsletters.unsubscribe(req.body);
+    return res.json({
+      message: "done",
```
rohan09-raj, line 35: Here too, we can change the message to something like `Email unsubscribed successfully`

models/newsletter.js
```diff
@@ -0,0 +1,78 @@
...
+
+const DB_ERROR = "Not able to conenct to DB";
+
+const subscribe = async (data) => {
+  try {
+    const { email } = data;
+    const check = await newsletterModel.where("email", "==", email).get();
+    let exist = false;
+    check.forEach((ch) => {
+      const data = ch.data();
```
rohan09-raj, line 13: Can we change the argument name `ch` to something more meaningful

controllers/newsletter.js
spverma536: done

controllers/newsletter.js
```diff
@@ -0,0 +1,48 @@
...
+    logger.error("Logging error from addEmail controllers");
+    return res.boom.badImplementation(ERROR_MESSAGE);
+  }
+};
+
+const getMailingList = async (req, res) => {
+  try {
+    const list = await newsletters.getMailingList();
+    return res.json({
+      message: "here is your list",
```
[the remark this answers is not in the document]
spverma536: done

models/newsletter.js
```diff
@@ -0,0 +1,78 @@
...
+
+const unsubscribe = async (data) => {
+  try {
+    const { email } = data;
+    let exist = false;
+    const check = await newsletterModel.where("email", "==", email).get();
+    check.forEach((ch) => {
+      const data = ch.data();
+      if (data.active) {
+        const res = ch.id;
```
rohan09-raj, line 59: Here too, can we change `ch`  to something more meaningful ??

models/newsletter.js
spverma536: done

models/newsletter.js
```diff
@@ -0,0 +1,78 @@
...
+      exist = true;
+    });
+    if (!exist) {
+      await newsletterModel.add({
+        email: email,
+        active: true,
+      });
+      return "new added";
+    }
+    return "already added in the list";
```
rohan09-raj, line 30: This can be something like `Email already present in the mailing list`

models/newsletter.js
```diff
@@ -0,0 +1,78 @@
...
+        });
+      }
+      exist = true;
+    });
+    if (!exist) {
+      await newsletterModel.add({
+        email: email,
+        active: true,
+      });
+      return "new added";
```
rohan09-raj, line 28: `New Email Added to Mailing List`

models/newsletter.js
spverma536: done

routes/newsletter.js
```diff
@@ -0,0 +1,12 @@
+const express = require("express");
+const router = express.Router();
+const newsletters = require("../controllers/newsletter");
+const { validateEmail } = require("../middlewares/validators/newsletter");
+
+router.post("/subscribe", validateEmail, newsletters.addEmail);
```
pallabez, line 6: Is there any reason we are going with `post` request here? Patch request might be better suited as we are updating data as well. 

ref- https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH

MehulKChaudhari (commented):
LGTM, very nicely done. 

But tests would be better to write.

routes/newsletter.js
harshith-venkatesh: Post also works the same @pallabez , as internally in the controller piyush is handling for both condition here

middlewares/validators/newsletter.js
```diff
@@ -0,0 +1,19 @@
+const joi = require("joi");
+
+const validateEmail = async (req, res, next) => {
+  const schema = joi.object().strict().keys({
+    email: joi.string().required(),
```
harshith-venkatesh, line 5: Could we look into validation of email via joi....
Since this is particularly on newsletter service via email
I feel We can add extra validation checks for emails

controllers/newsletter.js
```diff
@@ -0,0 +1,48 @@
+const newsletters = require("../models/newsletter");
+
+const ERROR_MESSAGE = "Something went wrong. Please try again or contact admin";
+
+const addEmail = async (req, res) => {
+  try {
+    const response = await newsletters.subscribe(req.body);
+    return res.json({
+      message: "Thank you for subscribing RDS Newsletter!",
+      response: response,
```
harshith-venkatesh, line 10: Here both message and response convey similar message right....
Is this the right way to proceed with?

harshith-venkatesh (changes requested):

middlewares/validators/newsletter.js
spverma536: good point, will check validation for email.

spverma536: added @harshith-venkatesh , pls have a look

controllers/newsletter.js
spverma536: any suggestions?

middlewares/validators/newsletter.js
pallabez: Can't we use email validator of joi? 

Ref - https://joi.dev/api/?v=17.6.0

spverma536: thanks, done

routes/newsletter.js
```diff
@@ -0,0 +1,12 @@
...
+
+router.get("/", newsletters.getMailingList);
+
+router.post("/unsubscribe", validateEmail, newsletters.removeEmail);
```
pallabez, line 10: Shouldn't we remove this as discussed? 
Currently, if we add this then I (or any user) can disable newsletter for any email.

spverma536: I don't remember being discussed where we are removing unsubscribe feature. we discussed that it is necessary as if user wants to discontinue then he will just report spam to RDS mailing id if he/she does not see option of unsubscribe.
@pallabez to address your question, we are not going to give an unsubscribe option on website, it will be inside the newsletter mails only and when user clicks that link, his/her id will be shared as per the account where newsletter was sent, so NO other person can unsubscribe the newsletter for you. @rohan09-raj  pls verify once if that is not the case or I missed some discussion

pallabez: >  his/her id will be shared as per the account where newsletter was sent.
So we are sending the newsletter document Id?

PR #599, opened 2022-05-08
Fixes getProfileDiff function

arun-esh:
User data was visible without encryption. In this PR, Email and Phone number are somwwhat encypted.
Showing only first and last two characters of User's Phone number.
For emails, only first two characters and last four characters are visible.
Rest of the details of Email and Phone are hidden using Asterisk.

Approve and Timestamp fields are not allowed. They are not returning in getProfileDiff function
as a field in user data.

Before:
Picture1

Email and Phone numbers are available.
Approve and Timestamps are available 

After:
Picture13

Emails and Phone numbers are encrypted.
Approve and Timestamps are not available.


models/profileDiffs.js
```diff
@@ -15,6 +15,24 @@ const fetchProfileDiffs = async () => {
...
+
+        phone:
+          doc.data().phone.substring(0, 2) +
+          doc
+            .data()
+            .phone.substring(3, doc.data().phone.length - 1)
+            .replace(/./g, "*") +
+          doc.data().phone.substring(doc.data().phone.length - 2),
+
+        approval: undefined,
```
lakshayman, line 34: don't set these undefined just remove them using destructuring https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring

lakshayman (changes requested):

lakshayman (approved):

Issue #598, opened 2022-05-08
Absolute import 


shreya-mishra:
Convert all relative imports to Absolute imports.

aakankshadhurandhar:
@shreya-mishra  Can you assign me this?

shreya-mishra:
> @shreya-mishra Can you assign me this?

Yea sure! I can assign you, no issues. let me know when you raised the PR and also let me know the ETA ...

shreya-mishra:
assigned the ticket! let me know the ETA, please ....

aakankshadhurandhar:
@shreya-mishra  ETA is monday - 16 may

shreya-mishra:
@aakankshadhurandhar Any updates ??

ZunedKhan07:
Hey! I would like to work on this issue.  
Please assign it to me if it's not already taken.  
Thank you!

PR #607, opened 2022-05-18
feat: added eslint for absolute imports

aakankshadhurandhar:
https://github.com/Real-Dev-Squad/website-backend/issues/598


shreya-mishra:
@aakankshadhurandhar can we hop on a short call just wanted to know what we are doing here ?

Issue #615, opened 2022-05-22
Replace authorizeUsers middleware with authorizeRoles middleware


pallabez:
Parent Ticket 
- https://github.com/Real-Dev-Squad/website-backend/issues/582

## Places where authorizeUsers are used

{{Mention and link it here}}

poonammohata:
@pallabez i would like to take this task

PR #622, opened 2022-06-20
Feature/update authorize role

poonammohata:
Resolves - #615

- replaced authorizeUsers middleware with authorizeRoles middleware
- removed LEGACY_ROLES from 'constants.roles.js' file
- where Roles are passed as string in authorize middleware updated that to Role's key

### DIFF
--- a/constants/roles.js
+++ b/constants/roles.js
@@ -1,10 +1,4 @@
-// Use Legacy Roles with authorizeUsers middleware
-const LEGACY_ROLES = {
-  SUPER_USER: "superUser",
-  APP_OWNER: "appOwner",
-};
-
 // Use Roles with authorizeRoles middleware
 const ROLES = { SUPERUSER: "super_user", APPOWNER: "app_owner", MEMBER: "member", ARCHIVED: "archived" };
 
-module.exports = { LEGACY_ROLES, ROLES };
+module.exports = { ROLES };
--- a/constants/roles.js
+++ b/constants/roles.js
@@ -6,4 +6,4 @@ const ROLES = {
   ARCHIVED: "archived",
 };
 
-module.exports = { ROLES };
+module.exports = ROLES;
--- a/constants/roles.js
+++ b/constants/roles.js
@@ -1,4 +1,9 @@
 // Use Roles with authorizeRoles middleware
-const ROLES = { SUPERUSER: "super_user", APPOWNER: "app_owner", MEMBER: "member", ARCHIVED: "archived" };
+const ROLES = {
+  SUPERUSER: "super_user",
+  APPOWNER: "app_owner",
+  MEMBER: "member",
+  ARCHIVED: "archived",
+};
 
 module.exports = { ROLES };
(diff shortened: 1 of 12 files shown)


constants/roles.js
```diff
@@ -1,10 +1,4 @@
-// Use Legacy Roles with authorizeUsers middleware
-const LEGACY_ROLES = {
-  SUPER_USER: "superUser",
-  APP_OWNER: "appOwner",
-};
-
```
pallabez, line 6: Thanks for removing this as well.

routes/tasks.js
```diff
@@ -159,7 +162,7 @@ router.post("/", authenticate, authorizeUser("appOwner"), createTask, tasks.addN
  *           schema:
  *             $ref: '#/components/schemas/errors/badImplementation'
  */
-router.patch("/:id", authenticate, authorizeUser("appOwner"), updateTask, tasks.updateTask);
+router.patch("/:id", authenticate, authorizeRoles([APPOWNER]), updateTask, tasks.updateTask);
```
pallabez, line 165: Previous middleware `authorizeUser("appOwner")` gives permission to both app_owener and super_user access. 
The current only gives it to mentioned role. So can we add SUPERUSER to the array as well? 
```suggestion
router.patch("/:id", authenticate, authorizeRoles([APPOWNER, SUPERUSER]), updateTask, tasks.updateTask);
```

poonammohata: okh that means where ever APPOWNER role is given, we need to pass SUPERUSER role also in that array, am i understanding it right? @pallabez

pallabez (dismissed):
LGTM :+1:

routes/tasks.js
```diff
@@ -3,13 +3,16 @@ const router = express.Router();
...
+} = require("../constants/roles");
 
 router.get("/", tasks.fetchTasks);
 router.get("/self", authenticate, tasks.getSelfTasks);
-router.get("/overdue", authenticate, authorizeUser("superUser"), tasks.overdueTasks);
-router.post("/", authenticate, authorizeUser("appOwner"), createTask, tasks.addNewTask);
-router.patch("/:id", authenticate, authorizeUser("appOwner"), updateTask, tasks.updateTask);
+router.get("/overdue", authenticate, authorizeRoles([SUPERUSER]), tasks.overdueTasks);
+router.post("/", authenticate, authorizeRoles([APPOWNER]), createTask, tasks.addNewTask);
+router.patch("/:id", authenticate, authorizeRoles([APPOWNER]), updateTask, tasks.updateTask);
```
pallabez, line 15:
```suggestion
router.post("/", authenticate, authorizeRoles([APPOWNER, SUPERUSER]), createTask, tasks.addNewTask);
router.patch("/:id", authenticate, authorizeRoles([APPOWNER, SUPERUSER]), updateTask, tasks.updateTask);
```

Previously `appOwner` gave permission to both app_owner and super_user. 
Now it gives access to only the role mentioned.

constants/roles.js
```diff
@@ -1,10 +1,4 @@
...
 // Use Roles with authorizeRoles middleware
 const ROLES = { SUPERUSER: "super_user", APPOWNER: "app_owner", MEMBER: "member", ARCHIVED: "archived" };
```
akshay1502, line 2: Can we have each key value pair on a new line.

poonammohata: made the changes

pallabez (approved):
Lgtm

pallabez (approved):
Thanks for the changes and keeping track of it

PR #630, opened 2022-07-07
resolved the issue of archieve member not working

shreya-mishra:


pallabez:
@shreya-mishra @ankushdharkar What are we planning to do with `archive_member`. Make him unable to use any api or just not show up in GET `/members` call?

Issue #608, opened 2022-05-18
Deprecate `/members` API


shubham-y:
Deprecating the [/members](https://api.realdevsquad.com/members) endpoint as the same purpose is served by the [/users](https://api.realdevsquad.com/users) endpoint. 

Changes needed related to `/members` endpoint-
- [ ] Remove GET `/members` from members route
- [ ] Remove `getMembers` function from members controller
- [ ] Remove `fetchUsers` function from members model
- [ ] Remove Integration tests for `/members`
- [ ] Remove API contract for `/members`
- [ ] Update `/users` endpoint with a filter to return users with `role : { members: true }`

Please feel free to add more points if more changes are required

Post Merging changes:

- Update [config](https://github.com/Real-Dev-Squad/website-backend/blob/912c397f8f229893f789477f39be99051ac17632/config/default.js#L77-L79) to replace `/members` with `/users` for caching

aakankshadhurandhar:
@shubham-y  Can you assign me this issue ?

shubham-y:
> @shubham-y Can you assign me this issue ?

Done! Assigned it to you. Please do let me know the ETA

pallabez:
Aren't we using `fetcheUsers` model function in a lot of places?

aakankshadhurandhar:
> > @shubham-y Can you assign me this issue ?
> 
> Done! Assigned it to you. Please do let me know the ETA

ETA 24th May

shubham-y:
> Aren't we using `fetcheUsers` model function in a lot of places?
> 
> Also, why are we removing GET `/users`? Shouldn't we try to migrate GET `/members` functionality to GET `/users` and remove GET `/members` instead?
> 
> Reason: `/members` is giving us non member details which kind of doesn't make sense.

Yes, it makes sense. Updated the issue details

RitikJaiswal75:
closing in lieu of: https://github.com/Real-Dev-Squad/website-members/issues/547

Issue #601, opened 2022-05-10
Archived members data not working


pallabez:
## After archiving a member (In database): 
![image](https://user-images.githubusercontent.com/63191499/167679802-9d1b38ce-7167-4f3f-bf5a-77636d19cbf7.png)
## In fixture: 
https://github.com/Real-Dev-Squad/website-backend/blob/cc9476dec38137093f7108bcd2566b4bb5b99053/test/fixtures/user/user.js#L112-L117

pallabez:
Are they both different or I'm missing something?

samarpan1738:
I'll be working on this issue

samarpan1738:
This issue will be resolved after all the PR mentioned in the below ticket are merged
https://github.com/Real-Dev-Squad/website-backend/issues/672

PR #631, opened 2022-07-09
feature:depreciated member api

aakankshadhurandhar:
https://github.com/Real-Dev-Squad/website-backend/issues/608


models/users.js
```diff
@@ -59,6 +59,7 @@ const addOrUpdate = async (userData, userId = null) => {
 const fetchUsers = async (query) => {
   try {
     const snapshot = await userModel
+      .where("isMember", "==", true)
```
shubham-y, line 62: We do not have to return users with Members role by default, instead we should filter the users based on role provided as query params by the client and in case no role is provided we return the data of all the users.
Eg: `/users?role=members`

aakankshadhurandhar:
Screenshot from 2022-07-25 18-23-39
Screenshot from 2022-07-25 18-23-23

models/users.js
```diff
@@ -58,10 +58,17 @@ const addOrUpdate = async (userData, userId = null) => {
...
       .limit(parseInt(query.size) || 100)
       .offset((parseInt(query.size) || 100) * (parseInt(query.page) || 0))
       .get();
+    if (query.role === "members") {
+      snapshot = await userModel
+        .limit(parseInt(query.size) || 100)
+        .offset((parseInt(query.size) || 100) * (parseInt(query.page) || 0))
+        .where("isMember", "==", true)
+        .get();
+    }
```
shubham-y, line 71: There seem to be repeated lines of code. We can avoid it by calling the `get()` method at the end
```suggestion
    let query = userModel
    query = query.limit(parseInt(query.size) || 100)
      .offset((parseInt(query.size) || 100) * (parseInt(query.page) || 0))
      .get();
    if (query.role === "members") {
      query= query.where("isMember", "==", true);
    }
    const snapshot = await query.get();
```

shubham-y: Also, we can make the API more generic by allowing the client to pass any valid role and we can return the users with that role. The client can also pass a list of comma-separated roles and we return all users with those roles.

shubham-y (commented):
Good work 👍
Suggesting a few changes. Please take a look.

models/users.js
shubham-y: Please avoid using string literal "members" `query.role === "members"`, instead import roles from the [roles.js](https://github.com/Real-Dev-Squad/website-backend/blob/develop/constants/roles.js) file

models/users.js
```diff
@@ -58,10 +58,17 @@ const addOrUpdate = async (userData, userId = null) => {
  */
 const fetchUsers = async (query) => {
   try {
-    const snapshot = await userModel
+    let snapshot = await userModel
       .limit(parseInt(query.size) || 100)
       .offset((parseInt(query.size) || 100) * (parseInt(query.page) || 0))
       .get();
+    if (query.role === "members") {
```
ankushdharkar, line 65: @aakankshadhurandhar Could we move the `members` string into an enum or a constants file please?

ankushdharkar (changes requested):
Changes requested in conversation.

models/users.js
```diff
@@ -58,10 +58,17 @@ const addOrUpdate = async (userData, userId = null) => {
  */
 const fetchUsers = async (query) => {
   try {
-    const snapshot = await userModel
+    let snapshot = await userModel
```
ankushdharkar, line 61: So, if `query.role === "members"`, we should still be making this call for no reason, correct?

models/users.js
aakankshadhurandhar: @ankushdharkar  I have imported it directly now from [roles.js](https://github.com/Real-Dev-Squad/website-backend/blob/develop/constants/roles.js)

samarpan1738:
Hi @aakankshadhurandhar . I think #601 is similar to the scope of this task. Can we discuss this?
Also I wanted to know how soon will the /members api be deprecated

models/users.js
aakankshadhurandhar: Done

aakankshadhurandhar:
As discussed with @shubham-y  closing this PR as we now have archived members . Seperate PR for depreciating member API will be raised .

aakankshadhurandhar:
As discussed with @shubham-y  closing this PR as we now have archived members . Seperate PR for depreciating member API will be raised .

PR #643, opened 2022-07-30
Enhance idle member API

### LINKED ISSUE #619: Enchance Idle Members API
- In the following, we are getting only the members and fetching the members having active tasks but do we also need to check if the user status is `ACTIVE` or `IDLE` or `OOO`
- Also, we are sending only the usernames from the backend but in frontend, for availability panel, we need user image to be shown from the user data and also for Tasks level feature we need users existing data so, it's if we send complete data of user instead of only usernames

https://github.com/Real-Dev-Squad/website-backend/blob/27ff92b7fdb37af413d93167b332ed3bd56800aa/controllers/members.js#L36-L51

SanthoshBalagur:
Closes #619

### DIFF
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -1,7 +1,6 @@
-const { ROLES } = require("../constants/users");
 const members = require("../models/members");
+const users = require("../models/users");
 const tasks = require("../models/tasks");
-const { fetchUser } = require("../models/users");
 
 const ERROR_MESSAGE = "Something went wrong. Please try again or contact admin";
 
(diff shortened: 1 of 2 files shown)


models/users.js
```diff
@@ -82,6 +82,28 @@ const fetchUsers = async (query) => {
   }
 };
 
+const queryUser = async (query) => {
+  try {
+    let call = userModel;
+    Object.keys(query).forEach((key) => {
+      // eslint-disable-next-line security/detect-object-injection
+      call = call.where(key, "==", query[key]);
```
alishvagadiya, line 90: It's not good practice to use dynamic column name selection. checkout and read xss injection

controllers/members.js
```diff
@@ -1,7 +1,6 @@
-const { ROLES } = require("../constants/users");
 const members = require("../models/members");
+const users = require("../models/users");
```
alishvagadiya, line 2: Taking a complete model into one object, it's better to use the import required particular function, using destructing.

controllers/members.js
```diff
@@ -35,14 +34,13 @@ const getMembers = async (req, res) => {
 
 const getIdleMembers = async (req, res) => {
   try {
-    const onlyMembers = await members.fetchUsersWithRole(ROLES.MEMBER);
+    const allusers = await users.queryUser({ status: "idle" });
```
alishvagadiya, line 37: Variable name should be `idleUsers` instead of `allusers`

alishvagadiya (changes requested):
Look into the requested changes and comment in this pull request.

ankushdharkar:
Will be taken care of in another PR. The initial author has been MIA.

Issue #633, opened 2022-07-24
Refactor Roles List constant and export it


ankushdharkar:
Export object at [this location](https://github.com/Real-Dev-Squad/website-backend/blob/18aa66dd9331dc2e1b44feb3073608ed502bc12d/middlewares/authorization.js#L8-L12) as just `ROLES_LIST`, so that it can be used in code as `ROLES_LIST.SUPER_USER`

(e.g  [here](https://github.com/Real-Dev-Squad/website-backend/blob/ccc712575ba3dfb7e35a0f9fb92659d2fcbccce1/routes/tasks.js#L101-L102) we can then change it to  `authorizeUser(ROLES_LIST.SUPER_USER)`, and prevent issues arising due to typos like "super-user", or "superUser", or "super_user". In ideal world, we will soon have TypeScript!)

pallabez:
Issue has been resolved on 
- https://github.com/Real-Dev-Squad/website-backend/issues/582

Issue #582, opened 2022-04-23
Update authorizeUser middleware


pallabez:
Checkout why this update on
- #576

### Create authorizeRoles
- [ ] No role hierarchy 
- [ ] Throws error on passing invalid arguments
- [ ] Integration tests
- [ ] Unit tests

###  Migrate authorizeUser to authorizeRoles
- [ ] Define global contants for `super_user` , `app_owner` & `member`
- [ ] Tests out the middleware in prod on feature-flag (old) API
- [ ] Make changes on all API
- [ ] Clean up authorizeUser

PR #646, opened 2022-07-30
Removed duplicate ROLES constant

### LINKED ISSUE #644: [RFC] Remove duplicate ROLES constant
Currently we have ROLES constant object defined in 2 files .

1. [constants/user.js](https://github.com/Real-Dev-Squad/website-backend/blob/15e6242972b7f674af869b618ed42a1c5ae080fd/constants/users.js#L2)
```
const ROLES = {
  ADMIN: "admin",
  APPOWNER: "app_owner",
  DEFAULT: "default",
  MEMBER: "member",
  SUPERUSER: "super_user",
};
```

2.  [constants/roles.js](https://github.com/Real-Dev-Squad/website-backend/blob/15e6242972b7f674af869b618ed42a1c5ae080fd/constants/roles.js#L8)
```
const ROLES = { 
  SUPERUSER: "super_user", 
  APPOWNER: "app_owner", 
  MEMBER: "member"
};
```

Ideally ROLES should only be defined in the file dedicated for it i.e. `/constants/roles.js`

samarpan1738:
* Removed ROLES constant from user.js
* Changed import paths for the same in depending files

fix: Passing correct role (superUser) to authorizeUser instead of undefined in logs route

Closes #644

### DIFF
--- a/routes/logs.js
+++ b/routes/logs.js
@@ -3,8 +3,8 @@ const router = express.Router();
 const logs = require("../controllers/logs");
 const authenticate = require("../middlewares/authenticate");
 const { authorizeUser } = require("../middlewares/authorization");
-const { ROLES } = require("../constants/users");
+const { LEGACY_ROLES } = require("../constants/roles");
 
-router.get("/:type", authenticate, authorizeUser(ROLES.SUPER_USER), logs.fetchLogs);
+router.get("/:type", authenticate, authorizeUser(LEGACY_ROLES.SUPER_USER), logs.fetchLogs);
 
 module.exports = router;
--- a/constants/users.js
+++ b/constants/users.js
@@ -1,14 +1,7 @@
 const userStatusEnum = ["ooo", "idle", "active"];
-const ROLES = {
-  ADMIN: "admin",
-  APPOWNER: "app_owner",
-  DEFAULT: "default",
-  MEMBER: "member",
-  SUPERUSER: "super_user",
-};
 const profileStatus = {
   PENDING: "PENDING",
   APPROVED: "APPROVED",
   NOT_APPROVED: "NOT APPROVED",
 };
-module.exports = { userStatusEnum, ROLES, profileStatus };
+module.exports = { userStatusEnum, profileStatus };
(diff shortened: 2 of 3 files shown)


samarpan1738:
Also added the fix for passing correct parameter to authorizeUser for /logs/:type route

routes/logs.js
```diff
@@ -3,8 +3,7 @@ const router = express.Router();
 const logs = require("../controllers/logs");
 const authenticate = require("../middlewares/authenticate");
 const { authorizeUser } = require("../middlewares/authorization");
-const { ROLES } = require("../constants/users");
 
-router.get("/:type", authenticate, authorizeUser(ROLES.SUPER_USER), logs.fetchLogs);
+router.get("/:type", authenticate, authorizeUser("superUser"), logs.fetchLogs);
```
rohan09-raj, line 7: Are we sure that, this is the correct role for the super user ??
At most places I guess `super_user` is the role and can we please import it from a constant from `constants/roles.js`

samarpan1738: The parameter for authorizeUser can have these 3 values : superUser, appOwner, default (Check out https://github.com/Real-Dev-Squad/website-backend/blob/15e6242972b7f674af869b618ed42a1c5ae080fd/middlewares/authorization.js#L8)

There's an existing ticket for moving this parameter value to a constant: https://github.com/Real-Dev-Squad/website-backend/issues/633

rohan09-raj: `superUser` is a Legacy role and will not be used further I guess, and also the file you are pointing to, the actual value is `super_user` only.

rohan09-raj: We can use the LEGACY ROLE from constants for now, so that things work now. After the middleware changes, we will take care of it in refactoring

samarpan1738: As discussed we can use superUser legacy role currently and #582 will remove this later

prakashchoudhary07 (dismissed):
LGTM

pallabez (dismissed):
LGTM :+1:

prakashchoudhary07 (approved):

ankushdharkar (approved):

PR #647, opened 2022-07-31
new paginated functionality added in GET /tasks

ivinayakg:
tests are yet to be created
[API Contract](https://github.com/Real-Dev-Squad/website-api-contracts/issues/83)

### DIFF
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -44,9 +44,27 @@ const updateTask = async (taskData, taskId = null) => {
  *
  * @return {Promise<tasks|Array>}
  */
-const fetchTasks = async () => {
+const fetchTasks = async (limit, offset) => {
+  let lastDoc;
+
+  //if offset given
+  if (offset) {
+    lastDoc = await tasksModel.doc(offset).get();
+    lastDoc = lastDoc.data();
+  }
+
+  // default case for limit
+  if (!limit) {
+    limit = 20;
+  }
+  limit = Number(limit);
+
   try {
-    const tasksSnapshot = await tasksModel.get();
+    const tasksSnapshot = await tasksModel
+      .orderBy("startedOn", "desc")
+      .startAfter(lastDoc ? lastDoc.startedOn : "")
+      .limit(limit)
+      .get();
     const tasks = buildTasks(tasksSnapshot);
     const promises = tasks.map(async (task) => fromFirestoreData(task));
     const updatedTasks = await Promise.all(promises);
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -46,25 +46,15 @@ const updateTask = async (taskData, taskId = null) => {
  */
 const fetchTasks = async (limit, offset) => {
   let lastDoc;
-
-  // if offset given
   if (offset) {
     lastDoc = await tasksModel.doc(offset).get();
     lastDoc = lastDoc.data();
   }
-
-  // default case for limit
-  if (!limit) {
-    limit = 20;
-  }
-  limit = Number(limit);
+  let limitDocuments = Number(limit);
 
   try {
-    const tasksSnapshot = await tasksModel
-      .orderBy("startedOn", "desc")
-      .startAfter(lastDoc ? lastDoc.startedOn : "")
-      .limit(limit)
-      .get();
+    const tasksSnapshotQuery = tasksModel.orderBy("startedOn", "desc").startAfter(lastDoc ? lastDoc.startedOn : "");
+    const tasksSnapshot = limit ? await tasksSnapshotQuery.limit(limitDocuments).get() : await tasksSnapshotQuery.get();
     const tasks = buildTasks(tasksSnapshot);
     const promises = tasks.map(async (task) => fromFirestoreData(task));
     const updatedTasks = await Promise.all(promises);
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -47,7 +47,7 @@ const updateTask = async (taskData, taskId = null) => {
 const fetchTasks = async (limit, offset) => {
   let lastDoc;
 
-  //if offset given
+  // if offset given
   if (offset) {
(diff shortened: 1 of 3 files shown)


models/tasks.js
```diff
@@ -44,9 +44,27 @@ const updateTask = async (taskData, taskId = null) => {
  *
  * @return {Promise<tasks|Array>}
  */
-const fetchTasks = async () => {
+const fetchTasks = async (limit, offset) => {
+  let lastDoc;
+
+  //if offset given
```
ankushdharkar: This comment serves no purpose here. Please remove.

models/tasks.js
```diff
@@ -44,9 +44,27 @@ const updateTask = async (taskData, taskId = null) => {
...
+  if (offset) {
+    lastDoc = await tasksModel.doc(offset).get();
+    lastDoc = lastDoc.data();
+  }
+
+  // default case for limit
```
ankushdharkar, line 56: This comment serves no purpose here. Please remove.

models/tasks.js
```diff
@@ -44,9 +44,27 @@ const updateTask = async (taskData, taskId = null) => {
  *
  * @return {Promise<tasks|Array>}
  */
-const fetchTasks = async () => {
+const fetchTasks = async (limit, offset) => {
+  let lastDoc;
+
+  // if offset given
```
ankushdharkar, line 50: This comment serves no purpose here. Please remove.

test/integration/tasks.test.js
```diff
@@ -138,6 +138,46 @@ describe("Tasks", function () {
...
+              expect(res).to.have.status(200);
+              expect(res.body).to.be.a("object");
+              expect(res.body.tasks).to.be.a("array");
+              const taskWithParticipants = res.body.tasks;
+              taskWithParticipants.forEach((task) => {
+                if (task.type === "group") {
+                  expect(task.participants).to.include(appOwner.username);
+                } else {
+                  expect(task.assignee).to.equal(appOwner.username);
+                }
```
ankushdharkar, line 174: I wonder if this responsibility to validate the correct type of task should ideally belong to a test that checks for types. Not to the test that checks for pagination.

test/integration/tasks.test.js
```diff
@@ -138,6 +138,46 @@ describe("Tasks", function () {
...
+              });
+              return chaiRequester.close();
+            });
+          return done();
+        });
+    });
```
ankushdharkar, line 180: Are we missing a test that checks for default limit?

test/integration/tasks.test.js
```diff
@@ -138,6 +138,46 @@ describe("Tasks", function () {
...
+                if (task.type === "group") {
+                  expect(task.participants).to.include(appOwner.username);
+                } else {
+                  expect(task.assignee).to.equal(appOwner.username);
+                }
+              });
+              return chaiRequester.close();
+            });
+          return done();
+        });
```
ankushdharkar, line 179: Which test checks for the case where no limit is passed, but available tasks are less than the default allowed? What if we encounter an exception that breaks the API?

models/tasks.js
```diff
@@ -44,9 +44,27 @@ const updateTask = async (taskData, taskId = null) => {
...
+  if (!limit) {
+    limit = 20;
+  }
+  limit = Number(limit);
+
   try {
-    const tasksSnapshot = await tasksModel.get();
+    const tasksSnapshot = await tasksModel
+      .orderBy("startedOn", "desc")
+      .startAfter(lastDoc ? lastDoc.startedOn : "")
```
ankushdharkar, line 65: If `lastDoc` is undefined (or null), does this call error out?

ankushdharkar (changes requested):
I feel we need to cover more test cases here.

models/tasks.js
ivinayakg: no

test/integration/tasks.test.js
ivinayakg: the first test for the "GET /tasks" is the representation that if the limit is not passed the API will not break, but adding to this, I have made changes to the API to better handle the case if there is no limit passed.

test/integration/tasks.test.js
ivinayakg: there is no default as such, if you don't pass in the limit parameter, the API will return all the tasks, at least that's how I implemented it so that even if someone is not updated about this API change and has implemented in a way to get all the tasks then their functioning won't break.
I can change it if you say so.

test/integration/tasks.test.js
ivinayakg: Noted

test/integration/tasks.test.js
ivinayakg: and the test are still working

models/tasks.js
```diff
@@ -44,9 +44,17 @@ const updateTask = async (taskData, taskId = null) => {
  *
  * @return {Promise<tasks|Array>}
  */
-const fetchTasks = async () => {
+const fetchTasks = async (limit, offset) => {
+  let lastDoc;
+  if (offset) {
+    lastDoc = await tasksModel.doc(offset).get();
```
ankushdharkar, line 50: Offset should be a number, but here, it seems to be used as a document_id.

models/tasks.js
```diff
@@ -44,9 +44,17 @@ const updateTask = async (taskData, taskId = null) => {
...
+    lastDoc = lastDoc.data();
+  }
+  const limitDocuments = Number(limit);
```
ankushdharkar, line 53: Can we request any number of documents from Firestore? Should there be an upper limit here?

ankushdharkar (changes requested):

prakashchoudhary07:
Closing this, as this feature is already implemented (https://github.com/Real-Dev-Squad/website-backend/blob/develop/controllers/tasks.js)

PR #649, opened 2022-08-02
Modifies /members to return only unarchived users (by default)

### LINKED ISSUE #601: Archived members data not working

samarpan1738:
* Added a new role archived in ROLES constant

* Added a query param includeArchived(default value false)

* By default only unarchived users are returned. We can use includeArchived = true flag to get all users

* Updated and added integration test for /members endpoint

* Updated swagger doc for /members endpoint

* Rename all occurrences of archivedMember to archived

Closes #601 

### DIFF
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -14,7 +14,7 @@ const ERROR_MESSAGE = "Something went wrong. Please try again or contact admin";
 
 const getMembers = async (req, res) => {
   try {
-    const allUsers = await members.fetchUsers();
+    const allUsers = await members.fetchUsers(req.query);
 
     return res.json({
       message: allUsers.length ? "Members returned successfully!" : "No member found",
--- a/models/members.js
+++ b/models/members.js
@@ -16,14 +16,13 @@ const fetchUsers = async (queryParams = {}) => {
     const snapshot = await userModel.get();
     const allMembers = [];
 
-    let { includeArchived } = queryParams;
-    if (includeArchived === "true") includeArchived = true;
-    else includeArchived = false;
+    let { showArchived } = queryParams;
+    showArchived = showArchived === "true";
 
     if (!snapshot.empty) {
       snapshot.forEach((doc) => {
         const memberData = doc.data();
-        if (!includeArchived && memberData?.roles && memberData.roles[ROLES.ARCHIVED] === true) return;
+        if (!showArchived && memberData?.roles && memberData.roles[ROLES.ARCHIVED] === true) return;
         const curatedMemberData = {
           id: doc.id,
           ...memberData,
--- a/models/members.js
+++ b/models/members.js
@@ -5,21 +5,25 @@
 
 const firestore = require("../utils/firestore");
 const userModel = firestore.collection("users");
-
+const { ROLES } = require("../constants/roles");
 /**
  * Fetches the data about our users
  * @return {Promise<userModel|Array>}
  */
 
-const fetchUsers = async () => {
+const fetchUsers = async (queryParams = {}) => {
   try {
     const snapshot = await userModel.get();
-
     const allMembers = [];
+
+    let { includeArchived } = queryParams;
+    if (includeArchived === "true") includeArchived = true;
+    else includeArchived = false;
 
     if (!snapshot.empty) {
       snapshot.forEach((doc) => {
         const memberData = doc.data();
+        if (!includeArchived && memberData?.roles && memberData.roles[ROLES.ARCHIVED] === true) return;
         const curatedMemberData = {
           id: doc.id,
           ...memberData,
@@ -158,8 +162,8 @@ const addArchiveRoleToMembers = async (userId) => {
   try {
     const userDoc = await userModel.doc(userId).get();
     const user = userDoc.data();
-    if (user?.roles?.archivedMember) return { isArchived: true };
-    const roles = { ...user.roles, archivedMember: true };
+    if (user?.roles && user.roles[ROLES.ARCHIVED]) return { isArchived: true };
+    const roles = { ...user.roles, [ROLES.ARCHIVED]: true };
     await userModel.doc(userId).update({
       roles,
     });
--- a/models/members.js
+++ b/models/members.js
@@ -17,7 +17,7 @@ const fetchUsers = async (queryParams = {}) => {
     const allMembers = [];
 
     const { showArchived } = queryParams;
-    let isArchived = showArchived === "true";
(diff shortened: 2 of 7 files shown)


controllers/members.js
```diff
@@ -14,7 +14,7 @@ const ERROR_MESSAGE = "Something went wrong. Please try again or contact admin";
 
 const getMembers = async (req, res) => {
   try {
-    const allUsers = await members.fetchUsers();
+    const allUsers = await members.fetchUsers(req.query);
```
ankushdharkar, line 17: Is this XSS-safe?

models/members.js
```diff
@@ -5,21 +5,25 @@
...
-const fetchUsers = async () => {
+const fetchUsers = async (queryParams = {}) => {
   try {
     const snapshot = await userModel.get();
-
     const allMembers = [];
 
+    let { includeArchived } = queryParams;
+    if (includeArchived === "true") includeArchived = true;
+    else includeArchived = false;
```
ankushdharkar, line 21:
```suggestion
    let { showArchived } = queryParams;
    cosnt includeArchived = showArchived === "true";
```

models/members.js
```diff
@@ -5,21 +5,25 @@
...
+
     if (!snapshot.empty) {
       snapshot.forEach((doc) => {
         const memberData = doc.data();
+        if (!includeArchived && memberData?.roles && memberData.roles[ROLES.ARCHIVED] === true) return;
```
ankushdharkar, line 26: Instead of fetching everything and then removing archived users, can we write a firestore query such that we only fetch non-archived users in the first place.
Imagine if there are 100 non-archived users and 10000 archived users. We would have paid for 10100 calls, and used only a small portion of it.
Every DB/firestore read costs us.

test/fixtures/user/user.js
```diff
@@ -112,7 +112,7 @@ module.exports = () => {
       status: "active",
       roles: {
         app_owner: true,
-        archivedMember: true,
+        archived: true,
```
ankushdharkar, line 115: Thank you!

ankushdharkar (changes requested):

models/members.js
samarpan1738: Even with index we can't query for unarchived users if roles or archived field is absent https://firebase.google.com/docs/firestore/query-data/queries#limitations

Also there's an index for map fields by default
https://firebase.google.com/docs/firestore/query-data/index-overview#automatic_indexing

models/members.js
samarpan1738: Change done

test/fixtures/user/user.js
samarpan1738: Welcome

controllers/members.js
samarpan1738: I think this is safe from XSS since the value of the query param won't be going in the response and further to the HTML output.
Also if there were some other query parameter also being used for this api endpoint e.g. private which we think can be misused by some user then we need to restrict the it's usage by setting some role based authorisation.

samarpan1738: Added validation middleware to validate the query params for the /members endpoint

models/members.js
samarpan1738: Resolves #651

alishvagadiya: What is the full form of DNE hear?

alishvagadiya: > What is the full form of DNE hear?

DNE - does not exist

test/fixtures/user/user.js
```diff
@@ -132,7 +132,7 @@ module.exports = () => {
       status: "active",
       roles: {
         member: true,
-        archivedMember: true,
+        archived: false,
```
alishvagadiya, line 135: Why is value also changed with a key name?

test/integration/members.test.js
```diff
@@ -22,6 +22,10 @@ const nonSuperUser = userData[0];
 const userDoesNotExists = userData[1];
 const userToBeArchived = userData[3];
 const userAlreadyArchived = userData[5];
+const userWithRolesObjectWithoutArchivedProperty = userData[0];
+const userWithoutRolesObject = userData[1];
+const userWithRolesObjectWithArchivedTrue = userData[5];
+const userWithRolesObjectWithArchivedFalse = userData[6];
```
alishvagadiya, line 28: short variable name from the long variable name.

alishvagadiya (changes requested):
Look into review comments for changes.

test/fixtures/user/user.js
samarpan1738: I wanted the value false for a test

models/members.js
```diff
@@ -5,21 +5,24 @@
...
 
-const fetchUsers = async () => {
+const fetchUsers = async (queryParams = {}) => {
   try {
     const snapshot = await userModel.get();
-
     const allMembers = [];
 
+    let { showArchived } = queryParams;
+    showArchived = showArchived === "true";
```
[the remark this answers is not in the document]
samarpan1738: Query params are always of type string. 
In validation middleware I have used joi.boolean() in non-strict mode which checks for "true" or "false" or true or false
In model showArchived will be a string

test/integration/members.test.js
samarpan1738: Sure. We can shorten them a bit. Done that in the later PR https://github.com/Real-Dev-Squad/website-backend/pull/654

test/integration/members.test.js
```diff
@@ -69,8 +105,30 @@ describe("Members", function () {
           expect(res.body).to.be.a("object");
           expect(res.body.message).to.equal("Members returned successfully!");
           expect(res.body.members).to.be.a("array");
-          expect(res.body.members[0].roles.member).to.eql(true);
+          expect(res.body.members.length).to.be.equal(unarchivedUsersGithubIds.length);
+          const memberGithubIds = res.body.members.map((member) => member.github_id);
+          expect(memberGithubIds).to.include.all.members(unarchivedUsersGithubIds);
+          expect(memberGithubIds).to.not.include.any.members(archivedUsersGithubIds);
```
ankushdharkar, line 111: 🧑‍🍳 Awesome!

ankushdharkar (approved):

Issue #455, opened 2021-12-19
[Members Skills] API to return members skills


sumitd94:
API Contract ticket - https://github.com/Real-Dev-Squad/website-api-contracts/issues/70

> Start this once the Contract is finalised

Mir-SA:
Yo!

PR #650, opened 2022-08-03
Adds routes for Skills API

Mir-SA:
Adds routes for
 - Awarding skill to user ` /skills/:username `
 - Fetching all skills ` /skills `

Adds tests for both routes

Closes https://github.com/Real-Dev-Squad/website-backend/issues/455

### DIFF
--- a/controllers/skills.js
+++ b/controllers/skills.js
@@ -1,0 +1,52 @@
+const skills = require("../models/skills");
+const logger = require("../utils/logger");
+
+/**
+ * Award a skill
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+async function awardSkill(req, res) {
+  let addedData;
+
+  try {
+    const { username } = req.params;
+    addedData = await skills.awardSkill(req.body, username);
+  } catch (error) {
+    logger.error("Error posting skill data: ", error);
+    return res.boom.notFound("User doesn't exist");
+  }
+
+  return res.json({
+    message: "Added data successfully!",
+    content: addedData,
+  });
+}
+
+/**
+ * Fetch all skills
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+async function fetchSkills(req, res) {
+  let allSkills;
+
+  try {
+    allSkills = await skills.fetchSkills();
+  } catch (error) {
+    logger.error("Error fetching skill data: ", error);
+    return res.boom.serverUnavailable("Something went wrong please contact admin");
+  }
+
+  return res.json({
+    message: "Skills returned successfully!",
+    skills: allSkills.length > 0 ? allSkills : [],
+  });
+}
+
+module.exports = {
+  fetchSkills,
+  awardSkill,
+};
--- a/models/skills.js
+++ b/models/skills.js
@@ -1,0 +1,54 @@
+const firestore = require("../utils/firestore");
+const skillsModel = firestore.collection("skills");
+const users = require("./users");
+
+/**
+ * Award a Skill
+ *
+ * @param skillData {Object}:  Skill data object to be stored in DB
+ * @param userName   String:   Username of user to fetch their details.
+ * @return {Promise<skill|Object>}
+ */
+async function awardSkill(skillData, userName) {
+  try {
+    const skillSnapshot = await skillsModel.add(skillData);
+    const userInfo = await users.fetchUser({ username: userName });
+
+    const addedData = await skillSnapshot.get();
+    const data = userInfo.user.skills?.length
+      ? { skills: [...userInfo.user.skills, addedData.id] }
+      : { skills: [addedData.id] };
+
+    await users.addOrUpdate(data, userInfo.user.id);
+    return addedData.data();
+  } catch (error) {
+    logger.error("Error posting skill data", error);
+    throw error;
+  }
+}
+
+/**
+ * Fetch all Skills
+ *
+ * @return {Promise<skills|Array>}
+ */
+async function fetchSkills() {
+  try {
+    const skillsSnapshot = await skillsModel.get();
+    const skills = skillsSnapshot.docs.map((skill) => {
+      return {
+        id: skill.id,
+        ...skill.data(),
+      };
+    });
+    return skills;
+  } catch (error) {
+    logger.error("error getting tasks", error);
+    throw error;
+  }
+}
+
+module.exports = {
+  fetchSkills,
+  awardSkill,
+};
--- a/routes/skills.js
+++ b/routes/skills.js
@@ -1,0 +1,9 @@
+const express = require("express");
+const router = express.Router();
+const skills = require("../controllers/skills");
+const authenticate = require("../middlewares/authenticate");
+
+router.post("/:username", authenticate, skills.awardSkill);
+router.get("/", skills.fetchSkills);
+
+module.exports = router;
--- a/test/integration/skills.test.js
+++ b/test/integration/skills.test.js
@@ -1,0 +1,93 @@
+const chai = require("chai");
+const { expect } = chai;
+const chaiHttp = require("chai-http");
+
+const app = require("../../server");
+const authService = require("../../services/authService");
+const addUser = require("../utils/addUser");
+const { fetchUser } = require("../../models/users");
+const skillsData = require("../fixtures/skills/skills")();
+const cleanDb = require("../utils/cleanDb");
+
+const config = require("config");
+const cookieName = config.get("userToken.cookieName");
+
+chai.use(chaiHttp);
+
+describe("Skills", function () {
+  let jwt;
+  let userName;
+
+  beforeEach(async function () {
+    const userId = await addUser();
+    jwt = authService.generateAuthToken({ userId });
+
+    const { user } = await fetchUser({ userId });
+    userName = user.username;
+  });
+
+  afterEach(async function () {
+    await cleanDb();
+  });
+  describe("POST /skills/:username", function () {
+    it("Should add data to skills model & return successful response", function (done) {
+      chai
+        .request(app)
+        .post(`/skills/${userName}`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send(skillsData[0])
+        .end((err, res) => {
+          if (err) return done(err);
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Added data successfully!");
+          expect(res.body.content).to.be.a("object");
+          expect(res.body.content.name).to.be.a("string");
+          expect(res.body.content.awarded_on).to.be.a("string");
+          expect(res.body.content.awarded_by).to.be.a("string");
+          expect(res.body.content.awarded_for).to.be.a("string");
+
+          return done();
+        });
+    });
+
+    it("Should return 401 for anunthenticated user", function (done) {
+      chai
+        .request(app)
+        .post(`/skills/${userName}`)
+        .send(skillsData[0])
+        .end((err, res) => {
+          if (err) return done(err);
+
+          expect(res).to.have.status(401);
+          expect(res.body).to.be.a("object");
+          expect(res.body).to.deep.equal({
+            statusCode: 401,
+            error: "Unauthorized",
+            message: "Unauthenticated User",
+          });
+
+          return done();
+        });
+    });
+  });
+
+  describe("GET /skills", function () {
+    it("Should get the list of skills", function (done) {
+      chai
+        .request(app)
+        .get("/skills")
+        .end((err, res) => {
+          if (err) {
+            return done();
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Skills returned successfully!");
+          expect(res.body.skills).to.be.a("array");
+
+          return done();
+        });
+    });
+  });
+});
--- a/routes/index.js
+++ b/routes/index.js
@@ -12,6 +12,7 @@ app.use("/logs", require("./logs.js"));
 app.use("/members", require("./members.js"));
 app.use("/pullrequests", require("./pullrequests.js"));
 app.use("/stocks", require("./stocks.js"));
+app.use("/skills", require("./skills.js"));
 app.use("/tasks", require("./tasks.js"));
 app.use("/trade", require("./trading.js"));
 app.use("/users", require("./users.js"));
--- a/test/fixtures/skills/skills.js
+++ b/test/fixtures/skills/skills.js
@@ -1,0 +1,27 @@
+/**
+ * Dummy data for tests
+ * @return  {object}
+ */
+
+module.exports = () => {
+  return [
+    {
+      name: "node",
+      awarded_on: "10-05-2022",
+      awarded_by: "ankur",
+      awarded_for: "writing skills api",
+    },
+    {
+      name: "react",
+      awarded_on: "10-05-2022",
+      awarded_by: "sumit",
+      awarded_for: "fixing broken social links",
+    },
+    {
+      name: "js",
+      awarded_on: "10-05-2022",
+      awarded_by: "ankush",
+      awarded_for: "building lift sim",
+    },
+  ];
+};


Mir-SA:
@ankushdharkar can you review it?

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+ * @param req {Object} - Express request object
+ * @param req.body {Object} - User object
+ * @param res {Object} - Express response object
+ */
+async function awardSkill(req, res) {
+  let addedData;
+
+  try {
+    const { username } = req.params;
+    addedData = await skills.awardSkill(req.body, username);
```
rohan09-raj, line 16: Can we please use `const` here ?

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+  } catch (error) {
+    logger.error("Error posting skill data: ", error);
+    return res.boom.notFound("User doesn't exist");
+  }
+
+  return res.json({
+    message: "Added data successfully!",
+    content: addedData,
+  });
```
rohan09-raj, line 25: Why it is outside the try block ?

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+ * Fetch all skills
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+async function fetchSkills(req, res) {
+  let allSkills;
+
+  try {
+    allSkills = await skills.fetchSkills();
```
rohan09-raj, line 38: Can we please use `const` here ?

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+  } catch (error) {
+    logger.error("Error fetching skill data: ", error);
+    return res.boom.serverUnavailable("Something went wrong please contact admin");
+  }
+
+  return res.json({
+    message: "Skills returned successfully!",
+    skills: allSkills.length > 0 ? allSkills : [],
+  });
```
rohan09-raj, line 47: Why this return is outside the try block ?

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+ * Fetch users with given skill
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+async function userWithSkill(req, res) {
+  let { skill } = req.params;
+  // In case user makes query with - or _ for multiple words
+  const regex = /[-_]/g;
+  // replace those chars with whitespace
```
rohan09-raj, line 81: This comment is not needed here, code itself is self explanatory

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+  skill = regex.test(skill) ? skill.replace(regex, " ") : skill;
```
rohan09-raj, line 82: Instead of modifying the actual value, can we have two separate constants ?

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Fetch users with given skill
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+async function userWithSkill(req, res) {
+  let { skill } = req.params;
+  // In case user makes query with - or _ for multiple words
+  const regex = /[-_]/g;
```
rohan09-raj, line 80: Can we please import this from a constant file ?

controllers/skills.js
```diff
@@ -0,0 +1,105 @@
...
+  // replace those chars with whitespace
+  skill = regex.test(skill) ? skill.replace(regex, " ") : skill;
+  let filteredData;
+
+  try {
+    filteredData = await skills.userWithSkill(skill);
```
rohan09-raj, line 86: Can we please use `const` here ?

routes/skills.js
```diff
@@ -0,0 +1,11 @@
+const express = require("express");
+const router = express.Router();
+const skills = require("../controllers/skills");
+const authenticate = require("../middlewares/authenticate");
+
+router.post("/:username", authenticate, skills.awardSkill);
+router.get("/", skills.fetchSkills);
+router.get("/:username", authenticate, skills.fetchUserSkills);
+router.get("/users/:skill", authenticate, skills.userWithSkill);
```
rohan09-raj, line 9: This should be in the `users` not `skills`

models/skills.js
```diff
@@ -0,0 +1,102 @@
...
+    filteredData = newData.filter((user) => {
+      return user.skills.includes(skill.toLowerCase());
+    });
+    if (!filteredData.length) throw Error();
+  } catch (error) {
+    logger.error("Error getting user with given skill", error);
+  }
+
+  return filteredData;
+}
```
rohan09-raj, line 95: - Can we please use `const` in here for variables ?
- This model function with return an array of users right ?? So can we please correct the naming ?
- This should be in `/models/users.js`

controllers/skills.js
Mir-SA: I had some lint errors for no-return-value that's why I moved those return statements outside block with let scope.
I'll make the change

controllers/skills.js
Mir-SA: just the regex? why?
I mean it's only a single line of code for a specific case...

routes/skills.js
Mir-SA: I was trying `/skills/:skill` but it didn't seem to work hence I wrote it this way...
But your point is right. It would be more obvious if routed through `/users`

Mir-SA: @rohan09-raj  
is it ok if that `/users/:skill` is `/users/skills/:skill`?
`/users/:skill` gets overlapped by `/users/:username` and tries to return user with `/:skill` as username

rohan09-raj: Yeah I missed that, thanks for pointing it out. `/users/skills/:skill` is better

models/skills.js
```diff
@@ -0,0 +1,72 @@
+const firestore = require("../utils/firestore");
+const skillsModel = firestore.collection("skills");
```
samarpan1738, line 2: This variable can be named better
```suggestion
const skillsCollection = firestore.collection("skills");
```

controllers/skills.js
```diff
@@ -0,0 +1,70 @@
...
+ * Award a skill
+ *
+ * @param req {Object} - Express request object
+ * @param req.body {Object} - User object
+ * @param res {Object} - Express response object
+ */
+async function awardSkill(req, res) {
+  try {
+    const { username } = req.params;
+    const addedData = await skills.awardSkill(req.body, username);
```
samarpan1738, line 14: Add validation for data being passed to model

models/skills.js
```diff
@@ -0,0 +1,72 @@
...
+ * Award a Skill
+ *
+ * @param skillData { Object }:  Skill data object to be stored in DB
+ * @param userName  { String }:   Username of user to fetch their details.
+ * @return {Promise<skill|Object>}
+ */
+async function awardSkill(skillData, userName) {
+  try {
+    const skillSnapshot = await skillsModel.add(skillData);
+    const userInfo = await users.fetchUser({ username: userName });
```
samarpan1738, line 15: What happens when the user is not found?

controllers/skills.js
```diff
@@ -0,0 +1,70 @@
...
+    return res.json({
+      message: "Added data successfully!",
```
samarpan1738, line 16: Can we return a better response message?

controllers/skills.js
```diff
@@ -0,0 +1,70 @@
...
+      content: addedData,
+    });
+  } catch (error) {
+    logger.error("Error posting skill data: ", error);
+    return res.boom.notFound("User doesn't exist");
```
samarpan1738, line 21: If there is any error we'll be sending a 404 response
Is this the intended behaviour?

controllers/skills.js
```diff
@@ -0,0 +1,70 @@
...
+ * Fetch skills of User
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+async function fetchUserSkills(req, res) {
+  const { username } = req.params;
+
+  try {
+    const userData = await skills.fetchUserSkills(username);
```
samarpan1738, line 55: Sanitise and validate parameters

test/integration/skills.test.js
```diff
@@ -0,0 +1,93 @@
...
+          expect(res.body.content.name).to.be.a("string");
+          expect(res.body.content.awarded_on).to.be.a("string");
+          expect(res.body.content.awarded_by).to.be.a("string");
+          expect(res.body.content.awarded_for).to.be.a("string");
+
+          return done();
+        });
+    });
+
+    it("Should return 401 for anunthenticated user", function (done) {
```
samarpan1738, line 54: Typo `anunthenticated`

test/integration/skills.test.js
```diff
@@ -0,0 +1,93 @@
...
+    jwt = authService.generateAuthToken({ userId });
+
+    const { user } = await fetchUser({ userId });
+    userName = user.username;
+  });
+
+  afterEach(async function () {
+    await cleanDb();
+  });
+  describe("POST /skills/:username", function () {
```
samarpan1738, line 32: More cases need to be covered here like when username is not found, when there's no user found with the username,etc

samarpan1738 (changes requested):

models/skills.js
Mir-SA: There's a 404 response for that with an error and a message 
For e.g.
`
{ 
   "statusCode": 404, 
   "error": "Not Found", 
   "message": "User doesn't exist" 
}
`

controllers/skills.js
Mir-SA: as in more descriptive? or something else entirely?

controllers/skills.js
Mir-SA: yes

controllers/skills.js
Mir-SA: validation for?

controllers/skills.js
samarpan1738: Can there be any other cause of error other than `user not found`. Even in that case should be return 404 or 400 since 404 is meant when the requested resource is not found. Here we are not requesting for a resource

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404

Mir-SA: 1. I think there can be an error response for no data(empty object) in req.body as well
2. 400 status code would be better since there is no req for resource...

controllers/skills.js
samarpan1738: A message that should be based on the acknowledgment of the action this api performs
Currently it's very generic `Added data successfully!`. It's ambiguous that what data is actually added

test/integration/skills.test.js
```diff
@@ -0,0 +1,153 @@
...
+    it("Should return 400 if by field is not specified", function (done) {
+      chai
+        .request(app)
+        .post(`/skills/${userName}`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send(skillsData[3])
+        .end((err, res) => {
+          if (err) return done(err);
+
+          expect(res).to.have.status(400);
```
ankushdharkar, line 124: It's hard to tell why this test fails.

It would have been nice if you do something like this:

const skillData  = skillsData[3];
delete skillData.by;  -- creates an object with the missing field(s)

ankushdharkar (changes requested):

test/integration/skills.test.js
```diff
@@ -0,0 +1,153 @@
...
+        });
+    });
+
+    it("Should return 400 if name field is not specified", function (done) {
+      chai
+        .request(app)
+        .post(`/skills/${userName}`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send(skillsData[2])
+        .end((err, res) => {
```
ankushdharkar, line 100: Please refer [this](https://github.com/Real-Dev-Squad/website-backend/pull/650#discussion_r953236334)

When reading the test, I have no insight that `skills[2]` is a particular type of incomplete object.

controllers/skills.js
```diff
@@ -0,0 +1,70 @@
...
+  try {
+    const { username } = req.params;
+    const addedData = await skills.awardSkill(req.body, username);
+    return res.json({
+      message: "Added skill for user successfully!",
+      skill: addedData,
+    });
+  } catch (error) {
+    logger.error("Error posting skill data: ", error);
+    return res.boom.badRequest(error.message);
```
prakashchoudhary07, line 21: Won't this return 400 for every type of error, that has been thrown?

controllers/users.js
```diff
@@ -292,6 +293,33 @@ const rejectProfileDiff = async (req, res) => {
...
+ * Fetch users with given skill
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+async function userWithSkill(req, res) {
+  const { skillName } = req.params;
+  const skillWithoutSpecialChar = checkDashUnderscore.test(skillName)
+    ? skillName.replace(checkDashUnderscore, " ")
+    : skillName;
```
prakashchoudhary07, line 306: If there is some error here, how is that handled?

models/skills.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @param skillData { Object }:  Skill data object to be stored in DB
+ * @param userName  { String }:   Username of user to fetch their details.
+ * @return {Promise<skill|Object>}
+ */
+async function awardSkill(skillData, userName) {
+  try {
+    const skillSnapshot = await skillsCollection.add(skillData);
+    const userInfo = await users.fetchUser({ username: userName });
+    if (!userInfo.userExists) throw new Error("User doesn't exist");
```
prakashchoudhary07, line 16: Shouldn't the user be fetched before any skills are created?
What are your thoughts on it?

models/skills.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Award a Skill
+ *
+ * @param skillData { Object }:  Skill data object to be stored in DB
+ * @param userName  { String }:   Username of user to fetch their details.
+ * @return {Promise<skill|Object>}
+ */
+async function awardSkill(skillData, userName) {
+  try {
+    const skillSnapshot = await skillsCollection.add(skillData);
```
prakashchoudhary07, line 14: Won't this create multiple records for the same skills?

models/users.js
```diff
@@ -186,6 +186,36 @@ const fetchUserImage = async (users) => {
...
+ *
+ * Fetch Users with given Skill
+ *
+ * @param skill { String }: Get users with given skill
+ * @return {Promise<skills|Array>}
+ */
+async function userWithSkill(skill) {
+  try {
+    const userData = await userModel.orderBy("skills", "asc").get();
+    const userSkills = userData.docs.map((doc) => doc.data());
```
prakashchoudhary07, line 200: userSkills has not only user skills data but also users' other data,
What are your thoughts on renaming the variable?

prakashchoudhary07 (changes requested):

middlewares/validators/skills.js
```diff
@@ -0,0 +1,20 @@
...
+    by: joi.string().required(),
+    on: joi.string().optional(),
+    for: joi.string().optional(),
+  });
+
+  try {
+    await schema.validateAsync(req.body);
+    next();
+  } catch (error) {
+    logger.error("Error posting skill data: ", error);
```
poonammohata, line 15: I think in the log, instead of this message "Error posting..", "Error in validating...", would make more sense. What do you think? and in all the other validators in this module, template literal is used in the logger error message, So it would be better that we use template literals in this message also, to maintain consistency

controllers/skills.js
```diff
@@ -0,0 +1,70 @@
...
+  try {
+    const allSkills = await skills.fetchSkills();
+
+    return res.json({
+      message: "Skills returned successfully!",
+      skills: allSkills.length > 0 ? allSkills : [],
+    });
+  } catch (error) {
+    logger.error("Error fetching skill data: ", error);
+    return res.boom.serverUnavailable("Something went wrong please contact admin");
```
poonammohata, line 41: Shouldn't this be 500 (server) error i.e badImplementation ?

controllers/skills.js
```diff
@@ -0,0 +1,70 @@
...
+
+  try {
+    const userData = await skills.fetchUserSkills(username);
+    return res.json({
+      message: "Skills returned successfully!",
+      skills: userData.length > 0 ? userData : [],
+    });
+  } catch (error) {
+    logger.error("Error fetching user skills: ", error);
+    return res.boom.notFound(error.message);
```
poonammohata, line 62: Shouldn't this be 500 (server) error i.e badImplementation ?

controllers/users.js
```diff
@@ -292,6 +293,33 @@ const rejectProfileDiff = async (req, res) => {
...
+
+    if (filteredData) {
+      return res.json({
+        message: "Users returned successfully!",
+        data: filteredData,
+      });
+    } else throw new Error("No users with this skill");
+  } catch (error) {
+    logger.error("Error fetching user with skill: ", error);
+    return res.boom.badRequest("Invalid Skill. Please re-check input data");
```
poonammohata, line 319: Shouldn't this be 500 (server) error i.e badImplementation?

prakashchoudhary07:
Closing this as this feature is being built as a stand-alone product, in Skill-tree

https://github.com/Real-Dev-Squad/skill-tree-frontend/
https://github.com/Real-Dev-Squad/skill-tree-backend/

PR #654, opened 2022-08-05
Adds an endpoint to add default archived role only where it does not exists

### LINKED ISSUE #651: Create a new endpoint to set default archived role for users where it DNE
Reason : To query on a field in firestore, the field should exist in the document. Discussed in detail here : https://github.com/Real-Dev-Squad/website-backend/pull/649#discussion_r935730449

1. Endpoint : /users/add-default-archived-role
2. Allowed only for superuser 
3. Add roles object for users where it DNE 
4. Add roles.archived = false for users where roles.archived DNE
5. This is a one-time-use endpoint and should be deleted after use - https://github.com/Real-Dev-Squad/website-backend/issues/671

samarpan1738:
## This PR can only be merged after #661
- #661 

* Endpoint : `PATCH /users/add-default-archived-role`
* Added model,route,controller methods. Default archived = `false` role will be added to all users where either roles DNE or role.archived DNE
* Added integration test for this endpoint
* Load tested this endpoint with `5000` users (all without archived property in roles map). 
Time taken ≈ `15-20 sec` . 
Breakdown : 5000 reads(≈4s) + 5000 updates(≈12s) 

Closes #651 

### DIFF
--- a/test/integration/users.test.js
+++ b/test/integration/users.test.js
@@ -13,6 +13,10 @@ const checkChaincode = require("../utils/checkChaincode");
 const userData = require("../fixtures/user/user")();
 const profileDiffData = require("../fixtures/profileDiffs/profileDiffs")();
 const superUser = userData[4];
+const nonSuperUser = userData[0];
+const userRolesDoesNotExists = userData[1];
+const userArchivedRoleDoesNotExists = userData[0];
+const userAlreadyArchived = userData[5];
 
 const config = require("config");
 const cookieName = config.get("userToken.cookieName");
@@ -535,6 +539,7 @@ describe("Users", function () {
         });
     });
   });
+
   describe("POST /users/verify", function () {
     it("Should queue the Request", function (done) {
       chai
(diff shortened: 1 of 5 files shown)


controllers/auth.js
```diff
@@ -29,6 +29,9 @@ const githubAuth = (req, res, next) => {
         tokens: {
           githubAccessToken: accessToken,
         },
+        roles: {
+          archived: false,
+        },
```
rohan09-raj, line 34: In my opinion, we can put this `roles` object somewhere else. Not sure where, but yes I don't think this is the appropriate place to put this `roles` object.
Can you find some other place where we can add this object to the user document ??

samarpan1738: Should I add it here ? Inside addOrUpdate model function
https://github.com/Real-Dev-Squad/website-backend/blob/15e6242972b7f674af869b618ed42a1c5ae080fd/models/users.js#L45

routes/members.js
```diff
@@ -284,4 +291,39 @@ router.patch("/delete-isMember", authenticate, authorizeUser("superUser"), membe
...
+ *           schema:
+ *             $ref: '#/components/schemas/errors/forbidden'
+ *     500:
+ *       description: badImplementation
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/components/schemas/errors/badImplementation'
+ */
+router.patch("/add-default-archived-role", authenticate, authorizeUser("superUser"), members.addDefaultArchivedRole);
```
rohan09-raj, line 327: The naming of API can be better, this is too long to have
These can help you 
https://www.thoughtworks.com/en-us/insights/blog/rest-api-design-resource-modeling
https://aws.amazon.com/what-is/restful-api/

controllers/auth.js
rohan09-raj: Yes, I think we can put it here

routes/members.js
samarpan1738: Ok. I'll give them a read. 
The points I considered for naming this api were :
1. It's intended to be used just once
2. Similar naming convention has been followed for other migration endpoints like delete-isMember and member-to-role-migration

rohan09-raj: Ok, then I guess it's fine, thanks for clarifying

controllers/auth.js
samarpan1738: If we move this logic in users model's addOrUpdateUser method it'll break test cases where the fixtures already have a roles object

samarpan1738: As discussed with @ankushdharkar , a user should not have roles when it's added in firestore for the first time. For breaking tests we need to fix the fixtures and setup part of those tests

test/integration/users.test.js
```diff
@@ -567,4 +572,63 @@ describe("Users", function () {
...
+          expect(res.body.message).to.equal("You are not authorized for this action.");
+
+          return done();
+        });
+    });
+
+    it("Should add default archived role to user only where it does not exists", function (done) {
+      chai
+        .request(app)
+        .patch(`/users/add-default-archived-role`)
```
ankushdharkar, line 607: Can we have a better API name here please.

API names should be nouns, not verbs.

ankushdharkar (changes requested):

ankushdharkar (approved):

test/integration/users.test.js
samarpan1738: Yes. This was a miss on my part. Will think proper naming from next time onwards.

samarpan1738:
`PATCH /users/add-default-archived-role` had been executed by @ankushdharkar on 20th August 2022
So now all the existing unarchived users have archived = false in their roles object

PR #656, opened 2022-08-05
Uses firestore query to fetch only unarchived users

### LINKED ISSUE #652: Use firestore query to fetch only unarchived users for /members endpoint
Currently in #649 we are fetching all users from firestore and then filtering the unarchived users at application level
So in cases where we have 1000 archived users and 10 unarchived users we'll be charged for 1010 reads but we only needed 10 of them. 
To improve this we can write a firestore query to fetch only the unarchived users . 
But for this to work we'll need the archived role be present in all documents otherwise that document will be skipped([why?](https://firebase.google.com/docs/firestore/query-data/queries#limitations))

This will work only after executing `PATCH /users/add-default-archived-role` that'll be added by
- #651

samarpan1738:
> **Warning**
> This PR can only be merged after #662 and after `PATCH /users/add-default-archived-role` has been executed
- #662

* Updated members model with firestore query to fetch unarchived users

Closes #652

Please follow the merge order specified here : https://github.com/Real-Dev-Squad/website-backend/issues/601#issuecomment-1219781528

### DIFF
--- a/models/members.js
+++ b/models/members.js
@@ -13,16 +13,17 @@ const { ROLES } = require("../constants/roles");
 
 const fetchUsers = async (queryParams = {}) => {
   try {
-    const snapshot = await userModel.get();
-    const allMembers = [];
-
     const { showArchived } = queryParams;
-    const isArchived = showArchived === "true";
+    const shouldShowArchived = showArchived === "true";
+
+    const query = shouldShowArchived ? userModel : userModel.where(`roles.${ROLES.ARCHIVED}`, "==", false);
+    const snapshot = await query.get();
+
+    const allMembers = [];
 
     if (!snapshot.empty) {
       snapshot.forEach((doc) => {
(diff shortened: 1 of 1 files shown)


vinayak-trivedi (approved):
LGTM

models/members.js
```diff
@@ -13,16 +13,17 @@ const { ROLES } = require("../constants/roles");
 
 const fetchUsers = async (queryParams = {}) => {
   try {
-    const snapshot = await userModel.get();
-    const allMembers = [];
-
     const { showArchived } = queryParams;
-    const isArchived = showArchived === "true";
+    const shouldShowArchived = showArchived === "true";
```
rohan09-raj, line 17: Why is boolean a string ?

ankushdharkar (commented):
Thanks for these changes!

PR #661, opened 2022-08-07
Implements expected behaviour while adding user in test util

### LINKED ISSUE #659: Fix user fixtures and test setup
**Problem**
In real life scenario when a user is first added in the DB they do not have any roles.

But the [fixtures](https://github.com/Real-Dev-Squad/website-backend/blob/develop/test/fixtures/user/user.js) being used to add a new user (during the tests setup) has the roles property pre-populated and we end up with a scenario where a new user added in the DB already has roles. 

Hence we need to fix the fixtures and modify the test setup part where the new user added in DB must not have any roles. 
If we want to add roles it must be done in the same way as it'll be done in a real life scenario

Basically the tests should mimic the business flow and use cases 

**How we found this issue?**
Read this discussion : https://github.com/Real-Dev-Squad/website-backend/pull/654#discussion_r939069800

samarpan1738:
## This PR can only be merged after #649 

- #649 

* Add user in firestore without roles

* Update user's roles after adding the user in firestore

Closes #659 

### DIFF
--- a/test/utils/addUser.js
+++ b/test/utils/addUser.js
@@ -12,7 +12,21 @@ module.exports = async (user) => {
   const isValid = user && Object.keys(user).length !== 0 && user.constructor === Object;
   // Use the user data sent as arguments, else use data from fixtures
   user = isValid ? user : userData[0];
-  const { userId } = await users.addOrUpdate(user);
+  /*
+    Making a shallow copy of user allows us to safely modify the user object
+    without affecting the original fixture
+  */
+  const userWithoutRoles = { ...user };
+  const rolesToBeAdded = userWithoutRoles.roles;
+  // A new user to be added in the DB should not have any roles
+  delete userWithoutRoles.roles;
+  const { userId } = await users.addOrUpdate(userWithoutRoles);
 
+  if (rolesToBeAdded) {
+    const persistedUser = (await users.fetchUser({ userId })).user;
+    const existingRoles = persistedUser.roles;
+    const rolesToBeUpdated = { ...existingRoles, ...rolesToBeAdded };
+    await users.addOrUpdate({ roles: rolesToBeUpdated }, userId);
+  }
   return userId;
 };


test/utils/addUser.js
```diff
@@ -12,7 +12,19 @@ module.exports = async (user) => {
   const isValid = user && Object.keys(user).length !== 0 && user.constructor === Object;
   // Use the user data sent as arguments, else use data from fixtures
   user = isValid ? user : userData[0];
+  /*
+    Making a shallow copy of user allows us to safely modify the user object
+    without affecting the original fixture
+  */
+  user = { ...user };
```
ankushdharkar, line 19: Can we use a new variable here please? It's bad practice to reuse param from function params list.

test/utils/addUser.js
```diff
@@ -12,7 +12,19 @@ module.exports = async (user) => {
...
+  const roles = user.roles;
```
ankushdharkar, line 20: Can this be named better?

test/utils/addUser.js
```diff
@@ -12,7 +12,19 @@ module.exports = async (user) => {
...
+  delete user.roles;
```
ankushdharkar, line 21: Please add a comment here of why we are doing this.

test/utils/addUser.js
```diff
@@ -12,7 +12,19 @@ module.exports = async (user) => {
...
   const { userId } = await users.addOrUpdate(user);
-
+  if (roles) {
+    const persistedUser = (await users.fetchUser({ userId })).user;
+    const currentRoles = persistedUser.roles;
+    const updatedRoles = currentRoles ? { ...currentRoles, ...roles } : roles;
```
ankushdharkar, line 26: Will this work if `currentRoles` is an empty object?

ankushdharkar (changes requested):

test/utils/addUser.js
samarpan1738: Surely. Changes done

test/utils/addUser.js
samarpan1738: Done

test/utils/addUser.js
samarpan1738: Yes. It'll work. Removed the ternary

test/utils/addUser.js
samarpan1738: Done.

ankushdharkar (approved):

PR #662, opened 2022-08-07
Adds default archived role for new users

### LINKED ISSUE #660: Add default archived role for new users
Currently a new user does not have any roles to start with. But for #652 to work we'll require all users to have the archived role = false by default
Changes need to be made such that new users have archived role = false by default

samarpan1738:
## This PR can only be merged after #681 
- #681 

* Updated addOrUpdate model function

* Added deleteRoles and deleteRolesObject test util to delete all/specified roles

* Fixed integration tests /users/add-default-archived-role using delete utils

* Updated member integration tests since now we have only 1 kind of unarchived user : the user with archived = false

Closes #660 

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -41,7 +41,13 @@ const addOrUpdate = async (userData, userId = null) => {
       return { isNewUser: false, userId: user.docs[0].id };
     }
 
-    // Add user
+    // Add new user
+    /*
+      Adding default archived role enables us to query for only
+      the unarchived users in the /members endpoint
+      For more info : https://github.com/Real-Dev-Squad/website-backend/issues/651
+    */
+    userData.roles = { archived: false };
     userData.incompleteUserDetails = true;
     const userInfo = await userModel.add(userData);
     return { isNewUser: true, userId: userInfo.id };
--- a/test/integration/members.test.js
+++ b/test/integration/members.test.js
@@ -165,10 +165,6 @@ describe("Members", function () {
       jwt = authService.generateAuthToken({ userId: superUserId });
     });
 
-    afterEach(async function () {
-      await addUser();
-    });
-
     it("Should return 404 if user doesn't exist", function (done) {
       chai
         .request(app)
@@ -253,10 +249,6 @@ describe("Members", function () {
     beforeEach(async function () {
       const superUserId = await addUser(superUser);
       jwt = authService.generateAuthToken({ userId: superUserId });
-    });
-
-    afterEach(async function () {
-      await addUser();
     });
 
     it("Should return 404 if user doesn't exist", function (done) {
--- a/test/integration/members.test.js
+++ b/test/integration/members.test.js
@@ -22,17 +22,10 @@ const nonSuperUser = userData[0];
 const userDoesNotExists = userData[1];
 const userToBeArchived = userData[3];
 const userAlreadyArchived = userData[5];
-const userWithRolesObjectWithoutArchivedProperty = userData[0];
-const userWithoutRolesObject = userData[1];
-const userWithRolesObjectWithArchivedTrue = userData[5];
-const userWithRolesObjectWithArchivedFalse = userData[6];
+const userArchivedRoleFalse = userData[6];
 
-const archivedUsersGithubIds = [userWithRolesObjectWithArchivedTrue.github_id];
-const unarchivedUsersGithubIds = [
-  userWithoutRolesObject.github_id,
-  userWithRolesObjectWithoutArchivedProperty.github_id,
-  userWithRolesObjectWithArchivedFalse.github_id,
-];
+const archivedUsersGithubIds = [userAlreadyArchived.github_id];
+const unarchivedUsersGithubIds = [userArchivedRoleFalse.github_id];
 
 describe("Members", function () {
   let jwt;
@@ -79,12 +72,7 @@ describe("Members", function () {
 
     describe("When the users collection is not empty", function () {
       beforeEach(async function () {
-        await Promise.all([
-          addUser(userWithoutRolesObject),
-          addUser(userWithRolesObjectWithoutArchivedProperty),
-          addUser(userWithRolesObjectWithArchivedTrue),
-          addUser(userWithRolesObjectWithArchivedFalse),
-        ]);
+        await Promise.all([addUser(userArchivedRoleFalse), addUser(userAlreadyArchived)]);
       });
 
       it("Should return all the unarchived users in the database", function (done) {
@@ -177,6 +165,10 @@ describe("Members", function () {
       jwt = authService.generateAuthToken({ userId: superUserId });
     });
 
+    afterEach(async function () {
+      await addUser();
+    });
+
     it("Should return 404 if user doesn't exist", function (done) {
       chai
         .request(app)
@@ -261,6 +253,10 @@ describe("Members", function () {
     beforeEach(async function () {
       const superUserId = await addUser(superUser);
       jwt = authService.generateAuthToken({ userId: superUserId });
+    });
+
+    afterEach(async function () {
+      await addUser();
     });
 
     it("Should return 404 if user doesn't exist", function (done) {
--- a/test/utils/deleteRoles.js
+++ b/test/utils/deleteRoles.js
@@ -1,0 +1,25 @@
+const firestore = require("../../utils/firestore");
+const userCollection = firestore.collection("users");
+
+/**
+ * @return {boolean} success - are roles deleted or not
+ */
+module.exports = async (userId = null, rolesToBeDeleted = [], deleteRolesObject = false) => {
+  if (!userId) return false;
+
+  try {
+    const userDoc = await userCollection.doc(userId).get();
+
+    if (!userDoc.exists) return false;
+
+    const userData = userDoc.data();
+    if (deleteRolesObject) delete userData.roles;
+    else rolesToBeDeleted.forEach((role) => delete userData.roles[String(role)]);
+
+    await userCollection.doc(userId).set(userData);
+
+    return true;
+  } catch (error) {
+    return false;
+  }
+};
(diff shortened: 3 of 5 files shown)


routes/members.js
```diff
@@ -284,4 +291,39 @@ router.patch("/delete-isMember", authenticate, authorizeUser("superUser"), membe
...
+ *           schema:
+ *             $ref: '#/components/schemas/errors/forbidden'
+ *     500:
+ *       description: badImplementation
+ *       content:
+ *         application/json:
+ *           schema:
+ *             $ref: '#/components/schemas/errors/badImplementation'
+ */
+router.patch("/add-default-archived-role", authenticate, authorizeUser("superUser"), members.addDefaultArchivedRole);
```
ankushdharkar, line 327: Nit: Naming can be improved. But since this is a temporary API, we can let this one slide in.

Do you have a cleanup ticket ready for it?

ankushdharkar, line 327: We should use the SUPER_USER constant here. Not a string "superUser"

models/users.js
```diff
@@ -42,6 +42,7 @@ const addOrUpdate = async (userData, userId = null) => {
     }
 
     // Add user
+    userData.roles = { archived: false };
```
ankushdharkar, line 45: A small comment why would be nice here.

controllers/members.js
```diff
@@ -139,11 +139,32 @@ const archiveMembers = async (req, res) => {
...
+/**
+ * Returns the lists of usernames where default archived role was added
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const addDefaultArchivedRole = async (req, res) => {
+  try {
+    const addedDefaultArchivedRoleData = await members.addDefaultArchivedRole();
```
ankushdharkar, line 151: Should we be worried about potential timeout here. What happens if we have a million members?
Does this apply to members or to users?

middlewares/validators/members.js
```diff
@@ -0,0 +1,19 @@
+const joi = require("joi");
+
+const validateGetMembers = async (req, res, next) => {
```
ankushdharkar, line 3: Does this only validate a 'get'?

models/members.js
```diff
@@ -5,21 +5,24 @@
...
 
-const fetchUsers = async () => {
+const fetchUsers = async (queryParams = {}) => {
   try {
     const snapshot = await userModel.get();
-
     const allMembers = [];
 
+    let { showArchived } = queryParams;
+    showArchived = showArchived === "true";
```
ankushdharkar, line 20: We are converting a string variable to boolean here? Can we use a new variable `isArchivedShown`?

ankushdharkar: Automatic types conversion is how we have bugs later.

models/members.js
```diff
@@ -5,21 +5,24 @@
...
+const { ROLES } = require("../constants/roles");
 /**
  * Fetches the data about our users
  * @return {Promise<userModel|Array>}
  */
 
-const fetchUsers = async () => {
+const fetchUsers = async (queryParams = {}) => {
   try {
     const snapshot = await userModel.get();
```
ankushdharkar, line 16: This doesn't do what we discussed, about fetching only the users who have `archived != true`

models/members.js
```diff
@@ -170,11 +173,44 @@ const addArchiveRoleToMembers = async (userId) => {
...
+ * @return {Promise<usersMigrated|Object>}
+ */
+const addDefaultArchivedRole = async () => {
+  try {
+    const userSnapShot = await userModel.get();
+    const migratedUsers = [];
+
+    const usersArr = [];
+
+    userSnapShot.forEach((doc) => usersArr.push({ id: doc.id, ...doc.data() }));
```
ankushdharkar, line 187: Does this work for 1 million users?

ankushdharkar (changes requested):

models/members.js
samarpan1738: Created a new issue and PR for this : https://github.com/Real-Dev-Squad/website-backend/pull/656

routes/members.js
samarpan1738: There's an existing ticket for it https://github.com/Real-Dev-Squad/website-backend/issues/633 which has been addressed by @pallabez

samarpan1738: Cleanup ticket for what?

models/members.js
samarpan1738: We could have used a new variable here. 
But since the only values showArchived can have are 'true' or 'false' and it's intended to be used as a boolean further in the code so I guess it won't be a problem

middlewares/validators/members.js
samarpan1738: Currently, yes

controllers/members.js
samarpan1738: This applies to all users

models/members.js
samarpan1738: I think we should paginate the queries to firestore

controllers/members.js
samarpan1738: The approach I have in mind is to convert this into an async. api endpoint
First time we call this we will return an identifier for the async operation started on the server.
We can create a new collection and for each async operation we can create a document in it and store the progress of the operation in the doc 
Next time we can pass the identifier to check the status of the task

routes/members.js
samarpan1738: https://github.com/Real-Dev-Squad/website-backend/issues/671

controllers/members.js
samarpan1738: By default firestore returns all documents in a collection if no limit is explicitly specifies

Docs link: https://firebase.google.com/docs/firestore/query-data/order-limit-data

samarpan1738: Currently there are a total of 656 users. I have tested with my personal firestore with 1000 users
It's taking around 5-10 seconds to [fetch all users + update all users archived role]
By default express has a timeout of 2 minutes. So timeout won't be a problem here

models/users.js
samarpan1738: Done

test/integration/members.test.js
```diff
@@ -51,10 +40,9 @@ describe("Members", function () {
     });
 
     afterEach(async function () {
-      await addUser(userWithoutRolesObject);
-      await addUser(userWithRolesObjectWithoutArchivedProperty);
-      await addUser(userWithRolesObjectWithArchivedTrue);
-      await addUser(userWithRolesObjectWithArchivedFalse);
+      await cleanDb();
+      await addUser(userArchivedRoleFalse);
+      await addUser(userAlreadyArchived);
```
ankushdharkar, line 45: Should this be in 'before'?

test/utils/deleteRoles.js
```diff
@@ -0,0 +1,25 @@
+const firestore = require("../../utils/firestore");
+const userCollection = firestore.collection("users");
+
+/**
+ * @return {boolean} success - are roles deleted or not
+ */
+module.exports = async (userId = null, rolesToBeDeleted = [], deleteRolesObject = false) => {
```
ankushdharkar, line 7: Is `deleteRolesObject` an object or a boolean?

ankushdharkar (changes requested):

test/integration/members.test.js
samarpan1738: Yes. Resolved by https://github.com/Real-Dev-Squad/website-backend/issues/680

test/utils/deleteRoles.js
samarpan1738: As discussed, I have broken this util into 2 separate utils : `deleteRoles` and `deleteRolesObject`

test/utils/deleteRoles.js
```diff
@@ -0,0 +1,25 @@
...
+    const userData = userDoc.data();
+    if (deleteRolesObject) delete userData.roles;
+    else rolesToBeDeleted.forEach((role) => delete userData.roles[String(role)]);
+
+    await userCollection.doc(userId).set(userData);
+
+    return true;
+  } catch (error) {
+    return false;
+  }
```
shubham-y, line 24: Can we log the error using a logger?

test/utils/deleteRoles.js
```diff
@@ -0,0 +1,25 @@
+const firestore = require("../../utils/firestore");
+const userCollection = firestore.collection("users");
+
+/**
+ * @return {boolean} success - are roles deleted or not
+ */
+module.exports = async (userId = null, rolesToBeDeleted = [], deleteRolesObject = false) => {
+  if (!userId) return false;
```
shubham-y, line 8: Is there any not include this line under try block

test/utils/deleteRoles.js
```diff
@@ -0,0 +1,25 @@
+const firestore = require("../../utils/firestore");
+const userCollection = firestore.collection("users");
+
+/**
+ * @return {boolean} success - are roles deleted or not
+ */
```
shubham-y, line 6: I think we can add comments for description and params for this function

test/utils/deleteRoles.js
samarpan1738: Since I know that this line won't ever throw an error therefore felt no need to keep it inside the try
But there is also no harm in keeping it inside the try

test/utils/deleteRoles.js
samarpan1738: Sure. I'll add that

test/utils/deleteRoles.js
samarpan1738: Sure. I'll add that

shubham-y (dismissed):

test/utils/deleteRoles.js
ankushdharkar: Nit: `shouldDeleteRoles` @samarpan1738

ankushdharkar (approved):

PR #681, opened 2022-08-14
Implements correct semantics in the test setup for members integration tests

### LINKED ISSUE #680: [rfc] Implement correct semantics in the test setup for members integration tests
Ideally in the test setup part
1. beforeEach should populate data(if any) - **Setup**
2. afterEach should perform the cleanup tasks(if any) - **Teardown**

But currently, the test setup in member integration tests is not following this. In fact the opposite of this is being done here.

One reason for doing so was that, in a describe block the first test required DB to be clean and the following tests required 1 or more users to be present in the DB. 
This is bad since a test should not depend on the sequence it's running. Because of this ,in the future, if we want to run the tests in parallel we won't be able to do so.

samarpan1738:
Rearranged beforeEach and afterEach blocks

Closes #680

### DIFF
--- a/test/integration/members.test.js
+++ b/test/integration/members.test.js
@@ -37,27 +37,12 @@ const unarchivedUsersGithubIds = [
 describe("Members", function () {
   let jwt;
 
-  afterEach(async function () {
-    await addUser();
-  });
-
-  after(async function () {
+  beforeEach(async function () {
     await cleanDb();
   });
 
   describe("GET /members", function () {
-    before(async function () {
-      await cleanDb();
-    });
-
-    afterEach(async function () {
-      await addUser(userWithoutRolesObject);
-      await addUser(userWithRolesObjectWithoutArchivedProperty);
-      await addUser(userWithRolesObjectWithArchivedTrue);
-      await addUser(userWithRolesObjectWithArchivedFalse);
-    });
-
-    it("Should return empty array if no user is found", function (done) {
+    it("Should return an empty array if no user is found", function (done) {
       chai
         .request(app)
         .get("/members")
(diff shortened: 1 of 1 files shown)


test/integration/members.test.js
```diff
@@ -38,41 +38,35 @@ describe("Members", function () {
   let jwt;
 
   afterEach(async function () {
-    await addUser();
-  });
-
```
rohan09-raj, line 43: Thanks for fixing this

test/integration/members.test.js
```diff
@@ -38,41 +38,35 @@ describe("Members", function () {
...
 
     it("Should return empty array if no user is found", function (done) {
-      chai
-        .request(app)
-        .get("/members")
-        .end((err, res) => {
-          if (err) {
-            return done(err);
-          }
+      cleanDb().then(() => {
```
rohan09-raj, line 53: Is after each doesn't taking care of this ??

test/integration/members.test.js
```diff
@@ -135,25 +129,28 @@ describe("Members", function () {
...
+
     it("Should return empty array if no idle member is found", function (done) {
-      chai
-        .request(app)
-        .get("/members/idle")
-        .end((err, res) => {
-          if (err) {
-            return done(err);
-          }
+      cleanDb().then(() => {
```
rohan09-raj, line 137: Is after each doesn't taking care of this ??

rohan09-raj (changes requested):

test/integration/members.test.js
samarpan1738: After each cleans the DB but then beforeEach adds the users in the DB.

test/integration/members.test.js
samarpan1738: After each cleans the DB but then beforeEach adds the users in the DB.

rohan09-raj: Continuing the conversation from here as its the same comment https://github.com/Real-Dev-Squad/website-backend/pull/681#discussion_r945517685

test/integration/members.test.js
rohan09-raj: I'm saying you have already added cleanDb in afterEach, so calling this function here and doing something on resolving the promise, is this necessary to do so ??

samarpan1738: As discussed we'll we creating separate describe blocks for the 2 cases : when there are no users in the DB and when there are users in DB

test/integration/members.test.js
```diff
@@ -156,31 +151,32 @@ describe("Members", function () {
...
-    it("Get all the idle members in the database", function (done) {
-      chai
-        .request(app)
-        .get("/members/idle")
-        .end((err, res) => {
-          if (err) {
-            return done(err);
-          }
+    it("Should return all the idle members", function (done) {
+      addUser().then(() => {
```
rohan09-raj, line 155: Can we set this in separate `describe` in a `before`  as we discussed earlier for previous comments ?

samarpan1738: Chai recommends that if a describe has only 1 it block better don't use describe for it. It gives lint error also

samarpan1738: Updated the above comment as well
https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-hooks-for-single-case.md

rohan09-raj (approved):

PR #693, opened 2022-08-22
Modified test.yml to remove test workflow on Push

RitikJaiswal75:
### Files Causing the issue
- `.github/workflows/test.yml`

### Changes Done:
Removed the lines that run the test on push

### Is the Development Tested
- [x] Yes
- [ ] No

### Screenshot of the Problem
image

### Screenshot after changing the File
image

#### Note:
The test will run every time you make changes to your branch. So, do not worry thinking if I make changes will the tests stop?
Currently, they run 2 times on (Pull_request and Push).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,10 +4,6 @@
 name: Tests
 
 on:
-  push:
-    branches:
-      - '**'
-
   pull_request:
     branches:
       - '**'


Maheima:
LGTM!

.github/workflows/test.yml
```diff
@@ -4,10 +4,6 @@
 name: Tests
 
 on:
-  push:
-    branches:
-      - '**'
-
   pull_request:
     branches:
       - '**'
```
samarpan1738, line 9: Will this run when new commits are pushed to a PR?

vinayak-trivedi (approved):
LGTM

.github/workflows/test.yml
samarpan1738: As discussed on vc it'll run when new commits are pushed to a PR since by default this workflow will run on these events : `opened, synchronize & reopened`

Docs : https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

samarpan1738:
LGTM

Issue #688, opened 2022-08-20
[RFC] redirect to origin site after login/signup


ivinayakg:
# **Objective**
- After login/sign up from any RDS site the user should get redirected to the origin of the request rather than the main site

# **Procedure**
- Setup a route for 'goto' from [here](https://github.com/Real-Dev-Squad/website-backend/blob/develop/config/default.js#L45) to take a URL from a query and redirect the user to the destination.
- Tweak the auth handler to redirect to `/goto/?state=<TARGET_URL | ORIGIN_URL>`
- the origin_url would be taken from the request

## Request will look like this.
`https://github.com/login/oauth/authorize?client_id=<CLIENT_ID>&state=<ORIGIN_URL>` this is how the request will look like on the frontend and later will be redirected to the backend

shubham-y:
@ivinayakg How do you plan to get the origin URL?

ivinayakg:
![Signup flow of RDS - Copy](https://user-images.githubusercontent.com/73058928/186966210-c39b7df8-93d9-4162-a60d-f528d2f80d1d.png)

ivinayakg:
# Implementation of this Feature Across RDS
- RFC for status site
- RFC for member site
- RFC for crypto site
- RFC for my site
- RFC for dashboard site

PR #694, opened 2022-08-24
Redirect to origin after Sign In with GitHub

ivinayakg:
#688 
The origin will get received as `state` through query at the backend and redirected to the origin


vinayak-trivedi (approved):
LGTM 👍

pallabez (approved):
LGTM

PR #703, opened 2022-09-10
Feat/get task using skill level

### LINKED ISSUE #701: create a tasks/:skill/:level endpoint
create a task/:skill/:level endpoint which will return a task based on the `skillLevel` passed in the route params. For more reference read `step 3` of [this](https://github.com/Real-Dev-Squad/todo-action-items/issues/72#issuecomment-1144031642)

vinayak-trivedi:
### what is the change?

- Added an API that will return a task based on the skill and level provided.
- closes #701 

 **need to write tests so changing the PR to draft**


vinayak-trivedi:
closing this PR as we do not need this API, a middleware will be made which will assign the task automatically

Issue #698, opened 2022-08-27
Caching module for API's


yesyash:
### Description
Create a module that helps us cache API responses and store them in memory for a set amount of time.

PR #713, opened 2022-09-20
Made a in memory cache module and added it to /tasks endpoint

yesyash:
### Description
- Created a cache module to get and store API responses from the cache pool for a controller.
- Cached responses from /tasks endpoint using the cache module. 

### Working
cache module flow - pr 1

### DIFF
--- a/utils/cache.js
+++ b/utils/cache.js
@@ -1,7 +1,13 @@
 const CACHEEXPIRYTIME = 2; // Default cache expriy time in Minutes
+const CACHESIZE = 10; // Default maximum size for cache pool (in Megabytes)
 const minutesToMilliseconds = (minutes) => minutes * 60000; // Converts given minutes into milliseconds
 
-const globalStore = () => {
+/**
+ * Cache pool to get and store API responses
+ * @param {Object} [opt] options for cache pool
+ * @param {number} opt.maximumSize Maximum size of the cache pool in Megabytes (MB)
+ */
+const cachePool = (opt = { maximumSize: CACHESIZE }) => {
   const cacheStore = new Map();
   let hits = 0;
 
@@ -30,8 +36,8 @@ const globalStore = () => {
   };
 
   /**
-   * Add API response to our cacheStore.
-   * @param {string} key {string}
+   * Add API response to cacheStore.
+   * @param {string} key
    * @param {Object} value Value to be stored inside the cache.
    * @param {number} value.priority Priority of the api
    * @param {number} value.expiry Expiry time of api
@@ -50,8 +56,8 @@ const globalStore = () => {
   return { get, set, hits, cacheStore };
 };
 
-// Initialize globalstore to be used.
-const store = globalStore();
+// Initialize cache pool.
+const pool = cachePool();
 
 /**
  * Caching middleware for API resposnes.
@@ -63,7 +69,7 @@ const store = globalStore();
 const cache = (data = { priority: 2, expiry: CACHEEXPIRYTIME }) => {
   return async (req, res, next) => {
     const key = "__cache__" + req.method + req.originalUrl;
-    const dataInCache = store.get(key);
+    const dataInCache = pool.get(key);
 
     if (dataInCache) {
       res.send(dataInCache);
@@ -95,7 +101,7 @@ const cache = (data = { priority: 2, expiry: CACHEEXPIRYTIME }) => {
           size: Buffer.byteLength(apiResponse),
         };
 
-        store.set(key, cacheValue);
+        pool.set(key, cacheValue);
         return oldEnd.apply(res, [chunk, ...args]);
       };
 
--- a/utils/cache.js
+++ b/utils/cache.js
@@ -1,13 +1,13 @@
-const CACHEEXPIRYTIME = 2; // Default cache expriy time in Minutes
-const CACHESIZE = 10; // Default maximum size for cache pool (in Megabytes)
-const minutesToMilliseconds = (minutes) => minutes * 60000; // Converts given minutes into milliseconds
+const CACHE_EXPIRY_TIME_MIN = 2;
+const CACHE_SIZE_MB = 10;
+const minutesToMilliseconds = (minutes) => minutes * 60000;
 
 /**
  * Cache pool to get and store API responses
  * @param {Object} [opt] options for cache pool
  * @param {number} opt.maximumSize Maximum size of the cache pool in Megabytes (MB)
  */
-const cachePool = (opt = { maximumSize: CACHESIZE }) => {
+const cachePool = (opt = { maximumSize: CACHE_SIZE_MB }) => {
   const cacheStore = new Map();
   let hits = 0;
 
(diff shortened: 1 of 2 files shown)


utils/cache.js
```diff
@@ -0,0 +1,113 @@
+const CACHEEXPIRYTIME = 2; // Default cache expriy time in Minutes
+const CACHESIZE = 10; // Default maximum size for cache pool (in Megabytes)
```
samarpan1738, line 2: 1. Use underscore to separate words
2. Should we specify units in name? Reason: We are not using the defaults like bytes for size and milliseconds for time
```suggestion
const CACHE_EXPIRY_TIME_MIN = 2; // Default cache expriy time in Minutes
const CACHE_SIZE_MB = 10; // Default maximum size for cache pool (in Megabytes)
```

utils/cache.js
```diff
@@ -0,0 +1,113 @@
...
+  const cacheStore = new Map();
+  let hits = 0;
+
+  /**
+   * Get an API Respnose from cacheStore.
+   * @param {string} key
+   * @returns {null | object}
+   */
+  const get = (key) => {
+    const apiData = cacheStore.get(key);
```
samarpan1738, line 20: Can this cache only be used for storing api data or it's a generic module?
If no, then we should rename this variable to something more generic like `cachedData` instead of `apiData `

utils/cache.js
```diff
@@ -0,0 +1,113 @@
...
+   * @returns {number} : statusCode
+   */
+  const set = async (key, value) => {
+    try {
+      cacheStore.set(key, value);
+      return 200;
+    } catch (err) {
+      return 500;
+    }
+  };
```
samarpan1738, line 54: 1. Our cache seems to be tightly coupled with HTTP response codes. Can we make it more generic? Ideally it should not return anything
2. Why are we returning 500 response code if we are unable to set cache? This would mean that even if we are successful in fetching data from our db but fail to set it in cache we'll communicate a failure to the client.

utils/cache.js
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Caching middleware for API resposnes.
+ * @param {Object} [data] Options to handle individual api cache.
+ * @param {number} data.priority The priority of api in cache store.
+ * @param {number} data.expiry Cache expiry time of api in minutes.
+ * @returns {function} middleware function to help cache api response.
+ */
+const cache = (data = { priority: 2, expiry: CACHEEXPIRYTIME }) => {
+  return async (req, res, next) => {
```
samarpan1738, line 70: Can we extract this to a separate middleware file?

utils/cache.js
```diff
@@ -0,0 +1,113 @@
...
+    const key = "__cache__" + req.method + req.originalUrl;
+    const dataInCache = pool.get(key);
```
samarpan1738, line 72: nit: This can be named better imo
```suggestion
    const cachedData = pool.get(key);
```

utils/cache.js
```diff
@@ -0,0 +1,113 @@
...
+        const cacheValue = {
+          priority: data.priority,
+          response: apiResponse,
+          expiry: new Date().getTime() + minutesToMilliseconds(data.expiry),
+          size: Buffer.byteLength(apiResponse),
+        };
+
+        pool.set(key, cacheValue);
+        return oldEnd.apply(res, [chunk, ...args]);
+      };
```
samarpan1738, line 106: Instead of intercepting request can we have a middleware after the request handler in the route that'll check if we need to write the value in cache or not?

yesyash: Here we are Intercepting request to get the response being sent from the controller of that route and then store its response in the cache, not sure how would it be done using middleware after the request handler. 

Would you please give an example of your implementation?

samarpan1738: I guess we’ll have to use this intercepting way of capturing the response body. I couldn’t find a better way of doing this

utils/cache.js
samarpan1738: Sounds good👍

utils/cache.js
rupal-2505: Isn't Middleware is called only before the route handler is called. And to have access to the response object, but  don't have the result of the route handler. So i think the intercepting way is way more good.

rupal-2505 (commented):
LGTM

Maheima (approved):
LGTM

ivinayakg (commented):
the code looks fantastic, there are pieces that I am not 100% sure I will read about that will come back.
I know you already know this but just a reminder, add tests before merging 😊

yesyash:
> the code looks fantastic, there are pieces that I am not 100% sure I will read about that will come back. I know you already know this but just a reminder, add tests before merging 😊

Working on writing tests will be out soon. Thank you for mentioning it.

utils/cache.js
```diff
@@ -0,0 +1,113 @@
...
+      res.end = (chunk, ...args) => {
+        if (chunk) {
+          chunks.push(chunk);
+        }
+
+        const apiResponse = Buffer.concat(chunks).toString();
+
+        const cacheValue = {
+          priority: data.priority,
+          response: apiResponse,
```
ankurnarkhede, line 99: The API response here seems to be in the var `apiResponse` but we are sending the whole `cacheValue` as the response on line 75, if data is found in cache? 🤔

utils/cache.js
```diff
@@ -0,0 +1,113 @@
...
+
+    if (!apiData) {
+      return null;
+    }
+
+    const isApiDataExpired = new Date().getTime() > apiData.expiry;
+
+    // If data is expired remove it from store, time to get a fresh copy.
+    if (isApiDataExpired) {
+      cacheStore.delete(key);
```
ankurnarkhede, line 30: Why are we creating this cache model ourselves rather than using any npm lib for this?

ankurnarkhede: We can use something like: https://www.npmjs.com/package/node-cache...

ankurnarkhede (changes requested):

utils/cache.js
yesyash: `pool.get` in line 72 will return the `apiResponse` and not the entire `cacheValue` so only the original response is being sent in case of a cache hit.

yesyash: Would it be better to have the `get` function return us the whole `cacheValue` and destructure it at line 72 for consistency?

utils/cache.js
yesyash: I had a look at this package and was thinking to use it at the start, but I think we will have to write a wrapper around it, later on, to implement features like: 
1. Priority flags
2. custom clean-up function logic which depends on priority flags 
3. ability to set maximum memory size for cache pool (node-cache has something similar with maxKeys which allows us to have a limit on the number of keys that can be stored but not the amount of memory cache pool takes)

And hence thought to go ahead with our own cache module.

Also, do we really need a package like `node-cache` when we will probably be not making use of the majority of its features at the current stage?

You can refer to issue ticket #698  which would give a clear understanding of the features I want to build for V1 of this module.

utils/cache.js
ankurnarkhede: I think the current implementation of just returning the `response` is fine 👍

utils/cache.js
ankurnarkhede: 👍

ankurnarkhede (approved):

PR #721, opened 2022-09-28
Feat/logs route

vinayak-trivedi:


vinayak-trivedi:
closing this as most of the code is not required we can just use an existing function present in the logs model function

Issue #702, opened 2022-09-05
Strict value acceptance for status in Task object


ivinayakg:
Make the task’s status accept any of these values `AVAILABLE | ASSIGNED | IN_PROGRESS | BLOCKED | SMOKE_TESING | COMPLETED | NEEDS_REVIEW | IN_REVIEW | APPROVED | MERGED | SANITY_CHECK | REGRESSION_CHECK | RELEASED | VERIFIED`
(Right now it can accept any value)

## Approach
Right now we are using a custom middleware using `joi` check [here](https://github.com/Real-Dev-Squad/website-backend/blob/develop/middlewares/validators/tasks.js#L16), we just need to add a `.valid([values we want to restrict the status to])` and that's it.

PR #723, opened 2022-10-02
Strict value acceptance for status in Task object

ivinayakg:
RFC - #702 
## What has been done?
- An additional test case has been added for this.
- updated the `joi` middleware for tasks for this issue.


pallabez (approved):
LGTM, thanks for the tests as well.

Issue #460, opened 2021-12-27
[RFC] New Task Workflow 


Pavangbhat:
## Project Description

---

This project will allow the tasks of Real Dev Squad to be handled in an efficient way. Each task can be in one state such as `completed`**,** `blocked,` etc. This project will allow us to have a new set of states that a task can be in and a few things will be automated such as if a member could not complete a task within the given period of time, the task will be unassigned from the member and it will be available for other members to be picked up.

The current status that a task can be in:- `active | assigned | unAssigned | blocked | completed | High Priority | pending`

**changes:-**  **`active → inProgress`, `unAssigned → available`, `completed →  verified`**

This feature allows the task status to be in:- `available | assigned | inProgress | blocked | smokeTesting | completed | needsReview | inReview | approved | merged | sanityCheck | regressionCheck | released | verified` 

Reference 👇🏻

[https://app.creately.com/diagram/heYDw2DIsqh/view](https://app.creately.com/diagram/heYDw2DIsqh/view)

## Tasks

---

## website-backend

---

1) Make task’s status to accept any of these values `available | assigned | inProgress | blocked | smokeTesting | completed | needsReview | inReview | approved | merged | sanityCheck | regressionCheck | released | verified`  (Right now it can accept any value**)**

2) create an endpoint that fires a function that loops through all tasks whose status is `assigned | inProgres`  and if `endsOn` > `currentTime` SET task’s status  back to `available`, assignee field to be cleared and log the member and task details (For future use)

3) migration script to make **active → inProgress, unAssigned → available (Does not need any changes on status except availability panel)**

## **website-status**

---

1) On the availability panel render only tasks that are **available** not **unAssigned(handle it before migration such that both available** and **unAssigned works)**

2) When the task is assigned to a member from the availability panel make the status of the task as **assigned** and **startedOn** should be computed and **endsOn** should be 14 days by default

## website-dashboard

---

1) Remove **startedOn** input field from the form**,** and  **endsOn** input field **** will be hidden by default 
2) default status field  should be **available** 

3) when the **status** field is selected as **assigned** show **assignee** field and make it compulsory and show **endsOn** date and default **endsOn** date is 14 days 

4) When a new task is created **startedOn** will be computed if the task is assigned else ignored 

## website-members

---

1) Makes use of the task’s status(**Active**) on members profile Active tasks section need to make it either **(Active || inProgress)**

Pavangbhat:
[...]

PR #725, opened 2022-10-02
Status changed by User on the task will be logged

ivinayakg:
# Parent RFC - #460 

## How will this work?
- Will create a new log and save to the database about the event

## What has been done?
- called the `addlog` API to create a log for the event
- Response after a successful call
image

## Edge Cases Covered
- If the user is trying to change the status to `MERGED`, it will be forbidden.
- If `status` is already `COMPLETED` and the user is trying to change only the `percentCompleted` to less than 100, it will give a bad request.
- If `status` is not `COMPLETED` and the user is trying to change the `percentCompleted` to 100, it will give a bad request.

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -173,23 +173,27 @@ const updateTaskStatus = async (req, res, next) => {
       body: {
         subType: "update",
         new: {},
-        messages: [],
+        message: "",
       },
     };
 
-    if (req.body.status) {
+    if (req.body.status && !req.body.percentCompleted) {
       taskLog.body.new.status = req.body.status;
-      taskLog.body.messages.push(`changed status to ${req.body.status}`);
+      taskLog.body.message = `@${username} changed status to ${req.body.status}`;
     }
-    if (req.body.percentCompleted) {
+    if (req.body.percentCompleted && !req.body.status) {
       taskLog.body.new.percentCompleted = req.body.percentCompleted;
-      taskLog.body.messages.push(`changed percentCompleted to ${req.body.percentCompleted}`);
+      taskLog.body.message = `@${username} changed percentCompleted to ${req.body.percentCompleted}`;
+    }
+
+    if (req.body.percentCompleted && req.body.status) {
+      taskLog.body.new.percentCompleted = req.body.percentCompleted;
+      taskLog.body.new.status = req.body.status;
+      taskLog.body.message = `@${username} changed percentCompleted to ${req.body.percentCompleted} and status was changed to ${req.body.status}`;
     }
 
     if (req.body.percentCompleted === 100 && req.body.status === TASK_STATUS.COMPLETED) {
-      taskLog.body.messages = [
-        `changed percentCompleted to ${req.body.percentCompleted} and status was changed to COMPLETED`,
-      ];
+      taskLog.body.message = `@${username} changed percentCompleted to ${req.body.percentCompleted} and status was changed to COMPLETED`;
     }
 
     const [, taskLogResult] = await Promise.all([
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -1,5 +1,6 @@
 const tasks = require("../models/tasks");
 const { TASK_STATUS, TASK_STATUS_OLD } = require("../constants/tasks");
+const { addLog } = require("../models/logs");
 const { OLD_ACTIVE, OLD_BLOCKED, OLD_PENDING } = TASK_STATUS_OLD;
 const { IN_PROGRESS, BLOCKED, SMOKE_TESTING, ASSIGNED } = TASK_STATUS;
 /**
@@ -142,7 +143,7 @@ const updateTaskStatus = async (req, res, next) => {
   try {
     const taskId = req.params.id;
     const { dev } = req.query;
-    const { id: userId } = req.userData;
+    const { id: userId, username } = req.userData;
     const task = await tasks.fetchSelfTask(taskId, userId);
 
     if (task.taskNotFound) return res.boom.notFound("Task doesn't exist");
@@ -152,13 +153,30 @@ const updateTaskStatus = async (req, res, next) => {
 
     await tasks.updateTask(req.body, taskId);
 
+    const taskLog = {
+      type: "task",
+      meta: {
+        userId,
+        taskId,
+        username,
+      },
+    };
+
+    if (req.body.status) {
+      taskLog.body = `${username} changed task status to ${req.body.status}`;
+    } else if (req.body.percentCompleted) {
+      taskLog.body = `${username} changed task percent Completed to ${req.body.percentCompleted}`;
+    }
+
+    await addLog(taskLog.type, taskLog.meta, taskLog.body);
+
     if (dev) {
       if (req.body.percentCompleted === 100) {
         return next();
       }
     }
 
-    return res.json({ message: "Task updated successfully!" });
+    return res.json({ message: "Task updated successfully!", taskLog });
   } catch (err) {
     logger.error(`Error while updating task status : ${err}`);
     return res.boom.badImplementation("An internal server error occured");
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -173,14 +173,17 @@ const updateTaskStatus = async (req, res, next) => {
       body: {
         subType: "update",
         new: {},
+        messages: [],
       },
     };
 
     if (req.body.status) {
       taskLog.body.new.status = req.body.status;
+      taskLog.body.messages.push(`${username} changed status to ${req.body.status}`);
     }
     if (req.body.percentCompleted) {
       taskLog.body.new.percentCompleted = req.body.percentCompleted;
+      taskLog.body.messages.push(`${username} changed percentCompleted to ${req.body.percentCompleted}`);
     }
 
     const [, taskLogResult] = await Promise.all([
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -151,8 +151,6 @@ const updateTaskStatus = async (req, res, next) => {
     if (task.taskData.status === "VERIFIED")
       return res.boom.forbidden("Status cannot be updated. Please contact admin.");
 
-    await tasks.updateTask(req.body, taskId);
-
     const taskLog = {
       type: "task",
       meta: {
@@ -168,8 +166,11 @@ const updateTaskStatus = async (req, res, next) => {
       taskLog.body = `${username} changed task percent Completed to ${req.body.percentCompleted}`;
     }
 
-    const { id: taskLogId } = await addLog(taskLog.type, taskLog.meta, taskLog.body);
-    taskLog.id = taskLogId;
+    const [taskResult] = await Promise.all([
+      tasks.updateTask(req.body, taskId),
+      addLog(taskLog.type, taskLog.meta, taskLog.body),
+    ]);
+    taskLog.id = taskResult.taskId;
 
     if (dev) {
       if (req.body.percentCompleted === 100) {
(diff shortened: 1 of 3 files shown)


controllers/tasks.js
```diff
@@ -142,23 +143,44 @@ const updateTaskStatus = async (req, res, next) => {
...
+        userId,
+        taskId,
+        username,
+      },
+      body: [],
+    };
+
+    if (req.body.status) {
+      taskLog.body.push(`${username} changed task status to ${req.body.status}`);
+    }
```
vinayak-trivedi, line 166: if in the frontend I want to check the value of status, then I wouldn't be able to check it right? I think we should send the status as well with this. what do you think?

controllers/tasks.js
```diff
@@ -142,23 +143,48 @@ const updateTaskStatus = async (req, res, next) => {
...
+      },
+    };
+
+    if (req.body.status) {
+      taskLog.body.messages.push(`${username} changed task status to ${req.body.status}`);
+      taskLog.body.newStatus = req.body.status;
+    }
+    if (req.body.percentCompleted) {
+      taskLog.body.messages.push(`${username} changed task percent Completed to ${req.body.percentCompleted}`);
+      taskLog.body.newPercentCompleted = req.body.percentCompleted;
```
vinayak-trivedi, line 172: do we need to put username here as we are already sending the username in meta above

test/integration/tasks.test.js
```diff
@@ -394,6 +402,79 @@ describe("Tasks", function () {
...
+        endsOn: 1234,
+        startedOn: 4567,
+        status: "REVIEW",
+        percentCompleted: 10,
+        participants: [],
+        assignee: appOwner.username,
+        completionAward: { [DINERO]: 3, [NEELAM]: 300 },
+        lossRate: { [DINERO]: 1 },
+        isNoteworthy: true,
+      };
```
RitikJaiswal75, line 442: Please store this data in a fixture and pass it here rather than retyping it everywhere

dynamic data can be passed as a parameter

controllers/tasks.js
```diff
@@ -142,23 +143,59 @@ const updateTaskStatus = async (req, res, next) => {
...
     const taskId = req.params.id;
     const { dev } = req.query;
-    const { id: userId } = req.userData;
+    const { id: userId, username } = req.userData;
     const task = await tasks.fetchSelfTask(taskId, userId);
 
     if (task.taskNotFound) return res.boom.notFound("Task doesn't exist");
     if (task.notAssignedToYou) return res.boom.forbidden("This task is not assigned to you");
-    if (task.taskData.status === "VERIFIED")
+    if (task.taskData.status === "VERIFIED" || req.body.status === "MERGED")
```
RitikJaiswal75, line 151: image
Are all these states being checked?
like for needs review Percentage should be > 80

test/integration/tasks.test.js
```diff
@@ -313,6 +313,19 @@ describe("Tasks", function () {
...
+      type: "feature",
+      endsOn: 1234,
+      startedOn: 4567,
+      status: "VERIFIED",
+      percentCompleted: 10,
+      participants: [],
+      completionAward: { [DINERO]: 3, [NEELAM]: 300 },
+      lossRate: { [DINERO]: 1 },
+      isNoteworthy: true,
+    };
```
RitikJaiswal75, line 327: This is being re-used a lot create a function in fixtures and call it with valid statuses

test/integration/tasks.test.js
```diff
@@ -371,13 +392,47 @@ describe("Tasks", function () {
...
+      const res = await chai
+        .request(app)
+        .patch(`/tasks/self/${taskId}`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send(taskStatusData);
+
+      expect(res).to.have.status(403);
+      expect(res.body.message).to.be.equal("Status cannot be updated. Please contact admin.");
+    });
+    it("Should give 403 if new status is 'MERGED' ", async function () {
```
RitikJaiswal75, line 405: An HTTP 403 response code means that a client is forbidden from accessing a valid URL

Does status 403 hold good here? if yes can you explain

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403

test/integration/tasks.test.js
```diff
@@ -371,13 +392,47 @@ describe("Tasks", function () {
...
+        .request(app)
+        .patch(`/tasks/self/${taskId}`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send({ ...taskStatusData, status: "COMPLETED" });
+
+      expect(res).to.have.status(403);
+      expect(res.body.message).to.be.equal("Status cannot be updated. Task is not completed yet");
+    });
+
+    it("Should give 403 if status is COMPLETED and newpercent is less than 100", async function () {
```
RitikJaiswal75, line 428: how is it different from above test, please make a proper sentence

controllers/tasks.js
```diff
@@ -142,23 +143,59 @@ const updateTaskStatus = async (req, res, next) => {
...
+        new: {},
+      },
+    };
+
+    if (req.body.status) {
+      taskLog.body.new.status = req.body.status;
+    }
+    if (req.body.percentCompleted) {
+      taskLog.body.new.percentCompleted = req.body.percentCompleted;
+    }
```
yesyash, line 184: In case of `req.body.status` & `req.body.percentCompleted` not being present then `taskLog.body.new.status` & `taskLog.body.new.percentCompleted` will be undefined, is this okay ?

ivinayakg: There are if statements for that reason only.

yesyash: What I meant was are `taskLog.boy.new.status` & `taskLog.body.new.percentCompleted` optional fields ? Is it fine okay if we do not have them ?

controllers/tasks.js
```diff
@@ -142,23 +143,59 @@ const updateTaskStatus = async (req, res, next) => {
...
 
     if (task.taskNotFound) return res.boom.notFound("Task doesn't exist");
     if (task.notAssignedToYou) return res.boom.forbidden("This task is not assigned to you");
-    if (task.taskData.status === "VERIFIED")
+    if (task.taskData.status === "VERIFIED" || req.body.status === "MERGED")
       return res.boom.forbidden("Status cannot be updated. Please contact admin.");
 
-    await tasks.updateTask(req.body, taskId);
+    if (task.taskData.status === "COMPLETED" && req.body.percentCompleted < 100) {
+      if (req.body.status === "COMPLETED" || !req.body.status) {
```
pallabez, line 155: What are we checking? Can we add a comment here?

controllers/tasks.js
```diff
@@ -142,23 +143,59 @@ const updateTaskStatus = async (req, res, next) => {
...
+      tasks.updateTask(req.body, taskId),
+      addLog(taskLog.type, taskLog.meta, taskLog.body),
+    ]);
+    taskLog.id = taskLogResult.id;
 
     if (dev) {
       if (req.body.percentCompleted === 100) {
         return next();
       }
     }
```
pallabez, line 196: What does it do? Is it a middleware?

controllers/tasks.js
ivinayakg: no, it wouldn't be as the approach was to log the task event in a manner that the raw data should be self-explanatory about what was the event that happened with specificity

controllers/tasks.js
ivinayakg: not part of this PR, I belive this is something to do with task leveling

controllers/tasks.js
ivinayakg: im trying to cover as many edge cases as possible, so this is the code to cover one of the edge cases.
what do you think should we write docs for it or write comments there?

test/integration/tasks.test.js
ivinayakg: yes, it is good here as I'm trying to forbid the user to put the status as merged 
I want to give that functionality only to the super-user only

test/integration/tasks.test.js
ivinayakg: they are very different check again

controllers/tasks.js
ivinayakg: for that i guess we should have a proper discussion with other members on the call

controllers/tasks.js
```diff
@@ -142,23 +143,59 @@ const updateTaskStatus = async (req, res, next) => {
...
-    if (task.taskData.status === "VERIFIED")
+    if (task.taskData.status === "VERIFIED" || req.body.status === "MERGED")
       return res.boom.forbidden("Status cannot be updated. Please contact admin.");
 
-    await tasks.updateTask(req.body, taskId);
+    if (task.taskData.status === "COMPLETED" && req.body.percentCompleted < 100) {
+      if (req.body.status === "COMPLETED" || !req.body.status) {
+        return res.boom.forbidden("Task percentCompleted can't updated as status is COMPLETED");
+      }
+    }
```
ivinayakg, line 158: if the task status is `COMPLETED` and the user is trying to change the `percentCompleted` with a value of less than 100, if so then to check wether the user is also trying to change the status of the task to something else then `COMPLETED` otherwise will give an error.
### Basically the status `COMPLETED` should be there only with percentCompleted as `100`

controllers/tasks.js
```diff
@@ -142,23 +143,59 @@ const updateTaskStatus = async (req, res, next) => {
...
+
+    if (req.body.status === "COMPLETED" && task.taskData.percentCompleted !== 100) {
+      if (req.body.percentCompleted !== 100) {
+        return res.boom.forbidden("Status cannot be updated. Task is not completed yet");
+      }
+    }
```
ivinayakg, line 164: If the user is trying to change the status only with the value of `COMPLETED` and the value of percentCompleted is not 100 and the new value (if so) is also not 100 then will give an error.

### Basically the status `COMPLETED` should be there only with percentCompleted as `100`

controllers/tasks.js
Pavangbhat, line 151: Let's please use constants

Pavangbhat (changes requested):
Please make use of constants where ever possible

controllers/tasks.js
ivinayakg: can you please elaborate a bit, I didn't get what you were saying...

Pavangbhat: I meant we can place these strings in a seperate file ( can be referred as constants ) and use it. probably there is already a file for task status constants

ivinayakg: ok, got it make sense

vinayak-trivedi (dismissed):
LGTM

ivinayakg:
# DO NOT MERGE

Maheima (approved):
LGTM!

controllers/tasks.js
```diff
@@ -157,23 +158,64 @@ const updateTaskStatus = async (req, res, next) => {
...
+    const taskLog = {
+      type: "task",
+      meta: {
+        userId,
+        taskId,
+        username,
+      },
+      body: {
+        subType: "update",
+        new: {},
```
prakashchoudhary07, line 190: Why are we passing empty object here?

prakashchoudhary07 (approved):

prakashchoudhary07:
Maybe we can create a separate middleware that has all rules regarding the tasks?

test/integration/tasks.test.js
```diff
@@ -453,13 +473,47 @@ describe("Tasks", function () {
...
+
+    it("Should give 400 if status is COMPLETED and newpercent is less than 100", async function () {
       const taskData = {
         title: "Test task",
         type: "feature",
         endsOn: 1234,
         startedOn: 4567,
-        status: "VERIFIED",
-        percentCompleted: 10,
+        status: "COMPLETED",
```
ankushdharkar, line 515: Any constants should come from a constants file. "MERGED", "COMPLETED", etc

ankushdharkar: (Commented to create the placeholder for #859)

PR #726, opened 2022-10-03
Added tests to CachePool

yesyash:
### Issue #698 

Added tests for the cache pool function committed before. 
Does not cover the cache middleware that is used to cache routes.

### DIFF
--- a/test/fixtures/cache/cache.js
+++ b/test/fixtures/cache/cache.js
@@ -1,0 +1,16 @@
+const tasksList = require("../tasks/tasks");
+
+// Returns a dummy sample value to be stored in cache
+const getDummyResponse = (expiryTime = 2) => {
+  const apiResponse = { message: "Tasks returned successfully!", tasks: tasksList() };
+  const stringifiedResponse = JSON.stringify(apiResponse);
+
+  return {
+    expiry: new Date().getTime() + expiryTime * 60000,
+    priority: 1,
+    response: stringifiedResponse,
+    size: Buffer.byteLength(stringifiedResponse),
+  };
+};
+
+module.exports = { getDummyResponse };
--- a/test/unit/utils/cache.test.js
+++ b/test/unit/utils/cache.test.js
@@ -1,0 +1,47 @@
+const { expect } = require("chai");
+
+const timeData = require("../../fixtures/time/time");
+const { getDummyResponse } = require("../../fixtures/cache/cache");
+
+const { cachePool, minutesToMilliseconds } = require("../../../utils/cache");
+
+describe.skip("cache pool", function () {
+  const testKey = "__cache__/testkey";
+  const pool = cachePool();
+
+  describe("minutesToMilliSeconds", function () {
+    it("should convert minutes and return in milliseconds", function () {
+      timeData.minutesToMilliseconds.forEach((time) => {
+        const result = minutesToMilliseconds(time.param);
+        expect(result).to.be.equal(time.result);
+      });
+    });
+  });
+  describe("cachePool.set", function () {
+    it("should add an item to cache pool and return value `true`", async function () {
+      const response = await pool.set(testKey, getDummyResponse());
+      expect(response).to.be.true;
+    });
+  });
+
+  describe("cachePool.get", function () {
+    it("should return value `null` if key is not present in cache", function () {
+      const value = pool.get("dummyInvalidKey");
+      expect(value).to.be.null;
+    });
+
+    it("successfully read from cache pool", function () {
+      const value = pool.get(testKey);
+
+      expect(value).to.have.all.keys("message", "tasks");
+      expect(pool.hits).to.be.equal(1);
+    });
+
+    it("should return value `null` for expired data", function () {
+      pool.set(testKey, getDummyResponse(-10));
+      const response = pool.get(testKey);
+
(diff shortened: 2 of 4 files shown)


test/fixtures/cache/cache.js
```diff
@@ -0,0 +1,16 @@
+const tasksList = require("../tasks/tasks");
+
+// Returns a dummy sample value to be stored in cache
+const getDummyResponse = (expiryTime = 2) => {
+  const apiResponse = { message: "Tasks returned successfully!", tasks: tasksList() };
+  const stringifiedResponse = JSON.stringify(apiResponse);
+
+  return {
+    expiry: new Date().getTime() + expiryTime * 60000,
```
Roli24, line 9: I think it's better to declare numbers as constant and use them

test/unit/utils/cache.test.js
```diff
@@ -0,0 +1,47 @@
...
+    it("should return value `null` if key is not present in cache", function () {
+      const value = pool.get("dummyInvalidKey");
+      expect(value).to.be.null;
+    });
+
+    it("successfully read from cache pool", function () {
+      const value = pool.get(testKey);
+
+      expect(value).to.have.all.keys("message", "tasks");
+      expect(pool.hits).to.be.equal(1);
```
Roli24, line 37: IMO it should be reflected in all places where ever we are trying to declare or use some static value i.e 1, 1000, "test" that we can declare in constant

Roli24 (commented):
Don't have knowledge about JS but basic comments i have put.

test/fixtures/cache/cache.js
yesyash: I was thinking of doing this, but what if we want to test for different expiry times like 2mins, 10mins ...etc?

Roli24: IMO,
DEFAULT_EXPIRY_TIME_MIN
inspite of that
EXPIRY_TIME_IN_MIN is more meaningful.

test/unit/utils/cache.test.js
```diff
@@ -0,0 +1,50 @@
...
+  describe("cachePool.set", function () {
+    it("should add an item to cache pool and return value `true`", async function () {
+      const response = await pool.set(testKey, getDummyResponse());
+      expect(response).to.be.true;
+    });
+  });
+
+  describe("cachePool.get", function () {
+    it("should return value `null` if key is not present in cache", function () {
+      const value = pool.get("dummyInvalidKey");
```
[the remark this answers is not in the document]
yesyash: Roli can you please check now I have moved the hard-coded strings to `test/fixtures/cache/cache.js` is this method right?

prakashchoudhary07 (dismissed):
LGTM

Maheima (dismissed):
LGTM!

test/unit/utils/cache.test.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    it("successfully read from cache pool", function () {
+      const value = pool.get(testKey);
+
+      expect(value).to.have.all.keys("message", "tasks");
+      // commenting it for now
+      // this test will fail due to a bug in the cache module.
+      // expect(pool.hits).to.be.equal(1);
+    });
```
samarpan1738, line 45: This seems to be a flaky test. It will pass if the above describe block for "cachePool.set" runs before this one. Otherwise it'll fail.
It'll also fail if the next it block runs before this one

test/unit/utils/cache.test.js
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  describe("cachePool.get", function () {
+    it("should return value `null` if key is not present in cache", function () {
+      const value = pool.get(INVALID_KEY);
+      expect(value).to.be.equal(null);
+    });
+
+    it("successfully read from cache pool", function () {
```
samarpan1738, line 38: nit
```suggestion
    it("should successfully read from cache pool", function () {
```

samarpan1738 (changes requested):
Please check the review comments

test/unit/utils/cache.test.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    it("should successfully read from cache pool if a key is present in cache", function () {
+      pool.set(testKey, getDummyResponse());
+      const value = pool.get(testKey);
+
+      expect(value).to.have.all.keys("message", "tasks");
+      // commenting it for now
+      // this test will fail due to a bug in the cache module.
+      // expect(pool.hits).to.be.equal(1);
```
harshith-venkatesh, line 45: Is it not handled yet?

harshith-venkatesh (approved):

test/unit/utils/cache.test.js
yesyash: This needs to be changed in the `utils/cache.js` before handling it here, will require a different pr where I first change that, and then change it here in `cache.test.js`

harshith-venkatesh: Got it 
Have approved the PR
Please create a ticket to track it

yesyash: Done. #730

samarpan1738 (approved):

prakashchoudhary07:
Closing this, as tests for this have been written based of functionality change

PR #732, opened 2022-10-04
Feat/get task endpoint

### LINKED ISSUE #727: create a get endpoint for `/tasks/details/:id`
I need a `GET` endpoint for `/tasks/details/:id` which returns all the details of that particular task, which we will be using in the logs page of dashboard site

vinayak-trivedi:
closes #727 

### what is the change
created a `GET` endpoint `/details/:id` for the task which is going to return all the details of a particular task

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -112,6 +112,20 @@ const getSelfTasks = async (req, res) => {
     return res.boom.badImplementation("An internal server error occurred");
   }
 };
+
+const getTask = async (req, res) => {
+  try {
+    const taskId = req.params.id;
+    const { taskData } = await tasks.fetchTask(taskId);
+
+    if (!taskData) {
+      return res.boom.notFound("Task not found");
+    }
+    return res.json({ message: "task returned successfully", taskData });
+  } catch (err) {
+    return res.boom.badImplementation("An internal server error occurred");
+  }
+};
 /**
  * Updates the task
  *
@@ -195,6 +209,7 @@ module.exports = {
   updateTask,
   getSelfTasks,
   getUserTasks,
+  getTask,
   updateTaskStatus,
   overdueTasks,
 };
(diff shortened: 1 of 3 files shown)


RitikJaiswal75 (commented):
LGTM

routes/tasks.js
```diff
@@ -13,6 +13,7 @@ router.get("/self", authenticate, tasks.getSelfTasks);
 router.get("/overdue", authenticate, authorizeRoles([SUPERUSER]), tasks.overdueTasks);
 router.post("/", authenticate, authorizeRoles([APPOWNER, SUPERUSER]), createTask, tasks.addNewTask);
 router.patch("/:id", authenticate, authorizeRoles([APPOWNER, SUPERUSER]), updateTask, tasks.updateTask);
+router.get("/details/:id", tasks.getTask);
```
ivinayakg, line 16: why not just do `/:id` for the task API I believe that would be a better way to manage the API, I do acknowledge that it will have a conflict with `/:username` but why not change that instead? What do you think?
cause ofcourse we will have to edit all the consumers of this `/:username` endpoint but that should work for the betterment in the long run as we would be able to follow a standard way of doing CRUD operations on every collection we have.

ivinayakg (changes requested):
PR looks good 🙌
Just one suggestion I have below.

Maheima (approved):
LGTM!

routes/tasks.js
vinayak-trivedi: this will be done in another PR.

ivinayakg (dismissed):

ivinayakg (approved):
Go ahead and merge

controllers/tasks.js
```diff
@@ -112,6 +112,20 @@ const getSelfTasks = async (req, res) => {
...
+
+const getTask = async (req, res) => {
+  try {
+    const taskId = req.params.id;
+    const { taskData } = await tasks.fetchTask(taskId);
+
+    if (!taskData) {
+      return res.boom.notFound("Task not found");
+    }
+    return res.json({ message: "task returned successfully", taskData });
```
ankushdharkar, line 124: Is 'taskData' reduntant. Can it just be `data`?

vinayak-trivedi: Yes, this can be changed, I did it because the function in the tasks model was returning `taskData` from there.

vinayak-trivedi: this will be fixed in #736

Maheima (approved):

PR #734, opened 2022-10-06
Task updated by SUPERUSER | APPOWNER on the task will be logged 

ivinayakg:
## How will this work?
- Will create a new log and save to the database about the event

## What has been done?
- called the `addlog` API to create a log for the event
- Response after a successful call
image


controllers/tasks.js
RitikJaiswal75, line 1: we have a log collection arent we using that here?

Issue #712, opened 2022-09-20
Show colored UI avatars for new users


vickydonor-99:
What is to be done ?
Changed ui-avatars of new members

PR #735, opened 2022-10-06
Feature/one time user colors

isVivek99:
**This is a one-time PR**

Issue Ticket Number:- #712 
API Contract: - API contract https://github.com/Real-Dev-Squad/website-api-contracts/pull/147
Frontend PR: https://github.com/Real-Dev-Squad/website-members/pull/396

Backend changes
- [x]  Yes
- [ ] No

Frontend Changes
- [ ]  Yes
- [x] No

Database changes
- [x]  Yes
- [ ] No

- [ ]  Yes
- [x] No

Deployment notes
None

Description
- This `PR` will add a colors id for each user which will then be used on frontend for users to see their profile is a specific color.

Testing Stats:
file: `controllers/userMigrations.js`

Screenshot 2023-08-15 at 11 15 11 AM

file: `models/userMigrations.js`

 Screenshot 2023-08-09 at 10 35 55 AM

file `utils/helpers.js`

Screenshot 2023-08-15 at 11 18 19 AM

### DIFF
--- a/utils/helper.js
+++ b/utils/helper.js
@@ -73,6 +73,14 @@ const getPaginatedLink = ({
  * @returns random Index number : index between the range 0 to array.length
  */
 const getRandomIndex = (maxLength = 10) => {
+  if (typeof maxLength !== "number") {
+    throw new Error("maxLength must be a number");
+  }
+
+  if (maxLength <= 0) {
+    throw new Error("maxLength must be a positive number");
+  }
+
   return Math.floor(Math.random() * maxLength);
 };
 


constants/cardColorArray.js
```diff
@@ -0,0 +1,12 @@
...
+  { color_primary: "#E9C46A", color_secondary: "#FFE094" },
+  { color_primary: "#F4A261", color_secondary: "#FFBF8C" },
+  { color_primary: "#2A9D8F", color_secondary: "#42E0CD" },
+  { color_primary: "#165692", color_secondary: "#3D8BD3" },
+  { color_primary: "#264653", color_secondary: "#387892" },
+  { color_primary: "#B93160", color_secondary: "#D75281" },
+  { color_primary: "#A5C9CA", color_secondary: "#E7F6F2" },
+  { color_primary: "#6E85B7", color_secondary: "#B2C8DF" },
+  { color_primary: "#FBA1A1", color_secondary: "#FBC5C5" },
+];
```
pallabez, line 12: Shouldn't it be UPPER_CASE for constants?

models/users.js
```diff
@@ -188,6 +189,39 @@ const fetchUserImage = async (users) => {
...
+      }
+    }
+    await Promise.all(updateUserPromises);
+    return { count: migratedUsers.length, users: migratedUsers };
+  } catch (err) {
+    logger.error("Error adding default colors to users", err);
+    throw err;
+  }
+};
+
```
pallabez, line 224: Can we move this to `models/migration.js`?

utils/users.js
```diff
@@ -81,9 +81,19 @@ const getParticipantUserIds = async (participantArray) => {
...
 };
 
+/**
+ * Returns a random object from the array of colors to user
+ * @param cardColorArray {array} : color containing array
+ * @returns userColorIndex number : index between the range 0 to cardColorArray.length
+ */
+const addColorsProperty = (cardColorArray) => {
+  return Math.floor(Math.random() * (cardColorArray.length - 0) + 0);
+};
```
pallabez, line 91: Can we make it more generic? 

something like 
`const getRandomItem = (array = []) => { return ...}

models/users.js
```diff
@@ -188,6 +189,39 @@ const fetchUserImage = async (users) => {
...
+    usersSnapshot.forEach((doc) => usersArr.push({ id: doc.id, ...doc.data() }));
+
+    for (const user of usersArr) {
+      const colors = user.colors ? user.colors : {};
+      if (user.colors === undefined) {
+        const userColorIndex = addColorsProperty(cardColorArray);
+
+        colors.primary_color = cardColorArray[parseInt(userColorIndex)].color_primary;
+        colors.secondary_color = cardColorArray[parseInt(userColorIndex)].color_secondary;
```
pallabez, line 212: Instead of setting color code directly here, can we store id that relates to that color.

Eg - In constant file we can have -
```
module.exports = [
  { color_id: 1, color_primary: "#DB1212", color_secondary: "#F88181" },
   ...
 ```
 and do `colors.colorId = randomColor.color_id`

File: test/integration/users.test.js [hunk 5]
```diff
@@ -565,4 +569,48 @@ describe("Users", function () {
...
+          );
+
+          expect(migratedUsernames).to.not.include.any.members(
+            userWithColorProperty,
+            "Should not modify color property of user with color object"
+          );
+          return done();
+        });
+    });
+  });
```
pallabez, line 615: thanks for the test

utils/users.js
vickydonor-99: why should we make it generic, when we have a specific use case for the function?

pallabez: Good point out. We are making it generic as to be able to use it in other places in future .

models/users.js
vickydonor-99: i got the part where we use id, but
`colors.colorId = randomColor.color_id`
would just add a property to colors like this:
`{colorId: 1}`

pallabez: Yes that's what we want. So in frontend, we assign colors to `colorId`. 

Let's say we want to change users with color red to blue. By the original approach, we need to update color of every user with color red. Using color_id we can change color in one line in frontend.

vickydonor-99: oh, ok I get it, so does it make sense to mantain a colorArray at the backend as well. I think we would have to mantain the array on both FE and BE with this approach, wdyt?

pallabez: Ideally this should be stored in a remote config. Eg - Firebase remote config.

But for now we can use keep it in frontend only, if needed make a api, `/colors` for the color array.

vickydonor-99: I have made the change, Can we approve the PR?

utils/users.js
```diff
@@ -81,9 +81,19 @@ const getParticipantUserIds = async (participantArray) => {
...
 
+/**
+ * Returns a random object from the array of colors to user
+ * @param array {array} : array containing objects
+ * @returns random Index number : index between the range 0 to array.length
+ */
+const getRandomIndex = (array = []) => {
+  return Math.floor(Math.random() * (array.length - 0) + 0);
+};
+
```
pallabez, line 92: Can we put this function on a more generic file? Such as `utils/helper.js`

constants/cardColorArray.js
```diff
@@ -0,0 +1,12 @@
...
+  { COLOR_ID: 2, COLOR_PRIMARY: "#F4A261", COLOR_SECONDARY: "#FFBF8C" },
+  { COLOR_ID: 3, COLOR_PRIMARY: "#2A9D8F", COLOR_SECONDARY: "#42E0CD" },
+  { COLOR_ID: 4, COLOR_PRIMARY: "#165692", COLOR_SECONDARY: "#3D8BD3" },
+  { COLOR_ID: 5, COLOR_PRIMARY: "#E9C46A", COLOR_SECONDARY: "#FFE094" },
+  { COLOR_ID: 6, COLOR_PRIMARY: "#264653", COLOR_SECONDARY: "#387892" },
+  { COLOR_ID: 7, COLOR_PRIMARY: "#B93160", COLOR_SECONDARY: "#D75281" },
+  { COLOR_ID: 8, COLOR_PRIMARY: "#A5C9CA", COLOR_SECONDARY: "#E7F6F2" },
+  { COLOR_ID: 9, COLOR_PRIMARY: "#6E85B7", COLOR_SECONDARY: "#B2C8DF" },
+  { COLOR_ID: 10, COLOR_PRIMARY: "#FBA1A1", COLOR_SECONDARY: "#FBC5C5" },
+];
```
pallabez, line 12: We are not using COLOR_PRIMARY or COLOR_SECONDARY, can we remove it?

pallabez:
Other than the 2 comments above, everything looks good to me. We can proceed with merging after those are resolved. Thanks for the PR.

constants/cardColorArray.js
vickydonor-99: we would be using `/colors` api, in the future, wouldn't we require it then?

utils/users.js
vickydonor-99: should i make a `migrations.js` file?

pallabez: getRandomIndex is a generic function that can be used anywhere. I don't think migrations.js would be suitable here.

vickydonor-99: got it, made the changes

pallabez (approved):
LGTM

utils/helpers.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Returns a random object from the array of colors to user
+ * @param array {array} : array containing objects
+ * @returns random Index number : index between the range 0 to array.length
+ */
+const getRandomIndex = (array = []) => {
+  return Math.floor(Math.random() * (array.length - 0) + 0);
```
ankushdharkar, line 7: Lot of zero operations :D
What value do they add?

harshith-venkatesh: I am curious now too😅

models/migrations.js
```diff
@@ -0,0 +1,40 @@
...
+    const usersArr = [];
+
+    usersSnapshot.forEach((doc) => usersArr.push({ id: doc.id, ...doc.data() }));
+
+    for (const user of usersArr) {
+      const colors = user.colors ? user.colors : {};
+      if (user.colors === undefined) {
+        const userColorIndex = getRandomIndex(cardColorIdArray);
+
+        colors.color_id = cardColorIdArray[parseInt(userColorIndex)].COLOR_ID;
```
harshith-venkatesh, line 24: Would it be better to handle parseInt or whether index is number in function itself?

vickydonor-99: actually I am returning a number , but there is an es-lint issue which was popping up

utils/helpers.js
vickydonor-99: doesn't add any value , will remove it

Maheima (approved):
LGTM!

rupal-2505 (dismissed):
LGTM!

test/integration/userMigrations.test.js
```diff
@@ -0,0 +1,81 @@
...
+const app = require("../../server");
+const authService = require("../../services/authService");
+const addUser = require("../utils/addUser");
+const cleanDb = require("../utils/cleanDb");
+// Import fixtures
+const userData = require("../fixtures/user/user")();
+const superUser = userData[4];
+const nonSuperUser = userData[0];
+const userWithColorProperty = [userData[5].username];
+const colorBearingUsernames = [superUser.username, nonSuperUser.username];
```
ankushdharkar, line 14: Why do these two type of users bear colors?

ankushdharkar (changes requested):

models/userMigrations.js
```diff
@@ -0,0 +1,40 @@
...
+      const colors = user.colors ? user.colors : {};
+      if (user.colors === undefined) {
+        const userColorIndex = getRandomIndex(cardColorIdArray);
+
+        colors.color_id = cardColorIdArray[parseInt(userColorIndex)].COLOR_ID;
+        updateUserPromises.push(userModel.doc(user.id).set({ ...user, colors }));
+        migratedUsers.push(user.username);
+      }
+    }
+
```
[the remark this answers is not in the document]
alishvagadiya: We can use transactions in firestore.

Link for reference - https://firebase.google.com/docs/firestore/manage-data/transactions

test/fixtures/user/user.js
```diff
@@ -144,6 +144,9 @@ module.exports = () => {
         app_owner: true,
         archived: true,
       },
+      colors: {
+        color_id: 2,
```
prakashchoudhary07, line 148: If its already a color object, then we can call it id itself instead of color_id?

controllers/userMigrations.js
```diff
@@ -0,0 +1,42 @@
...
+      if (user.colors === undefined) {
+        const userColorIndex = getRandomIndex(userColors);
+        colors.color_id = userColorIndex;
+        updateUserPromises.push(userModel.doc(user.id).set({ ...user, colors }));
+        migratedUsers.push(user.username);
+      }
+    }
+
+    await Promise.all(updateUserPromises);
+
```
prakashchoudhary07, line 29: What do you think on considering using firestore batched writes or transactions

https://firebase.google.com/docs/firestore/manage-data/transactions

isVivek99: yes makes sense

test/fixtures/user/user.js
```diff
@@ -144,6 +144,9 @@ module.exports = () => {
...
+      },
```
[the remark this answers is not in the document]
isVivek99: not sure ATM, we can make it just color_id, but if we are going to add more colors in the future then this would be better,  should i change the property?

models/userMigrations.js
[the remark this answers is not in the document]
isVivek99: updated with batch updates

controllers/userMigrations.js
isVivek99: updated with batch

test/fixtures/user/user.js
isVivek99: resolving this duplicate comment

iamitprakash (dismissed):

utils/helpers.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Returns a random object from the array of colors to user
+ * @param array {array} : array containing objects
+ * @returns random Index number : index between the range 0 to array.length
+ */
+const getRandomIndex = (arrayLength = 10) => {
+  return Math.floor(Math.random() * arrayLength);
+};
```
heyrandhir, line 8: this will return 1 less than array length as we are using math.floor. do we need even the upper bound i.e 10 here. if yes we should use `math.ceil`

controllers/userMigrations.js
```diff
@@ -0,0 +1,57 @@
...
+        totalCount++;
+        users.push(user.username);
+        if (operationCounter === MAX_TRANSACTION_WRITES) {
+          batchArray.push(firestore.batch());
+          batchIndex++;
+          operationCounter = 0;
+        }
+      }
+    }
+    batchArray.forEach(async (batch) => await batch.commit());
```
heyrandhir, line 43: this seems to be business logic. should better be handled in the model instead of controller

controllers/userMigrations.js
```diff
@@ -0,0 +1,57 @@
...
+
+    return res.json({
+      message: "User colors updated successfully!",
+      usersDetails: { count: totalCount, users },
+    });
```
heyrandhir, line 48: we are returning the entire users from the db. this will be huge. I think returning the no of users updated. no of users unaffected and other metrics should be fine.

heyrandhir (changes requested):
Can you also update the API Contracts and Data Model as well in the ticket.

controllers/userMigrations.js
```diff
@@ -0,0 +1,57 @@
...
+    let batchIndex = 0;
+    let totalCount = 0;
+
+    for (const user of usersArr) {
+      const colors = user.colors ?? {};
+      if (user.colors === undefined) {
+        const userColorIndex = getRandomIndex(USER_COLORS);
+        colors.color_id = userColorIndex;
+        const docId = userModel.doc(user.id);
+        batchArray[parseInt(batchIndex)].set(docId, { ...user, colors });
```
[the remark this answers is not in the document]
isVivek99: yes we are batch updating, 
docs for reference
https://firebase.google.com/docs/firestore/manage-data/transactions

isVivek99:
> Can you also update the API Contracts and Data Model as well in the ticket.

yes added

controllers/userMigrations.js
isVivek99: ok makes sense will update this

models/userMigrations.js
```diff
@@ -0,0 +1,25 @@
+const firestore = require("../utils/firestore");
+const userModel = firestore.collection("users");
+
+/**
+ * Adds default colors property
+ * @return {Promise<usersMigrated|Object>}
+ */
+
+const addDefaultColors = async () => {
```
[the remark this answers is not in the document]
isVivek99: i had all the logic inside the function, can you check this comment once
https://github.com/Real-Dev-Squad/website-backend/pull/735#discussion_r1014571527,  based on this I moved it out

utils/helpers.js
isVivek99: if we have an array of length 10, then we can only index it upto arr[9], thats why  `math.floor`

controllers/userMigrations.js
isVivek99: i had all the logic inside the function, can you check this comment once
https://github.com/Real-Dev-Squad/website-backend/pull/735#discussion_r1014571527, based on this I moved it out

controllers/userMigrations.js
isVivek99: what all other metrics should i send?

models/userMigrations.js
```diff
@@ -0,0 +1,25 @@
+const firestore = require("../utils/firestore");
+const userModel = firestore.collection("users");
+
+/**
+ * Adds default colors property
+ * @return {Promise<usersMigrated|Object>}
```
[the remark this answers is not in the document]
isVivek99: this was based on some old code, pushing updates

models/userMigrations.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+const addDefaultColors = async () => {
+  try {
+    const usersSnapshot = await userModel.get();
+    const usersArr = [];
+
+    usersSnapshot.forEach((doc) => usersArr.push({ id: doc.id, ...doc.data() }));
+
+    return { usersArr, userModel };
```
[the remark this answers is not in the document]
isVivek99: done

controllers/userMigrations.js
```diff
@@ -0,0 +1,57 @@
+const firestore = require("../utils/firestore");
+const userQuery = require("../models/userMigrations");
+const logger = require("../utils/logger");
```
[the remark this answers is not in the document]
isVivek99: what exactly are we talking about?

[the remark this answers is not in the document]
isVivek99: if it is logger, it is being used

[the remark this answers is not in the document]
heyrandhir: I think its a global variable, in other files it isnt being imported before using.

controllers/userMigrations.js
heyrandhir: something of the sort total no of users fetched for this out of which how many were updated and similarly other details whatever you feel relevant.

controllers/userMigrations.js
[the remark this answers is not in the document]
isVivek99: got it

routes/userMigrations.js
```diff
@@ -0,0 +1,10 @@
+const express = require("express");
+const router = express.Router();
+const authenticate = require("../middlewares/authenticate");
+const authorizeRoles = require("../middlewares/authorizeRoles");
+const { SUPERUSER } = require("../constants/roles");
+const migrations = require("../controllers/userMigrations");
+
+router.patch("/addDefaultColorProperty", authenticate, authorizeRoles([SUPERUSER]), migrations.addDefaultColors);
```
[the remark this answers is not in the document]
isVivek99: updated now

controllers/userMigrations.js
isVivek99: updated now

routes/userMigrations.js
```diff
@@ -0,0 +1,10 @@
+const express = require("express");
+const router = express.Router();
+const authenticate = require("../middlewares/authenticate");
+const authorizeRoles = require("../middlewares/authorizeRoles");
+const { SUPERUSER } = require("../constants/roles");
+const migrations = require("../controllers/userMigrations");
+
+router.patch("/userDefaultColor", authenticate, authorizeRoles([SUPERUSER]), migrations.addDefaultColors);
```
heyrandhir, line 8: we should typically use lowercase words separated by hyphens in route names instead of came case

controllers/userMigrations.js
```diff
@@ -0,0 +1,61 @@
...
+        totalCount++;
+        users.push(user.username);
+        if (operationCounter === MAX_TRANSACTION_WRITES) {
+          batchArray.push(firestore.batch());
+          batchIndex++;
+          operationCounter = 0;
+        }
+      }
+    }
+    batchArray.forEach(async (batch) => await batch.commit());
```
heyrandhir, line 43: I believe the batch operation will interact with Firestore, and it seems more appropriate to handle it within the models rather than the controller. as we are updating the docs in our DB. I'm still not convinced it should be placed in the controller. but still will rest the decision to you.

heyrandhir (changes requested):
we are modifying user resource here ideally this route should be in users.

controllers/userMigrations.js
heyrandhir: As per @ankushdharkar - `The model file should only be responsible for interactions with firestore. The manipulation should take place in controller file, if I am not wrong.` Don't you feel the batch operation is a  interaction with firestore @isVivek99 .

isVivek99: yes the batch is an interaction with firestore, moving it to models then

routes/userMigrations.js
isVivek99: ok thanks, i will update this

isVivek99:
> we are modifying user resource here ideally this route should be in users.

this was discussed with @pallabez , again the discussion took place long time ago and it was decided to use this route for migrations. It is a one time PR so the code will not exist after this. Let me know what should be done here

heyrandhir:
> > we are modifying user resource here ideally this route should be in users.
> 
> this was discussed with @pallabez , again the discussion took place long time ago and it was decided to use this route for migrations. It is a one time PR so the code will not exist after this. Let me know what should be done here

so do you mean that for new users these random values are being added. only for the old users this is a 1 time script ?

isVivek99:
> > > we are modifying user resource here ideally this route should be in users.
> > 
> > 
> > this was discussed with @pallabez , again the discussion took place long time ago and it was decided to use this route for migrations. It is a one time PR so the code will not exist after this. Let me know what should be done here
> 
> so do you mean that for new users this random color constant are being added. only for the old users this is a 1 time fix ?

this will only be for existing users, I am not sure if a new user signs up we are adding them a color property

heyrandhir:
> > > > we are modifying user resource here ideally this route should be in users.
> > > 
> > > 
> > > this was discussed with @pallabez , again the discussion took place long time ago and it was decided to use this route for migrations. It is a one time PR so the code will not exist after this. Let me know what should be done here
> > 
> > 
> > so do you mean that for new users this random color constant are being added. only for the old users this is a 1 time fix ?
> 
> this will only be for existing users, I am not sure if a new user signs up we are adding them a color property, we can also have a follow up `PR` for that

In this scenario, it seems more logical to write that part initially otherwise, the current script will function for all the current users, but the color property will be missing for new users and the super user will need to run this every time new users are added in the DB. Alternatively, if we add it for new users first, then we can ensure that everyone is covered. Once the script is executed, we can then remove the script for once and for all. What do you think @vivek-geekyants @isVivek99

test/integration/userMigrations.test.js
```diff
@@ -0,0 +1,71 @@
...
+        .set("cookie", `${cookieName}=${superUserAuthToken}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+
+          expect(res).to.have.status(200);
+          expect(res.body.usersDetails.totalUsersFetched).to.be.equal(colorBearingUsernames.length);
+          expect(res.body.usersDetails.totalUsersUpdated).to.be.equal(colorBearingUsernames.length);
+          expect(res.body.usersDetails.totalUsersUnaffected).to.be.equal(0);
```
heyrandhir, line 66: can we also write a test in which we test users who don't have the color property pre-existing are not affected.

heyrandhir (changes requested):
Just noticed model tests are missing 👀

test/integration/userMigrations.test.js
isVivek99: sure , i had that check, but after removing usernames i removed it, ill add it back in some other. way

isVivek99:
> Just noticed model tests are missing 👀

will be updating

isVivek99:
yes will raise a seperate `PR` for that, lets not stop this PR being approved for that.

test/integration/users.test.js
```diff
@@ -50,6 +50,7 @@ describe("Users", function () {
     userId = await addUser();
     jwt = authService.generateAuthToken({ userId });
     superUserId = await addUser(superUser);
+
```
heyrandhir, line 53: Nit : if this file we are not changing we can remove this unwanted change.

test/integration/userMigrations.test.js
heyrandhir: @isVivek99 still can't see the assertion for users who have the color property pre-existing and are not affected.

heyrandhir (changes requested):

models/userMigrations.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const addDefaultColors = async (batchSize = MAX_TRANSACTION_WRITES) => {
+  try {
+    if (batchSize > MAX_TRANSACTION_WRITES) {
+      throw new Error(`Error cannot add more than ${batchSize} users at once .`);
+    }
```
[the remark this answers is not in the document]
isVivek99: if user passes a higher batch size that what is expected then we throw an error beforehand

controllers/userMigrations.js
```diff
@@ -0,0 +1,27 @@
+const userQuery = require("../models/userMigrations");
+const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
+const MAX_TRANSACTION_WRITES = 499;
```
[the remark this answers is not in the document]
isVivek99: we can control batch sizes if we pass through controller

models/userMigrations.js
```diff
@@ -0,0 +1,63 @@
...
+    let batchIndex = 0;
+    let totalCount = 0;
+
+    for (const user of usersArr) {
+      const colors = user.colors ?? {};
+      if (!user.colors) {
+        const userColorIndex = getRandomIndex(USER_COLORS);
+        colors.color_id = userColorIndex;
+        const docId = userModel.doc(user.id);
+        batchArray[parseInt(batchIndex)].set(docId, { ...user, colors });
```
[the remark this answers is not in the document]
isVivek99: agree, will update

test/integration/userMigrations.test.js
isVivek99: i have added a similar test in the model test.
[test/unit/models/userMigrations.test.js](https://github.com/Real-Dev-Squad/website-backend/pull/735/files#diff-706d009a498138cfe882b0ee3428b8357316e116c8f4866382eb4ea1f328d6da)
does this suffice?

models/userMigrations.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param res {Object} - Express response object
+ */
+
+const addDefaultColors = async (batchSize = MAX_TRANSACTION_WRITES) => {
+  try {
+    if (batchSize > MAX_TRANSACTION_WRITES) {
+      throw new Error(`Error cannot add more than ${batchSize} users at once .`);
+    }
+    const usersSnapshot = await userModel.get();
+    const usersArr = [];
```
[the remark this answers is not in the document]
isVivek99: i thought we can only access user through id or usernames, ill check it

models/userMigrations.js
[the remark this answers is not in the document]
isVivek99: i got it, updating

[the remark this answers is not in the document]
isVivek99: also i am passing batchSize as a parameter for testing if batch works correctly , using tests

models/userMigrations.js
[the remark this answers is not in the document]
isVivek99: I have used the dataAccessLayer for this, but the layer has only provision to fetch un-archived users, but we want to get all users, what should we do in this case?

test/integration/userMigrations.test.js
heyrandhir: > i have added a similar test in the model test. [test/unit/models/userMigrations.test.js](https://github.com/Real-Dev-Squad/website-backend/pull/735/files#diff-706d009a498138cfe882b0ee3428b8357316e116c8f4866382eb4ea1f328d6da) does this suffice?

it would be great if we could add in integration test. This expansion in coverage will enhance our confidence in the system. 😊

models/userMigrations.js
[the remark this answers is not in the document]
heyrandhir: Your current code should suffice for now. We can proceed without blocking your pull request. Let's create a ticket to keep track of this task, and when you find the time and bandwidth, we can address it after this current work. 

CC @mailmesriza98 @prakashchoudhary07

models/userMigrations.js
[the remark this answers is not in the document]
heyrandhir: Still not resolved @isVivek99 sir 🥲

routes/index.js
```diff
@@ -19,6 +19,7 @@ app.use("/users/status", require("./userStatus.js"));
 app.use("/users", require("./users.js"));
 app.use("/profileDiffs", require("./profileDiffs.js"));
 app.use("/wallet", require("./wallets.js"));
+app.use("/migrations", require("./userMigrations.js"));
```
heyrandhir, line 22: if you are aware @ankushdharkar has strongly advised against approving pull requests that don't adhere to the correct naming convention. As `migrations` isn't a resource, could you kindly consider relocating this route under `\users` ?

heyrandhir (changes requested):

models/userMigrations.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const addDefaultColors = async (batchSize = MAX_TRANSACTION_WRITES) => {
+  try {
+    const usersSnapshotArr = await dataAccess.retrieveUsers({ query: { size: MAX_USERS_SIZE } });
+    const usersArr = usersSnapshotArr.users;
+    // usersSnapshot.users.forEach((doc) => usersArr.push({ id: doc.id, ...doc }));
```
heyrandhir, line 20: please remove the commented code if not needed !

test/unit/models/userMigrations.test.js
```diff
@@ -0,0 +1,52 @@
...
+    expect(response.totalUsersUnaffected).to.equal(1);
+  });
+  it("should not affect users already having color property", async function () {
+    // Manually add a color property to a user
+    const userId = await addUser(userData[0]);
+    await userModel.doc(userId).update({ colors: { color_id: 3 } });
+    const response = await userMigrationModel.addDefaultColors(MAX_TRANSACTION_WRITES);
+    expect(response.totalUsersFetched).to.equal(6);
+    expect(response.totalUsersUpdated).to.equal(4);
+    expect(response.totalUsersUnaffected).to.equal(2);
```
heyrandhir, line 46: we added color property to 1 user but why are 2 users unaffected 🤔

utils/helpers.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Returns a random object from the array of colors to user
+ * @param array {array} : array containing objects
+ * @returns random Index number : index between the range 0 to array.length
+ */
+const getRandomIndex = (arrayLength = 10) => {
```
heyrandhir, line 6: Nit: we call the function parameter `maxIndex` so that parameter represents the maximum index value that the random index should not exceed.

heyrandhir, line 6: since this is a helper function and might be used later. is negative number also permissible ?

isVivek99: nope it's not permissible

test/unit/models/userMigrations.test.js
isVivek99: we already have a user with color property who we are adding to the DB

utils/helpers.js
isVivek99: updated

isVivek99: updated now, does this work?

PR #744, opened 2022-10-20
Accepting redirectURL and mobile=true as query params

prakashchoudhary07:
`/auth/github/callback` route can now accept `redirectURL=url_to_redirect_to` and `mobile=true`

For redirecting to a particular page, `redirectURL=https://realdevsquad.com` query param can be passed
ex: `/auth/github/callback?redirectURL=https://realdevsquad.com`

And to let backend know weather the request is being made from mobile, `mobile=true` query param cab be passed
ex: `/auth/github/callback?mobile=true`


prakashchoudhary07:
This use case has been solved by other PR

PR #746, opened 2022-10-21
Refactor/change query fields

### LINKED ISSUE #745: change the firebase query of assign task middleware
- Previously it was decided that we will store level and category of a task in a object named `taskLevel` as follows

```
taskLevel: {
  category: react,
  level: 2
}
```

but it makes more sense to store `category` and `level` separately on every tasks instead of putting it in an object,

- Our middleware currently makes a query based on the Object data type, we need to change that

vinayak-trivedi:
closes #745


middlewares/validators/tasks.js
```diff
@@ -77,6 +84,13 @@ const updateTask = async (req, res, next) => {
           [NEELAM]: joi.number().optional(),
         })
         .optional(),
+      taskLevel: joi
+        .object()
+        .keys({
+          category: joi.string().optional(),
+          level: joi.number().optional(),
+        })
+        .optional(),
```
rohan09-raj, line 93: This should also be fixed as we are not using `taskLevel` object anymore, category and level should be individually validated.

vinayak-trivedi: I don't know why it's there, I pulled Muralidhar's PR some days back so I think when I created a new branch, those changes still were there, let me remove it

rohan09-raj (dismissed):
LGTM 💯 💯

middlewares/assignTask.js
```diff
@@ -4,7 +4,7 @@ const tasks = firestore.collection("tasks");
 
 const assignTask = async function (req, res) {
   try {
-    const { task } = await fetchSkillLevelTask("frontend", 1);
+    const { task } = await fetchSkillLevelTask("FRONTEND", 1);
```
shubham-y, line 7: Can you please explain why are we passing skill `FRONTEND` by default?

shubham-y: Also, I think it will be better to import it to from constants

vinayak-trivedi: it is hard coded for now, it will come from the frontend when we have skills in the user object

shubham-y: Got it. As discussed, please add a comment in code for now

shubham-y (dismissed):
LGTM 👍

middlewares/assignTask.js
```diff
@@ -4,7 +4,8 @@ const tasks = firestore.collection("tasks");
 
 const assignTask = async function (req, res) {
   try {
-    const { task } = await fetchSkillLevelTask("frontend", 1);
+    // this hardcoded value will be removed once we have user skill
```
ankushdharkar, line 7: Should this be having a ticket @vinayak-trivedi?

models/tasks.js
```diff
@@ -195,9 +195,9 @@ const fetchUserTasks = async (username, statuses = [], field, order) => {
 
 const getNewTask = async (skill, level) => {
   const task = await tasksModel
-    .where("taskLevel.category", "==", skill)
-    .where("taskLevel.level", ">=", level)
-    .where("taskLevel.level", "<=", level + 2)
+    .where("category", "==", skill)
+    .where("level", ">=", level)
+    .where("level", "<=", level + 2)
     .where("status", "==", "AVAILABLE")
```
ankushdharkar, line 201: These values should ideally come from a constants file

PR #747, opened 2022-10-22
added available from constant file

vinayak-trivedi:
Added available status from a constant file, instead of using the hardcoded value


ankushdharkar:
Thanks for the cleanup

Issue #749, opened 2022-10-23
Join Section | Fix naming model and keys


RitikJaiswal75:
Fix the namings and keys of join model

PR #750, opened 2022-10-23
fix model name and key names

RitikJaiswal75:
Fix model name for join section and key names for data.

closes : #749


ankushdharkar (approved):
Nice. Thank you for the quick turnaround.

PR #738, opened 2022-10-12
Idle users and members

SanketDhabarde:
### Assumption:
- Idle users are those who have `"status": "idle"` property 

### API Contract PR:
https://github.com/Real-Dev-Squad/website-api-contracts/pull/93

### Endpoint: (Public)
1. [GET] `/users/idle ` : It will return all users who are idle
   response : 
   ```json
   {
    "message": "Idle users returned successfully!",
    "idleUserUsernames": [
        "vrushalisjoshi",
        "ashleytitus18"
    ]
   }
   ```
2. [GET] `/users/idle?members=true` : It will return only members who are idle
   response:
     ```json
    {
      "message": "Idle users returned successfully!",
      "idleUserUsernames": [
          "vrushalisjoshi"
       ]
    }
    ```    
### Subtasks
1. #767 
2. #766 
3. #754 
4. #755 

### The order of PR
https://github.com/Real-Dev-Squad/website-backend/pull/738#issuecomment-1297216403

PR #758, opened 2022-10-26
Create a new endpoint to add default status for users where it does not exists

### LINKED ISSUE #754: Create a new endpoint to add default status for users where it DNE
### Requirement:
There is inconsistency in the data of users. Some of the users don't have a `status` property. And to query a field in firestore, the field should exist in the document. 

Firestore Limitation: https://firebase.google.com/docs/firestore/query-data/queries#limitations

### AC:
- Endpoint : /users/defaultStatus
- Allowed only for the superuser
- This is a one-time-use endpoint and should be deleted after use  https://github.com/Real-Dev-Squad/website-backend/issues/755

### Use case:
The `status` field will be helpful to query the results based on the `status` of the user. Currently, we have an issue where we need to filter the idle users https://github.com/Real-Dev-Squad/website-backend/pull/738

SanketDhabarde:
### Assumption:
Default `status` is "**idle**" for the user where it does not exists.

### Endpoint: (Only superuser is authorized)
[PATCH] `/users/defaultStatus ` : It will return usernames of user for which default status ("idle") is added
   response : 
   ```json
   {
    "message": "Default status added to users successfully!",
    "count": 2,
    "updatedUsers": [
        "vrushalisjoshi",
        "ashleytitus18"
    ]
   }
   ```

### The order of PR
https://github.com/Real-Dev-Squad/website-backend/pull/738#issuecomment-1297216403

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -375,10 +375,10 @@ const addDefaultArchivedRole = async (req, res) => {
  */
 const addDefaultStatus = async (req, res) => {
   try {
-    const addedDefaultStatusUsers = await userQuery.addDefaultStatus();
+    const updatedUsersData = await userQuery.addDefaultStatus();
     return res.json({
       message: "Default status added to users successfully!",
-      addedDefaultStatusUsers,
+      ...updatedUsersData,
     });
   } catch (error) {
     logger.error(`Error adding default status: ${error}`);
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -366,6 +366,26 @@ const addDefaultArchivedRole = async (req, res) => {
   }
 };
 
+/**
+ * Add default status to users where it does not exists
+ *
+ * @param {Object} req - Express request object
+ * @param {Object} res - Express response object
+ * @returns usernames of users for which default status is added
+ */
+const addDefaultStatus = async (req, res) => {
+  try {
+    const addedDefaultStatusUsers = await userQuery.addDefaultStatus();
+    return res.json({
+      message: "Default status added to users successfully!",
+      addedDefaultStatusUsers,
+    });
+  } catch (error) {
+    logger.error(`Error adding default status: ${error}`);
+    return res.boom.badImplementation("Something went wrong. Please contact admin");
+  }
+};
+
 module.exports = {
   verifyUser,
   generateChaincode,
@@ -382,4 +402,5 @@ module.exports = {
   addUserIntro,
   getUserIntro,
   addDefaultArchivedRole,
+  addDefaultStatus,
 };
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -297,26 +297,26 @@ const addUserIntro = async (req, res) => {
     const rawData = req.body;
     const data = {
       userId: req.userData.id,
-      personalDetails: {
+      biodata: {
         firstName: rawData.firstName,
         lastName: rawData.lastName,
       },
-      locationDetails: {
+      location: {
         city: rawData.city,
         state: rawData.state,
         country: rawData.country,
       },
-      professionalDetails: {
+      professional: {
         institution: rawData.college,
         skills: rawData.skills,
       },
-      PersonalIntroWithReason: {
+      intro: {
         introduction: rawData.introduction,
         funFact: rawData.funFact,
         forFun: rawData.forFun,
         whyRds: rawData.whyRds,
       },
-      heardFrom: rawData.heardAbout,
+      foundFrom: rawData.foundFrom,
     };
     await userQuery.addJoinData(data);
 
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -371,7 +371,7 @@ const addDefaultArchivedRole = async (req, res) => {
  *
  * @param {Object} req - Express request object
  * @param {Object} res - Express response object
- * @returns usernames of users for which default status is added
+ * @returns count and usernames of users for which default status is added
  */
 const addDefaultStatus = async (req, res) => {
   try {
--- a/models/users.js
+++ b/models/users.js
@@ -3,7 +3,7 @@
  * This will contain the DB schema if we start consuming an ORM for managing the DB operations
  */
 const walletConstants = require("../constants/wallets");
-const { userStatus } = require("../constants/users");
+const { USER_STATUS } = require("../constants/users");
 
 const firestore = require("../utils/firestore");
 const { fetchWallet, createWallet } = require("../models/wallets");
@@ -233,7 +233,7 @@ const addDefaultStatus = async () => {
       if (user.status === undefined) {
         updateStatusPromises.push(
           userModel.doc(user.id).update({
-            status: userStatus.IDLE,
+            status: USER_STATUS.IDLE,
           })
         );
         updatedUsers.push(user.username);
--- a/models/users.js
+++ b/models/users.js
@@ -7,7 +7,7 @@ const walletConstants = require("../constants/wallets");
 const firestore = require("../utils/firestore");
 const { fetchWallet, createWallet } = require("../models/wallets");
 const userModel = firestore.collection("users");
-const joinModel = firestore.collection("joining_data");
+const joinModel = firestore.collection("applicants");
 
 /**
  * Adds or updates the user data
--- a/models/users.js
+++ b/models/users.js
@@ -230,7 +230,7 @@ const addDefaultStatus = async () => {
 
     for (const user of users) {
       if (user.status === undefined) {
-        promisesToUpdateStatus.push(
+        updateStatusPromises.push(
           userModel.doc(user.id).update({
             status: "idle",
           })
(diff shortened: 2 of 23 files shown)


controllers/users.js
```diff
@@ -366,6 +366,26 @@ const addDefaultArchivedRole = async (req, res) => {
...
+/**
+ * Add default status to users where it does not exists
+ *
+ * @param {Object} req - Express request object
+ * @param {Object} res - Express response object
+ * @returns usernames of users for which default status is added
+ */
+const addDefaultStatus = async (req, res) => {
+  try {
+    const addedDefaultStatusUsers = await userQuery.addDefaultStatus();
```
samarpan1738, line 378: Should we be worried about any potential timeouts here?

models/users.js
```diff
@@ -215,6 +215,35 @@ const fetchUserImage = async (users) => {
...
 
+/**
+ * Add default status
+ * @returns {Promise<string[]>}
+ */
+const addDefaultStatus = async () => {
+  try {
+    const usersSnapshot = await userModel.get();
+    const users = [];
+    const addedDefaultStatusUsers = [];
```
samarpan1738, line 226: IMO migratedUsers or updatedUsers would be a better variable name

models/users.js
```diff
@@ -215,6 +215,35 @@ const fetchUserImage = async (users) => {
...
+
+    usersSnapshot.forEach((doc) => users.push({ id: doc.id, ...doc.data() }));
+
+    for (const user of users) {
+      if (user.status === undefined) {
```
samarpan1738, line 231:
```suggestion
      if (!user.status) {
```

models/users.js
```diff
@@ -215,6 +215,35 @@ const fetchUserImage = async (users) => {
...
+        const userDoc = userModel.doc(user.id);
+        await userDoc.update({
+          status: "idle",
+        });
```
samarpan1738, line 235: No need of making `userDoc` variable since we aren't using it anywhere else
```suggestion
        await userModel.doc(user.id).update({
          status: "idle",
        });
```

routes/users.js
```diff
@@ -18,6 +18,9 @@ router.get("/:username", users.getUser);
 router.get("/:userId/intro", authenticate, authorizeRoles([SUPERUSER]), users.getUserIntro);
 router.post("/self/intro", authenticate, userValidator.validateJoinData, users.addUserIntro);
 
+// one-time-use endpoint
+router.patch("/add-default-status", authenticate, authorizeRoles([SUPERUSER]), users.addDefaultStatus);
```
samarpan1738, line 22: Route can be named better

test/integration/users.test.js
```diff
@@ -186,6 +186,68 @@ describe("Users", function () {
     });
   });
 
+  describe("PATCH /add-default-status", function (done) {
+    beforeEach(async function () {
+      await addUser(userData[1]); // status does not exists user
+      await addUser(userData[0]); // status active user
```
samarpan1738, line 192: Create variables with meaningful names rather than comments
Comments should only be used to explain why rather than what

models/users.js
SanketDhabarde: But we are checking if the field exists or not. with your suggested case it will take all falsy values. And when the property does not exist in an object it will return `undefined`.

test/integration/users.test.js
```diff
@@ -186,6 +186,68 @@ describe("Users", function () {
...
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Default status added to users successfully!");
+          expect(res.body.addedDefaultStatusUsers).to.include.members(["nikhil"]);
+          expect(res.body.addedDefaultStatusUsers).to.not.include.members(["ankur"]);
```
samarpan1738, line 209: Use the same user objects here which were added in firestore in the beforeEach block.
It'll be more readable that way

Reference: https://github.com/Real-Dev-Squad/website-backend/pull/654/files#diff-a07acf94413fffcf0e2831b131eb44a057a498ed297a72190c8f23fc6561b50e

models/users.js
```diff
@@ -215,6 +215,35 @@ const fetchUserImage = async (users) => {
...
+        addedDefaultStatusUsers.push(user.username);
+      }
+    }
+
+    return addedDefaultStatusUsers;
```
samarpan1738, line 239: It would be helpful to return the count of updated users as well.

models/users.js
samarpan1738: If we are sure status will always be undefined then your approach is correct
Btw is there any problem in checking for all falsy values?

models/users.js
```diff
@@ -215,6 +215,34 @@ const fetchUserImage = async (users) => {
...
+    const users = [];
+    const addedDefaultStatusUsers = [];
+
+    usersSnapshot.forEach((doc) => users.push({ id: doc.id, ...doc.data() }));
+
+    for (const user of users) {
+      if (user.status === undefined) {
+        await userModel.doc(user.id).update({
+          status: "idle",
+        });
```
samarpan1738, line 234: Currently this line of code will wait for a user doc to be updated and then move on to update the next doc
Time taken to update all users = no. of users to be updated * time taken to update 1 user

This can be improved by updating the docs in parallel

models/users.js
SanketDhabarde: `updatedUsers` will be better. As from route, it will be clear what is updated for these users.

SanketDhabarde:
@samarpan1738 made the requested changes. Can you please check?

test/integration/users.test.js
```diff
@@ -11,6 +11,8 @@ const cleanDb = require("../utils/cleanDb");
 const userData = require("../fixtures/user/user")();
 const profileDiffData = require("../fixtures/profileDiffs/profileDiffs")();
 const superUser = userData[4];
+const statusDoesNotExistsUser = userData[1];
```
samarpan1738, line 14:
```suggestion
const userWithoutStatusField = userData[1];
```

test/integration/users.test.js
```diff
@@ -11,6 +11,8 @@ const cleanDb = require("../utils/cleanDb");
...
+const activeUser = userData[0];
```
samarpan1738, line 15:
```suggestion
const userWithActiveStatus = userData[0];
```

models/users.js
```diff
@@ -215,6 +215,39 @@ const fetchUserImage = async (users) => {
...
 
+/**
+ * Add default status
+ * @returns {Promise<string[]>}
+ */
+const addDefaultStatus = async () => {
+  try {
+    const usersSnapshot = await userModel.get();
+    const users = [];
+    const promisesToUpdateStatus = [];
```
samarpan1738, line 226: updateStatusPromises

samarpan1738, line 226:
```suggestion
    const updateStatusPromises = [];
```

models/users.js
```diff
@@ -215,6 +215,39 @@ const fetchUserImage = async (users) => {
   return images;
 };
 
+/**
+ * Add default status
+ * @returns {Promise<string[]>}
```
samarpan1738, line 220: Please update this based on the return value

SanketDhabarde: done

models/users.js
```diff
@@ -215,6 +215,39 @@ const fetchUserImage = async (users) => {
...
+    const updateStatusPromises = [];
+    const updatedUsers = [];
+
+    usersSnapshot.forEach((doc) => users.push({ id: doc.id, ...doc.data() }));
+
+    for (const user of users) {
+      if (user.status === undefined) {
+        updateStatusPromises.push(
+          userModel.doc(user.id).update({
+            status: "idle",
```
samarpan1738, line 235: We can use the user status enum here from the constants file

SanketDhabarde: Added constant enum for user status. As the existing one was an array of user statuses, so doing `userStatusEnum[1]` for idle doesn't make sense

samarpan1738: > As the existing one was an array of user statuses, so doing userStatusEnum[1] for idle doesn't make sense

Then we should change the userStatusEnum to be an object like [profileStatus](https://github.com/Real-Dev-Squad/website-backend/blob/4f9c29dcce231be09e12b886587e2be74c485ca0/constants/users.js#L2)

SanketDhabarde: Added userStatus enum in the current implementation. Using that enum in existing implementation can have a separate issue.

test/utils/deleteStatus.js
```diff
@@ -0,0 +1,34 @@
+const firestore = require("../../utils/firestore");
+const userModal = firestore.collection("users");
+
+/**
+ * Deling the status for user
```
SanketDhabarde, line 5:
```suggestion
 * Deleting the status of a user
```

models/users.js
samarpan1738: Adding the userStatus enum should also be done through a separate issue

SanketDhabarde: Created the issue https://github.com/Real-Dev-Squad/website-backend/issues/767

SanketDhabarde: done

samarpan1738 (approved):
LGTM

SanketDhabarde:
Closing this PR. As the new status collection is getting created which make this issue irrelevant https://github.com/Real-Dev-Squad/website-backend/issues/818

Issue #756, opened 2022-10-25
CORS Error on the website


Maheima:
### Issue:

On https://realdevsquad.com/ , there is a CORS error because of an incorrect regex in the allowed origin in configuration files. The error isn't present on the subdomains

![image](https://user-images.githubusercontent.com/42349041/197848735-e114cd43-9904-4c90-b975-c2154f33b176.png)

### Solution:
Correct the regex which makes sure that `.` is optional, anything before it and after `https://` is acceptable

ankushdharkar:
Fixed in #762

PR #762, opened 2022-10-26
fix CORS error

Maheima:
Solves https://github.com/Real-Dev-Squad/website-backend/issues/756
Changed the Regex from `/(\.realdevsquad\.com$)|(localhost)/` to `/(https:\/\/([a-zA-Z0-9-_]+\.)?realdevsquad\.com$)|(localhost)/` in Allowed origins

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
 /**
  * Default config to be used if environment specific config for the specific key is absent
  * Every config key to be added to `default.js` to keep a track of all config keys used in the project.
@@ -49,7 +50,7 @@ module.exports = {
   },
 
   cors: {
-    allowedOrigins: /https:\/\/(.*\.)*realdevsquad\.com$/,
+    allowedOrigins: /https:\/\/([a-zA-Z0-9-_]+\.)?realdevsquad\.com$/,
   },
 
   userToken: {
--- a/config/default.js
+++ b/config/default.js
@@ -49,7 +49,7 @@ module.exports = {
   },
 
   cors: {
-    allowedOrigins: /(https:\/\/([a-zA-Z0-9-_]+\.)?realdevsquad\.com$)/, // Allow realdevsquad.com, *.realdevsquad.com and localhost for non-production envs
+    allowedOrigins: /(https:\/\/([a-zA-Z0-9-_]+\.)?realdevsquad\.com$)/, // Allow realdevsquad.com, *.realdevsquad.com
   },
 
   userToken: {
--- a/config/default.js
+++ b/config/default.js
@@ -49,7 +49,7 @@ module.exports = {
   },
 
   cors: {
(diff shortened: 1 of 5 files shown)


config/development.js
```diff
@@ -25,7 +26,7 @@ module.exports = {
   },
 
   cors: {
-    allowedOrigins: /(\.realdevsquad\.com$)|(localhost)/, // Allow *.realdevsquad.com and localhost for non-production envs
+    allowedOrigins: /(https:\/\/(.*\.)*realdevsquad\.com$)|(localhost)/, // Allow *.realdevsquad.com and localhost for non-production envs
```
ankushdharkar, line 29: > Allow *.realdevsquad.com and localhost for non-production envs

The comment is slightly incorrect.

We also allow `realdevsquad.com` (without the subdomain), correct?

ankushdharkar (changes requested):

config/development.js
RitikJaiswal75, line 29: currently this will allow a.b.realdevsquad.com, that should not happen I guess please check and fix

Maheima: fixed it

Maheima: Done

config/staging.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
   },
 
   cors: {
-    allowedOrigins: /(\.realdevsquad\.com$)|(localhost)/, // Allow *.realdevsquad.com and localhost for non-production envs
+    allowedOrigins: /(https:\/\/([a-zA-Z0-9-_]+\.)?realdevsquad\.com$)|(localhost)/, // Allow *.realdevsquad.com and localhost for non-production envs
```
shubham-y, line 20: Please update the comment here as well as in `config/test.js`

config/staging.js
```diff
@@ -16,10 +16,6 @@ module.exports = {
     },
   },
 
-  cors: {
```
ankurnarkhede, line 19: We are not overriding the values in env specific config. So it shouldnt be added here... 🤔

ankurnarkhede (changes requested):

config/default.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
   },
 
   cors: {
-    allowedOrigins: /\.realdevsquad\.com$/,
+    allowedOrigins: /(https:\/\/([a-zA-Z0-9-_]+\.)?realdevsquad\.com$)/, // Allow realdevsquad.com, *.realdevsquad.com and localhost for non-production envs
```
ankurnarkhede, line 52: Also, can you add why this issue is occurring now, what has changed?

ankurnarkhede (dismissed):

config/staging.js
ankurnarkhede: Okay, this has been removed 👍

config/default.js
ankurnarkhede: As discussed, this is because of www not present in the domain.

ankushdharkar:
🚢

ankushdharkar (approved):
🚢

Issue #766, opened 2022-10-28
Add default status for new user


SanketDhabarde:
### AC
Add default status idle when a new user is created

### Description:
When a user joins RDS, the default status of the user is "idle". To represent this,  in the user modal we have `status: "idle"`. This field can also be used as a query in firestore to filter the idle users https://github.com/Real-Dev-Squad/website-backend/issues/635

SanketDhabarde:
Closing this Issue. As the new status collection is getting created which make this issue irrelevant https://github.com/Real-Dev-Squad/website-backend/issues/818

PR #768, opened 2022-10-29
Added object enum for user status.

SanketDhabarde:
### Changes made
Converted:
```js
const userStatusEnum = ["ooo", "idle", "active"];
```
To:
```js
const USER_STATUS = {
  OOO: "ooo",
  IDLE: "idle",
  ACTIVE: "active",
};
```
And used it at the places where `userStatusEnum` was used

### Dependent tasks
- https://github.com/Real-Dev-Squad/website-backend/issues/754
- https://github.com/Real-Dev-Squad/website-backend/issues/766

### DIFF
--- a/constants/users.js
+++ b/constants/users.js
@@ -1,7 +1,13 @@
-const userStatusEnum = ["ooo", "idle", "active"];
 const profileStatus = {
   PENDING: "PENDING",
   APPROVED: "APPROVED",
   NOT_APPROVED: "NOT APPROVED",
 };
-module.exports = { userStatusEnum, profileStatus };
+
+const userStatus = {
+  OOO: "ooo",
+  IDLE: "idle",
+  ACTIVE: "active",
+};
+
+module.exports = { profileStatus, userStatus };
--- a/constants/users.js
+++ b/constants/users.js
@@ -4,10 +4,10 @@ const profileStatus = {
   NOT_APPROVED: "NOT APPROVED",
 };
 
-const userStatus = {
+const USER_STATUS = {
   OOO: "ooo",
   IDLE: "idle",
   ACTIVE: "active",
 };
 
-module.exports = { profileStatus, userStatus };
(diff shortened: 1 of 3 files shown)


controllers/tasks.js
```diff
@@ -211,7 +212,7 @@ const assignTask = async (req, res) => {
   try {
     const { status, username } = req.userData;
 
-    if (status !== userStatusEnum[1]) {
+    if (status !== IDLE) {
```
samarpan1738, line 215: This would be more readable
```suggestion
    if (status !== userStatus.IDLE) {
```

SanketDhabarde: No, There is a better way. If you want to use more status suppose "active", you will do it like `userStatus.ACTIVE`. and for "ooo" you will use `userStatus.OOO`. So if we use an object to access property more than 2 times, we usually destructure the object.

constants/users.js
```diff
@@ -1,7 +1,13 @@
-const userStatusEnum = ["ooo", "idle", "active"];
 const profileStatus = {
   PENDING: "PENDING",
   APPROVED: "APPROVED",
   NOT_APPROVED: "NOT APPROVED",
 };
-module.exports = { userStatusEnum, profileStatus };
+
+const userStatus = {
```
samarpan1738, line 7: Constants are usually uppercase
```suggestion
const USER_STATUS = {
```

controllers/tasks.js
samarpan1738: In RDS we task and user both have ACTIVE status. If we destructure, it'll be ambiguous that which entity the status belongs

SanketDhabarde: In that case. it makes sense not to destructure. I will make the change thank you.

samarpan1738 (approved):
LGTM

vinayak-trivedi (approved):
LGTM 👍

PR #803, opened 2022-11-03
fix min char length in joi as per frontend

RitikJaiswal75:
## Issue
Unable to save data in the backend if college name is less than 5 words.

### Description
In front, we changed the institution size to be 1 word in the front end but did not fix it in the backend. 

This PR fixes the JOI validation in the backend.


heyrandhir (approved):
lgtm.

middlewares/validators/user.js
```diff
@@ -54,15 +54,15 @@ const validateJoinData = async (req, res, next) => {
     .strict()
     .keys({
       userId: joi.string().optional(),
-      firstName: joi.string().required(),
-      lastName: joi.string().required(),
-      college: joi.string().min(5).required(),
+      firstName: joi.string().min(1).required(),
+      lastName: joi.string().min(1).required(),
+      college: joi.string().min(1).required(),
       skills: joi.string().min(5).required(),
```
prakashchoudhary07, line 60: What if someone has only one skill like `c` only or no skill, and is joining for skills development

RitikJaiswal75: They can also mention their personal skills and use sentences like "I am good at x y z"

RitikJaiswal75: it was mandatory to use 5 words in skills

Issue #818, opened 2022-11-12
Create endpoints for userStatus collection


heyrandhir:
Create the following endpoints for the User Status Collection

EndPoint | Method | use
-- | -- | --
/user-status | GET | returns the user status data of all the users
/user-status/:userId | GET | returns the user status data of a single user
/user-status/:userId | POST | Adds the userStatus data for a particular user
/user-status/:userId | DELETE | Removes the userStatus data for a particular user
/user-status/:userId | PATCH | Updates the userStatus data for a particular user

The data model for the User Status will be as follows
```
{
  "userId": "String",
  "currentStatus":  {
      "state": "OOO | IDLE | ACTIVE",
      "updatedAt": "TimeStamp",
      "from": "TimeStamp ",
      "until": "TimeStamp",
      "message": "String"
    },
  "monthlyHours": {
      "committed": "Integer",
      "updatedAt": "TimeStamp"
    }
}
```

PR #819, opened 2022-11-12
User Status Endpoint Creation

heyrandhir:
Closes https://github.com/Real-Dev-Squad/website-backend/issues/818 , https://github.com/Real-Dev-Squad/website-backend/issues/834

Created the following endpoints.
**EndPoint | Method | use
-- | -- | --
/userStatus | GET | returns the User Status data of all the users
/userStatus/:userId | GET | returns the User status data of a single user
/userStatus/:userId | POST | Adds the User Status data for a particular user
/userStatus/:userId | DELETE | Removes the User Status data for a particular user
/userStatus/:userId | PATCH | Updates the User Status data for a particular user**

Tests are not written yet. 
Tentative Date for Test Completion - 16/11/2022

Update 17/11/22
Tests Updated


ivinayakg:
shouldn't the route be `user/:id/status` instead of this one??
What do you think?

heyrandhir:
> shouldn't the route be `user/:id/status` instead of this one?? What do you think?

yeah this makes more sense

ivinayakg:
> > shouldn't the route be `user/:id/status` instead of this one?? What do you think?
> 
> yeah this makes more sense

Hey i didn't know that there was a separate collection for this data,  if that is so then your approach is fine.

models/status.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+};
+
+/**
+ *
+ * @returns {Promise<tagModel|Array>}: returns an array of all the tags with its meta data
+ */
+
+const getAllTags = async () => {
+  const data = await tagModel.get();
```
alishvagadiya, line 43: Add a try-catch block.

models/userStatus.js
```diff
@@ -0,0 +1,115 @@
+const { userState } = require("../constants/userStatus");
```
alishvagadiya, line 1: Why is the return part outside the catch block in all model functions?

alishvagadiya (changes requested):

controllers/userStatus.js
```diff
@@ -0,0 +1,99 @@
...
+
+const deleteUserStatus = async (req, res) => {
+  try {
+    const deletedUserStatus = await userStatusModel.deleteUserStatus(req.params.userId);
+    return res.status(201).json({
+      message: "UserStatus Deleted successfully!",
+      ...deletedUserStatus,
+    });
+  } catch (error) {
+    logger.error(`Error while creating new User Status: ${error}`);
```
SanketDhabarde, line 38:
```suggestion
    logger.error(`Error while deleting new User Status: ${error}`);
```

controllers/userStatus.js
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Collects single User Status
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const getUserStatus = async (req, res) => {
+  try {
+    const userData = await userStatusModel.getUserStaus(req.params.userId);
```
SanketDhabarde, line 51: You can destructure the `req.params`. It will be helpful if we want to pass more params in future

models/userStatus.js
```diff
@@ -0,0 +1,115 @@
...
+    if (docs._size > 0) {
+      let id;
+      docs.forEach((docData) => {
+        id = docData.id;
+      });
+      if (Object.keys(updatedData).includes("currentStatus")) {
+        if ([userState.IDLE, userState.ACTIVE].includes(updatedData.currentStatus.state)) {
+          updatedData.currentStatus.until = "";
+        }
+        if ([userState.ACTIVE].includes(updatedData.currentStatus.state)) {
```
SanketDhabarde, line 100: Why the use of `includes` for a single value check? Can we do an equality check?

models/userStatus.js
heyrandhir: I am getting the error message `error  Expected to return a value at the end of async arrow function  consistent-return` while running the command `yarn lint-fix`. Hence I added the return outside the try-catch block.

models/status.js
heyrandhir: this was an unwanted file. Pushed by mistake. Thanks

models/userStatus.js
SanketDhabarde: IMO for a single value, it's better to use equality. We can use includes in the future if more fields need to check.

models/userStatus.js
```diff
@@ -0,0 +1,115 @@
...
+const updateUserStatus = async (userId, updatedData) => {
+  try {
+    const docs = await userStatusModel.where("userId", "==", userId).limit(1).get();
+    if (docs._size > 0) {
+      let id;
+      docs.forEach((docData) => {
+        id = docData.id;
+      });
+      if (Object.keys(updatedData).includes("currentStatus")) {
+        if ([userState.IDLE, userState.ACTIVE].includes(updatedData.currentStatus.state)) {
```
SanketDhabarde, line 97: What is this condition doing?  what I understood is if the user's updating state is not OOO then until is empty. right? I think the condition can be improved.

models/userStatus.js
heyrandhir: Sure I will update it.

models/userStatus.js
SanketDhabarde:
```js
const isUserOOO = updatedData.currentStatus.state === userState.OOO;

if(!isUserOOO) {
  updatedData.currentStatus.until = "";
}
```
What do you think about this?

heyrandhir: yeah, this is better. Thanks.

models/userStatus.js
```diff
@@ -0,0 +1,121 @@
...
+        id = docData.id;
+      });
+      if (Object.keys(updatedData).includes("currentStatus")) {
+        const isUserOOO = updatedData.currentStatus.state === userState.OOO;
+        if (!isUserOOO) {
+          updatedData.currentStatus.until = "";
+        }
+        if (updatedData.currentStatus.state === userState.ACTIVE) {
+          updatedData.currentStatus.message = "";
+        }
```
SanketDhabarde, line 108:
```js
        const updatedUserState = updatedData.currentStatus.state;
        const isUserOOO =  updatedUserState === userState.OOO;
        const isUserActive = updatedUserState === userState.ACTIVE;

        if (!isUserOOO) {
          updatedData.currentStatus.until = "";
        }
        if (isUserActive) {
          updatedData.currentStatus.message = "";
        }
```
If you think it's good. We can be consistent.

heyrandhir: Yeah @SanketDhabarde this is better and consistent. Thanks.

Maheima (approved):
LGTM

test/fixtures/userStatus/userStatus.js
```diff
@@ -0,0 +1,56 @@
+const userStsDataForOooState = (userId) => {
+  return {
+    userId,
+    currentStatus: {
+      until: "1669256009000",
```
ankushdharkar, line 5: @heyrandhir If these are strings, would we be able to perform range search queries on the dates?

Should these be numeric instead? And be indexable?

middlewares/validators/userStatus.js
```diff
@@ -0,0 +1,68 @@
...
+};
+
+const validateUpdatedUserStatus = async (req, res, next) => {
+  const schema = joi
+    .object({
+      currentStatus: joi.object().keys({
+        state: joi.string().trim().valid(userState.IDLE, userState.ACTIVE, userState.OOO),
+        updatedAt: joi.number().required(),
+        from: joi.number().required(),
+        until: joi.any().when("state", { is: userState.OOO, then: joi.number().required(), otherwise: joi.optional() }),
```
ankushdharkar, line 42: Why is this `any`? Why is `until` not the same datatype as `from` above`?

test/fixtures/userStatus/userStatus.js
```diff
@@ -0,0 +1,56 @@
...
+
+const invalidUserStsDataforPost = (userId) => {
+  return {
+    userId,
+    currentStatus: {
+      until: "",
+      message: "",
+      state: "ACTIVE",
+      updatedAt: "1668215609000",
+      from: "1668215609000",
```
ankushdharkar, line 26: Why are the tests passing? I though `joi` was enforcing `number` type for `from`. Here it is the incorrect string type.

ankushdharkar (changes requested):

middlewares/validators/userStatus.js
heyrandhir: the `until` field will only be present for the `OOO` state while the `from` field will be present for all three states `OOO`, `IDLE`, and `ACTIVE`.

routes/userStatus.js
```diff
@@ -0,0 +1,19 @@
...
+  updateUserStatus,
+} = require("../controllers/userStatus");
+const router = express.Router();
+const authenticate = require("../middlewares/authenticate");
+const { validateUserStatus, validateUpdatedUserStatus } = require("../middlewares/validators/userStatus");
+
+router.get("/", getAllUserStatus);
+router.get("/:userId", authenticate, getUserStatus);
+router.post("/:userId", authenticate, validateUserStatus, addUserStatus);
+router.delete("/:userId", authenticate, deleteUserStatus);
```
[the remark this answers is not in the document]
heyrandhir: yeah, only the `DELETE` is a super user-only feature. Other methods like `GET`,`POST`,`PATCH` can be done by a normal user as well.This comment is in outdated code hence it's not visible here. It's visible in the latest code [here](https://github.com/Real-Dev-Squad/website-backend/pull/819/files#diff-bf0502a0332dc78877a9e4fea7a9dfb52a1c0fc7ae157161eab434f32ca0fa7cR19)

test/fixtures/userStatus/userStatus.js
heyrandhir: The tests were passing because Joi converts strings if it can by default. However, if we want it to be a strict number we can add `.strict()` in the validation.

test/fixtures/userStatus/userStatus.js
heyrandhir: I need to check whether we can perform a range search on the date field. Yeah, it would be better to store it as numeric. I have updated the code to store it as numeric.

controllers/userStatus.js
```diff
@@ -0,0 +1,102 @@
...
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const getUserStatus = async (req, res) => {
+  try {
+    const { userId } = req.params;
+    const userData = await userStatusModel.getUserStaus(userId);
+    return res.json({ message: "User Status found successfully", ...userData });
+  } catch (err) {
+    // can return 404 error as well
```
harshith-venkatesh, line 56: This provides better information for user as well as easy to debug as well

middlewares/validators/userStatus.js
```diff
@@ -0,0 +1,72 @@
...
+  const schema = joi
+    .object({
+      userId: joi.string().trim().required(),
+      currentStatus: joi.object().keys({
+        state: joi.string().trim().valid(userState.IDLE, userState.ACTIVE, userState.OOO),
+        updatedAt: joi.number().required().strict(),
+        from: joi.number().required().strict(),
+        until: joi
+          .any()
+          .when("state", { is: userState.OOO, then: joi.number().required().strict(), otherwise: joi.optional() }),
```
harshith-venkatesh, line 14: Would until not be number | null ?
Do we have better way than any?

models/userStatus.js
```diff
@@ -0,0 +1,124 @@
...
+    logger.error(`error in deleting User Status Document . Reason - ${error}`);
+  }
+  return { message: `No UserStatus for ${userId} found.` };
+};
+
+/**
+ * @params userId {string} : id of the user
+ * @returns {Promise<userStatusModel|Object>} : returns the userStatus of a single user
+ */
+const getUserStaus = async (userId) => {
```
harshith-venkatesh, line 44: Spelling mistake

models/userStatus.js
```diff
@@ -0,0 +1,124 @@
...
+  } catch (error) {
+    logger.error(`error in fetching the User Status Document. Reason - ${error}`);
+  }
+  return { message: "User not found", userId, data: null };
+};
+
+/**
+ * @returns {Promise<userStatusModel|Array>} : returns an array of all the userStatus
+ */
+const getAllUserStaus = async (query) => {
```
harshith-venkatesh, line 63: Spelling mistake

harshith-venkatesh (changes requested):

middlewares/validators/userStatus.js
heyrandhir: for updating User Status `until` is an optional field in the request body. It may or may not be present. Its required only for `OOO` status. Hence only in the case of `OOO`, I am checking if it should be a number. And In the controller, we have set this to an empty string if the state is not `OOO`.

heyrandhir: Since anyway the value of `until` is going to be overwritten with an empty string for everything apart from `OOO` so kept it joi.any().

models/userStatus.js
heyrandhir: Oops 😅

controllers/userStatus.js
heyrandhir: Added a status code 404 for the record not found.

test/integration/userStatus.test.js
```diff
@@ -0,0 +1,227 @@
...
+        .set("cookie", `${cookieName}=${jwt}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("User Status found successfully");
+          expect(res.body).to.have.property("monthlyHours");
+          expect(res.body).to.have.property("currentStatus");
```
RitikJaiswal75, line 77: can we add expect body.userid = userId<variable>

test/integration/userStatus.test.js
```diff
@@ -0,0 +1,227 @@
...
+
+    beforeEach(async function () {
+      await cleanDb();
+      testUserId = await addUser();
+      testUserJwt = authService.generateAuthToken({ userId: testUserId });
+    });
+
+    afterEach(async function () {
+      await cleanDb();
+    });
```
RitikJaiswal75, line 95: do we need this as it is already done by the code in line 40

RitikJaiswal75 (changes requested):

constants/userStatus.js
```diff
@@ -0,0 +1,7 @@
+const userState = {
```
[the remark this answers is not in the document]
heyrandhir: we decided to call the collection status hence its acceptable values as state. and since these are states of user Status hence calling it user state.

controllers/userStatus.js
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const getUserStatus = async (req, res) => {
+  try {
+    const { userId } = req.params;
+    const userData = await userStatusModel.getUserStaus(userId);
+    if (userData.userStatusExists) {
+      return res.json({ message: "User Status found successfully", ...userData });
```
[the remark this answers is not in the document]
heyrandhir: conveys the same meaning.

controllers/userStatus.js
```diff
@@ -0,0 +1,120 @@
...
+    const deletedUserStatus = await userStatusModel.deleteUserStatus(userId);
+    if (deletedUserStatus.userStatusExisted) {
+      return res.status(200).json({
+        userId,
+        message: "UserStatus Deleted successfully.",
+      });
+    } else {
+      return res.status(404).json({
+        userId,
+        message: "UserStatus to delete could not be found.",
```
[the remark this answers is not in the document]
heyrandhir: updated the message.

Maheima (approved):
LGTM

models/userStatus.js
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+const updateUserStatus = async (userId, updatedData) => {
+  try {
+    const docs = await userStatusModel.where("userId", "==", userId).limit(1).get();
+    if (docs._size > 0) {
+      let id;
+      docs.forEach((docData) => {
+        id = docData.id;
+      });
```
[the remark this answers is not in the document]
prakashchoudhary07: Please try if this works https://github.com/Real-Dev-Squad/website-backend/blob/13ca52500f5def42a50f276003eb8e3988e9b423/models/wallets.js#L11-L16

controllers/userStatus.js
```diff
@@ -0,0 +1,121 @@
...
+        id: deletedUserStatus.id,
+        userId,
+        message: "UserStatus Deleted successfully.",
+      });
+    } else {
+      return res.status(404).json({
+        id: deletedUserStatus.id,
+        userId,
+        message: "Could not find the user status linked to the user for deletion.",
+      });
```
vinayak-trivedi, line 26: this is becoming a bit repetitive, can you look at this, we are sending the same response mostly in both the places

models/userStatus.js
```diff
@@ -0,0 +1,111 @@
...
+    const [userStatusDoc] = userStatusDocs.docs;
+    if (userStatusDoc) {
+      const id = userStatusDoc.id;
+      const data = userStatusDoc.data();
+      return { id, userId, ...data, userStatusExists: true };
+    }
+  } catch (error) {
+    logger.error(`error in fetching the User Status Document. Reason - ${error}`);
+  }
+  return { id: undefined, userId, userStatusExists: false };
```
vinayak-trivedi, line 40: can we put it in else

models/userStatus.js
```diff
@@ -0,0 +1,111 @@
...
+  try {
+    const userStatusDocs = await userStatusModel.where("userId", "==", userId).limit(1).get();
+    const [userStatusDoc] = userStatusDocs.docs;
+    if (userStatusDoc) {
+      const id = userStatusDoc.id;
+      const data = userStatusDoc.data();
+      return { id, userId, ...data, userStatusExists: true };
+    }
+  } catch (error) {
+    logger.error(`error in fetching the User Status Document. Reason - ${error}`);
```
vinayak-trivedi, line 38: we should put throw error here

controllers/userStatus.js
heyrandhir: As Mahima is blocked on this PR hence will resolve the pending comments in a new PR.[ Issue ticket ](https://github.com/Real-Dev-Squad/website-backend/issues/834) for the same.

models/userStatus.js
heyrandhir: As Mahima is blocked on this PR hence will resolve the pending comments in a new PR.[ Issue ticket ](https://github.com/Real-Dev-Squad/website-backend/issues/834) for the same.

models/userStatus.js
heyrandhir: As Mahima is blocked on this PR hence will resolve the pending comments in a new PR.[ Issue ticket ](https://github.com/Real-Dev-Squad/website-backend/issues/834) for the same.

heyrandhir: Since it was not merged, resolved in the same PR.

models/userStatus.js
heyrandhir: Since it was not merged, resolved in the same PR.

controllers/userStatus.js
heyrandhir: Since it was not merged, resolved in the same PR.

Maheima (approved):
LGTM!

PR #825, opened 2022-11-19
initial commit

### LINKED ISSUE #800: [RFC] Generating auto incrementing sequence for task number
Tracked in https://github.com/Real-Dev-Squad/todo-action-items/issues/77

### **Approach 1**
When a new task is created, update the counter by reading from and writing to the **taskCounter** collection that will store the previous value of the counter.

**Cons**
1.  Creating a whole new collection just for storing the previous counter value
2.  Two operations are performed - Reading the old counter value and then using it to increment the value and storing back to the taskCounter collection.

### **Approach 2**
Getting QuerySnapshot of the task collection and using size property

**Cons** 
1. The main problem is the cost, by executing this query we will be charged for one read for each doc of the sub-collection.

### **Approach 3**
Approach 1 can be also be achieved using cloud functions

YashJain24-chief:
Closes #800 
This PR will help to auto-generate a task number whenever a new task is created from the [dashboard site](https://dashboard.realdevsquad.com/) by a super-user. This task number will be utilized on the task details page as mentioned in https://github.com/Real-Dev-Squad/todo-action-items/issues/77.

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -16,6 +16,7 @@ const addNewTask = async (req, res) => {
     const body = {
       ...req.body,
       createdBy,
+      taskNo: await tasks.generateTaskNo(),
     };
     const task = await tasks.updateTask(body);
 
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -284,6 +284,12 @@ const overdueTasks = async (overDueTasks) => {
     throw err;
   }
 };
+
+const generateTaskNo = async () => {
+  const snapshot = await tasksModel.count().get();
+  return snapshot.data().count + 1;
+};
+
 module.exports = {
   updateTask,
   fetchTasks,
@@ -295,4 +301,5 @@ module.exports = {
   fetchSelfTask,
   fetchSkillLevelTask,
   overdueTasks,
+  generateTaskNo,
 };


models/tasks.js
```diff
@@ -284,6 +284,12 @@ const overdueTasks = async (overDueTasks) => {
     throw err;
   }
 };
+
+const generateTaskNo = async () => {
+  const snapshot = await tasksModel.count().get();
+  return snapshot.data().count + 1;
+};
```
Maheima, line 291: Thinking out loud:  Is there a better way to number tasks? because for the task that is not there anymore, a number assigned to it gets wasted.

YashJain24-chief: `the task that is not there anymore`. Can you provide some more clarity on it?

SimplySonu (dismissed):

models/tasks.js
```diff
@@ -284,6 +284,12 @@ const overdueTasks = async (overDueTasks) => {
     throw err;
   }
 };
+
+const generateTaskNo = async () => {
+  const snapshot = await tasksModel.count().get();
```
SanketDhabarde, line 289: Is this working for you?

YashJain24-chief: Yes, it is. Attached a screenshot of the firestore that created a task number when we hit a post request to create new task

SanketDhabarde: Okay. I have tried using the same but got an error count that is not a function. So just wanted to check if it's working.

YashJain24-chief: count function requires latest version of firebase SDK, make sure you have an updated one

controllers/tasks.js
```diff
@@ -16,6 +16,7 @@ const addNewTask = async (req, res) => {
     const body = {
       ...req.body,
       createdBy,
+      taskNo: await tasks.generateTaskNo(),
```
Muralidhar22, line 19: shouldn't this be like `taskNum: ` and where are we handling error?

YashJain24-chief: done, good find

test/integration/tasks.test.js
```diff
@@ -110,6 +110,7 @@ describe("Tasks", function () {
           expect(res.body.task.createdBy).to.equal(appOwner.username);
           expect(res.body.task.assignee).to.equal(appOwner.username);
           expect(res.body.task.participants).to.be.a("array");
+          expect(res.body.task.taskNumber).to.equal(3);
```
ankushdharkar, line 113: ❓Is this test a little brittle? How do we know this should be 3?

YashJain24-chief: In line number 34 https://github.com/Real-Dev-Squad/website-backend/pull/825/files#diff-8fd2efa3a5824bd99d050f5e6c5d5adffd59bf49902ac7bfb42bb8a7eaad4a52R34, we are creating a dummy task data array which has details of 2 tasks, so whenever we are making a post request shouldn't task number that is returned be 3 or should I check for it to be a number instead of just 3?

vinayak-trivedi (approved):
LGTM 👍

Muralidhar22 (approved):
LGTM

models/tasks.js
```diff
@@ -284,6 +284,17 @@ const overdueTasks = async (overDueTasks) => {
     throw err;
   }
 };
+
+const generateTaskNumber = async () => {
+  try {
+    const snapshot = await tasksModel.count().get();
+    return snapshot.data().count + 1;
```
ankushdharkar, line 291: How does this avoid race conditions?

YashJain24-chief: It doesn't handle race conditions. As of now isn't just the super user creating tasks using the dashboard site? For now, it won't cause race conditions, but in the future, if we are allowing many members or AOs to create then it might create race conditions. Also, you said till then we might migrate to better infrastructure. So, how do you want me to handle this?

PR #827, opened 2022-11-19
feat/paginating-logs-endpoint

### LINKED ISSUE #826: Paginating logs endpoint
Adding pagination to logs endpoint to limit the number of requests made to the firestore and getting the data as and when required.

YashJain24-chief:
Closes #826 
Paginating the logs endpoint to reduce the number of requests made to firestore. Practically, we don't need to see all the logs in one go, instead, we prefer to see the most recent and limited logs.
### **Features**
1. If we send `limit` as a query param, it will return the top 'n' documents.
2. If we send `lastDocId` as a query param, it will return all the documents after the document whose id will be equal to the `lastDocId`.
3. If we send both the `limit` and `lastDocId`, it will return 'n' limited documents after the specific document whose id will be equal to the `lastDocId`.
4. If we do not send any query params like `limit` or `lastDocId`, only top 10 logs will be returned.

Note: The data is sorted in descending order of **timestamp** so that we can get the most recent logs.

### DIFF
--- a/models/logs.js
+++ b/models/logs.js
@@ -41,14 +41,14 @@ const fetchLogs = async (query, param) => {
     });
 
     const { limit, lastDocId } = query;
-    let lastDoc;
+    let lastDoc = "";
     const limitDocuments = Number(limit);
 
     if (lastDocId) {
       lastDoc = await logsModel.doc(lastDocId).get();
     }
 
-    const logsSnapshotQuery = call.orderBy("timestamp", "desc").startAfter(lastDoc ? lastDoc : "");
+    const logsSnapshotQuery = call.orderBy("timestamp", "desc").startAfter(lastDoc);
     const snapshot = limit ? await logsSnapshotQuery.limit(limitDocuments).get() : await logsSnapshotQuery.get();
 
     const logs = [];
--- a/models/logs.js
+++ b/models/logs.js
@@ -49,7 +49,9 @@ const fetchLogs = async (query, param) => {
     }
 
     const logsSnapshotQuery = call.orderBy("timestamp", "desc").startAfter(lastDoc ?? "");
-    const snapshot = limit ? await logsSnapshotQuery.limit(limitDocuments).get() : await logsSnapshotQuery.get();
+    const snapshot = limit
+      ? await logsSnapshotQuery.limit(limitDocuments).get()
+      : await logsSnapshotQuery.limit(10).get();
 
     const logs = [];
     snapshot.forEach((doc) => {
--- a/models/logs.js
+++ b/models/logs.js
@@ -35,10 +35,22 @@ const fetchLogs = async (query, param) => {
     let call = logsModel.where("type", "==", param);
     Object.keys(query).forEach((key) => {
       // eslint-disable-next-line security/detect-object-injection
-      call = call.where(key, "==", query[key]);
+      if (key !== "limit" && key !== "lastDocId") {
+        call = call.where(key, "==", query[key]);
+      }
     });
 
-    const snapshot = await call.get();
+    const { limit, lastDocId } = query;
+    let lastDoc;
+    const limitDocuments = Number(limit);
+
+    if (lastDocId) {
+      lastDoc = await logsModel.doc(lastDocId).get();
+    }
+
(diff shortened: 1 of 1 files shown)


Maheima (dismissed):
LGTM

YashJain24-chief:
> LGTM

models/logs.js
```diff
@@ -35,10 +35,22 @@ const fetchLogs = async (query, param) => {
     let call = logsModel.where("type", "==", param);
     Object.keys(query).forEach((key) => {
       // eslint-disable-next-line security/detect-object-injection
-      call = call.where(key, "==", query[key]);
+      if (key !== "limit" && key !== "lastDocId") {
```
alishvagadiya, line 38: If condition will be passed for all keys value other than limit and lastDocId,
Why not check which are allowed only?

models/logs.js
```diff
@@ -35,10 +35,22 @@ const fetchLogs = async (query, param) => {
     let call = logsModel.where("type", "==", param);
     Object.keys(query).forEach((key) => {
       // eslint-disable-next-line security/detect-object-injection
```
alishvagadiya, line 37: eslint-disable is still required, as you are changing the below condition.
check once after removing this condition.

models/logs.js
YashJain24-chief: We don't know based on what keys we might want to filter, it can be anything or everything. Instead, I considered excluding the filtering process for `limit` and `lastDocId`

models/logs.js
YashJain24-chief: Need some more clarity on this @alishvagadiya

vinayak-trivedi:
> Closes #826 Paginating the logs endpoint to reduce the number of requests made to firestore. Practically, we don't need to see all the logs in one go, instead, we prefer to see the most recent and limited logs.
> 
> ### **Features**
> 1. If we send `limit` as a query param, it will return the top 'n' documents.
> 2. If we send `lastDocId` as a query param, it will return all the documents after the document whose id will be equal to the `lastDocId`.
> 3. If we send both the `limit` and `lastDocId`, it will return 'n' limited documents after the specific document whose id will be equal to the `lastDocId`.
> 4. If we do not send any query params like `limit` or `lastDocId`, all the logs will be returned. Included this so that if the endpoint is being used elsewhere before this pagination feature was implemented, things won't break.
> 
> Note: The data is sorted in descending order of **timestamp** so that we can get the most recent logs.

thanks for providing detailed description of the changes

models/logs.js
```diff
@@ -35,10 +35,22 @@ const fetchLogs = async (query, param) => {
...
-      call = call.where(key, "==", query[key]);
+      if (key !== "limit" && key !== "lastDocId") {
+        call = call.where(key, "==", query[key]);
+      }
     });
 
-    const snapshot = await call.get();
+    const { limit, lastDocId } = query;
+    let lastDoc = "";
```
vinayak-trivedi, line 44: why do we need to do that? does it give any error if it is undefined?

YashJain24-chief: yes, we cannot use undefined as a firestore value. Refactored the code using nullish coalescing

Maheima (dismissed):
LGTM!

prakashchoudhary07:
I think we should limit the response instead of sending it all. 
Ex: If no limit is passed then limit to 10 or 20 or some number, and continue from there

Log collection will have huge data, as logs are generated from many places.

models/logs.js
```diff
@@ -35,10 +35,24 @@ const fetchLogs = async (query, param) => {
...
-    const snapshot = await call.get();
+    const { limit, lastDocId } = query;
+    let lastDoc;
+    const limitDocuments = Number(limit);
+
+    if (lastDocId) {
+      lastDoc = await logsModel.doc(lastDocId).get();
+    }
+
+    const logsSnapshotQuery = call.orderBy("timestamp", "desc").startAfter(lastDoc ?? "");
```
prakashchoudhary07, line 51: Do we fix the order to descending?

YashJain24-chief: Ideally we would want to see the last updated logs and hence kept it desc. You want me to make it configurable by taking a query param from frontend?

prakashchoudhary07 (approved):

Issue #760, opened 2022-10-26
[RFC] - Feature Task ETA extension


ivinayakg:
- every eta extension will be logged , structure for log is... ```{}```
- ETA extension will be a different model
   ```
   {
   taskid, userId, username, New ETA, status, old ETA, reason
   }
   ```
- every request will get approved by `SUPERUSER | APPOWNER` and then only the ETA will be extended.

shivampandey0:
- Every ETA extension will be logged, structure for log is:
```
{ 
 type : 'task',
 meta : {
    taskId,
    username,
    userId
 }
 timestamp,
 body : {
   subType : UPDATE 
   new : {
       status, 
       endsOn,
       //many more...
       messages : []
    }
 }
}
```

- ETA extension will be a different model
```
{
  taskId: <id>,
  title: string,
  assignee: string,
  participants: [userId, userId],
  oldEndsOn: string,
  newEndsOn: string,
  reason: string,
  status
}
```

- Every request will get approved by **SUPERUSER | APPOWNER** and then only the ETA will be extended.

ivinayakg:
@vinayak-trivedi

ivinayakg:
- PR for API Contracts - https://github.com/Real-Dev-Squad/website-api-contracts/pull/96

PR #829, opened 2022-11-21
Extension Request APIs created

ivinayakg:
# Parent Issue - #760 

## APIs created
* `/extensionRequests GET` for `SUPERUSER | APPOWNER` to get all the extension requests for the ETA of tasks.
    * You can also pass query parameters to get a filter list of extension requests. Currently, you can filter by `?status=PENDING | APPROVED | DENIED` or `?taskId=<task id>` to get the extension request of a task or `?assignee=<user id>` to get all extension requests of a particular user or you can use all three of them.
    
* `/extensionRequests POST` for:- 
    * When `USER` uses this they can create extension requests for tasks assigned to them only.
    * When `SUPERUSER` uses this they can create an extension request for any task assigned to anyone.
    * Extension requests cannot be created for a task that already has an extension request.
    * Creation of extension request event will be logged in the following manner:-
```
{ 
 type: 'extensionRequest',
 meta: {
    createdBy,
    taskId,
 }
 timestamp,
 body: {
   oldEndsOn,
   newEndsOn,
   extensionRequestId,
   assignee
 }
}
```
* `/extensionRequest/self GET` for `USER` to get all the extension requests for the tasks assigned to them.
   * You can also pass query parameters to get a filter list of extension requests. Currently, you can filter by `?status=PENDING | 
      APPROVED | DENIED` or `?taskId=<task id>` to get the extension request of a task or you can use both of them.
      
* `/extensionRequest/:id GET` for `SUPERUSER | APPOWNER` to get the extension requests with the `id`.

* `/extensionRequest/:id PATCH` for `SUPERUSER | APPOWNER` to update the extension requests with the `id`.

* `/extensionRequest/:id/status PATCH` for `SUPERUSER | APPOWNER` to update status the extension requests with the `id`.


test/integration/extensionRequests.test.js
```diff
@@ -0,0 +1,631 @@
...
+            return done(err);
+          }
+
+          expect(res).to.have.status(400);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("This task is assigned to some different user");
+          return done();
+        });
+    });
+    it("Should return fail response if the new ETA is invalid", function (done) {
```
heyrandhir, line 295: nit: since we are testing the new eta should lie after the old eta. can we rename the description to `Should return fail response if the new ETA falls before old ETA`.

test/integration/extensionRequests.test.js
```diff
@@ -0,0 +1,631 @@
...
+          }
+          expect(res).to.have.status(404);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.be.equal("Extension Request not found");
+          return done();
+        });
+    });
+  });
+
+  describe("GET /extensionRequest", function () {
```
heyrandhir, line 425: since this API is only to be hit by the super user. can we have a test when a normal user tries to use this API.

test/integration/extensionRequests.test.js
```diff
@@ -0,0 +1,631 @@
...
+        })
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+
+          expect(res).to.have.status(400);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal('"status" is required');
+          return done();
```
heyrandhir, line 545: what happens if the user passes both the `title` and `status` in the request body.

heyrandhir (changes requested):

test/integration/extensionRequests.test.js
```diff
@@ -0,0 +1,631 @@
...
+        lossRate: { [DINERO]: 1 },
+      },
+      {
+        title: "Test task",
+        type: "feature",
+        endsOn: 1234,
+        startedOn: 4567,
+        status: "active",
+        percentCompleted: 10,
+        participants: [],
```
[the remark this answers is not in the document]
ivinayakg: It wasn't needed anywhere therefore didn't added any

heyrandhir (approved):
LGTM

controllers/extensionRequests.js
```diff
@@ -0,0 +1,233 @@
...
+      extensionBody.oldEndsOn = task.endsOn;
+    }
+
+    const prevExtensionRequest = await extensionRequestsQuery.fetchExtensionRequests({
+      taskId: extensionBody.taskId,
+      assignee: extensionBody.assignee,
+    });
+    if (prevExtensionRequest.length) {
+      return res.boom.forbidden("An extension request for this task already exists.");
+    }
```
yesyash, line 42: What happens in the case when a user has already created an extension request but has exceeded the `newEndsOn` time and wants to further extend the time for the task? 

Are we allowing such extension requests to happen?

models/extensionRequests.js
```diff
@@ -0,0 +1,89 @@
...
+    return extensionRequestId;
+  } catch (err) {
+    logger.error("Error in updating task", err);
+    throw err;
+  }
+};
+
+/**
+ * Fetch all Extension Requests
+ * @param extensionRequestquery { Object }: Body of the extension request
```
yesyash, line 45: Can we further document the extensionRequestquery param to make it clear what properties are required? 

[Link to JSDoc](https://jsdoc.app/tags-param.html)

RitikJaiswal75:
Hi, @ivinayakg where is the data model description and the API contract for this new API endpoint?

controllers/extensionRequests.js
```diff
@@ -0,0 +1,233 @@
...
+    if (task.assignee !== assigneeUsername) {
+      return res.boom.badRequest("This task is assigned to some different user");
+    }
+    if (task.endsOn >= extensionBody.newEndsOn) {
+      return res.boom.badRequest("The value for newEndsOn should be greater than the previous ETA");
+    }
+    if (extensionBody.oldEndsOn !== task.endsOn) {
+      extensionBody.oldEndsOn = task.endsOn;
+    }
```
RitikJaiswal75, line 34: NIT: this line confuses a bit above this, you are using early return if the value is less, why can't I check if the value is more than oldEndsOn here?

This is handled by the early return but was a bit confusing to me.

controllers/extensionRequests.js
RitikJaiswal75: As per the description, there is a patch API for this.

controllers/extensionRequests.js
```diff
@@ -0,0 +1,233 @@
...
+        taskId: extensionBody.taskId,
+        createdBy: req.userData.id,
+      },
+      body: {
+        extensionRequestId: extensionRequest.id,
+        oldEndsOn: task.endsOn,
+        newEndsOn: extensionBody.newEndsOn,
+        assignee: extensionBody.assignee,
+      },
+    };
```
RitikJaiswal75, line 58: where is the status in the schema? Pending, approved, denied

routes/extensionRequests.js
```diff
@@ -0,0 +1,32 @@
...
+const authenticate = require("../middlewares/authenticate");
+const authorizeRoles = require("../middlewares/authorizeRoles");
+const { SUPERUSER, APPOWNER } = require("../constants/roles");
+const {
+  createExtensionRequest,
+  updateExtensionRequest,
+  updateExtensionRequestStatus,
+} = require("../middlewares/validators/extensionRequests");
+
+router.post("/", authenticate, createExtensionRequest, extensionRequests.createTaskExtensionRequest);
```
RitikJaiswal75, line 13: the controller checks if the user is a superuser, what about a normal user? Is it a superuser-only feature, couldn't we use authorize roles method?

ivinayakg:
> Hi, @ivinayakg wondering if you could guide me to the data model description and the API contract for this new API endpoint?

Sure. 
https://github.com/Real-Dev-Squad/website-api-contracts/pull/96

models/extensionRequests.js
ivinayakg: sure will do that

controllers/extensionRequests.js
```diff
@@ -0,0 +1,233 @@
...
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const createTaskExtensionRequest = async (req, res) => {
+  try {
+    const extensionBody = req.body;
+
+    if (req.userData.id !== extensionBody.assignee && !req.userData.roles?.super_user) {
+      return res.boom.forbidden("Only Super User can create an extension request for this task.");
+    }
```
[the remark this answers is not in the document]
ivinayakg: So example, I'm not a super user. There is a task assigned to you therefore only you or the superuser should be able to create an extension request for that. That's exactly what this line does, it forces a check where if someone tries to create an extension request for the task which is not assigned to them and they are not a superuser, they shouldn't be able to do so.

controllers/extensionRequests.js
ivinayakg: great point, let me recheck this. 👍

routes/extensionRequests.js
ivinayakg: even a normal user should be able to create an extension request right?

controllers/extensionRequests.js
ivinayakg: so basically the reason I do this, is to make sure even if the user gives the wrong `oldEndsOn` I will just replace it with the current one from the main task object

heyrandhir (approved):

models/extensionRequests.js
```diff
@@ -0,0 +1,88 @@
...
+const { buildExtensionRequests, formatExtensionRequest } = require("../utils/extensionRequests");
+
+/**
+ * Create Extension Request
+ * @param extensionRequestData { Object }: Body of the extension request
+ */
+const createExtensionRequest = async (extensionRequestData) => {
+  try {
+    const request = {
+      timestamp: (new Date().getTime() / 1000).toFixed(0),
```
yesyash, line 12: What are we trying to store here?
If I calculate the value (at the time of making this comment) it gives me `'1673074694'` which is: 
1. type of string 
2. If I do `new Date(1673074694)`  the corresponding output is `Tue Jan 20 1970 14:14:34`.

Is this what we want to store? 
Am I missing something here?

yesyash (approved):

Issue #830, opened 2022-11-22
After sign in with github the redirect is not working properly in development


ivinayakg:
After the setup of the new signup flow, when using the backend in development the redirect of the user after signing with `https://github.com/login/oauth/authorize?client_id=<CLIENT_ID>` was not working correctly (only locally).

prakashchoudhary07:
Closing this, as this issue is irrelevant now

PR #831, opened 2022-11-22
fixed bug on development in signin with github

ivinayakg:
- RFC - #830

### DIFF
--- a/controllers/auth.js
+++ b/controllers/auth.js
@@ -12,7 +12,11 @@ const authService = require("../services/authService");
 const githubAuth = (req, res, next) => {
   let userData;
   const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
-  const authRedirectionUrl = req.query.state ?? rdsUiUrl;
+  let authRedirectionUrl = req.query.state ?? rdsUiUrl;
+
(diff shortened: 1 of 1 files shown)


controllers/auth.js
```diff
@@ -12,7 +12,11 @@ const authService = require("../services/authService");
 const githubAuth = (req, res, next) => {
   let userData;
   const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
-  const authRedirectionUrl = req.query.state ?? rdsUiUrl;
+  let authRedirectionUrl = req.query.state ?? rdsUiUrl;
+
+  if (process.env.NODE_ENV === "development" && !req.query.state) {
+    authRedirectionUrl = "/healthcheck";
+  }
```
ankushdharkar, line 19: ❓Why are we doing this? Is this absolutely necessary?

ivinayakg: So when someone sets up their backend locally and tries to sign in with github into their backend, as there is no query state it redirects the user to the root path which gives a 404 and that has confused a lot of people in thinking that they have done something wrong. 
This will fix that.

prakashchoudhary07:
Closing this, as this issue is irrelevant now

PR #840, opened 2022-11-27
Add new route for login with github

### LINKED ISSUE #839: Add login route for login with github
Currently, to authenticate, we have a link `https://github.com/login/oauth/authorize?client_id=<CLIENT_ID>`, which is to be hit when we need to authenticate and <CLIENT_ID> changes for every environment and users running it locally

And on all front-end apps, this is hard-coded, so if we need to do local development, we need to change links at multiple places where that is hard-coded and if in the future for some reason clientId is changed, then we will manually need to change it from all hard-coded places

This should be handled from the backend, where a login route is hit, and the rest of the things are taken care we don't need to change the <CLIENT_ID> at every place where the hard-coded URL its used, as the login route, is used it is will automatically vary based of configurations set in backend development

prakashchoudhary07:
Resolves #839 

Added a new route `/auth/github/login` which will handle the login with GitHub

Now we don't need to set CLIENT_ID in hardcode the login URL `https://github.com/login/oauth/authorize?client_id=<CLIENT_ID>` we can just hit `/auth/github/login` and it will handle the login based on the clientId and secrets set in the config file of the user

### DIFF
--- a/routes/auth.js
+++ b/routes/auth.js
@@ -1,6 +1,9 @@
 const express = require("express");
+const passport = require("passport");
 const router = express.Router();
 const auth = require("../controllers/auth");
+
+router.get("/github/login", passport.authenticate("github", { scope: ["user:email"] }));
 
 router.get("/github/callback", auth.githubAuth);
 


routes/auth.js
```diff
@@ -1,7 +1,10 @@
 const express = require("express");
+const passport = require("passport");
 const router = express.Router();
 const auth = require("../controllers/auth");
 
+router.get("/github/login", passport.authenticate("github", { scope: ["user:email"] }));
```
RitikJaiswal75, line 6: we can also keep a failure parameter here so that the user can know that login was not successful.

Something like this: https://github.com/jaredhanson/passport-github/blob/dee535a1d47ade9a15916a4f92522292b2e0ef46/README.md?plain=1#L104

prakashchoudhary07: Thanks for the suggestion 🎉 , but if unsuccessful then it will give 401 Error
That has been handled here
https://github.com/Real-Dev-Squad/website-backend/blob/6ec922dd9cbff63045b43cab187d7b60d21e4bb3/controllers/auth.js#L18-L22

RitikJaiswal75 (approved):

Issue #857, opened 2022-12-03
Refactor the user-status endpoint


heyrandhir:
Modify the following points in user-status endpoint.
- [ ] Add new route `/self` for GET and PATCH
- [ ] Change route name from `/user-status` to `/users/status`
- [ ] Update tests for the new routes
- [ ] Change the collection name from `userStatus` to `usersStatus`

Manas-E:
Is it open I want to work on this?
@heyrandhir

heyrandhir:
Hi @Manas-E PR is already raised by me for this one. Sorry, I should have assigned this to myself. I had missed that.

PR #858, opened 2022-12-03
Adds new routes & tests for users/status Endpoint

heyrandhir:
Closes https://github.com/Real-Dev-Squad/website-backend/issues/857
- [x] Adds new route `/self` for GET and PATCH
- [x] Changes route name from `/user-status` to `/users/status`
- [x] Updated tests for the new routes
- [x] Change the collection name from `userStatus` to `usersStatus`

### DIFF
--- a/controllers/userStatus.js
+++ b/controllers/userStatus.js
@@ -41,7 +41,7 @@ const getUserStatus = async (req, res) => {
     if (userId) {
       const userData = await userStatusModel.getUserStatus(userId);
       const { userStatusExists, id, data } = userData;
-      const responseObject = { id, userId, data: null, message: null };
+      const responseObject = { id, userId, data: null, message: "" };
       if (data) responseObject.data = data;
       let statusCode;
       if (userStatusExists) {
@@ -93,7 +93,7 @@ const updateUserStatus = async (req, res) => {
       const dataToUpdate = req.body;
       const updateStatus = await userStatusModel.updateUserStatus(userId, dataToUpdate);
       const { userStatusExists, userStatusUpdated, id, data } = updateStatus;
-      const responseObject = { id, userId, data: null, message: null };
+      const responseObject = { id, userId, data: null, message: "" };
       let statusCode;
       if (data) responseObject.data = data;
       if (userStatusExists) {
--- a/controllers/userStatus.js
+++ b/controllers/userStatus.js
@@ -36,20 +36,28 @@ const deleteUserStatus = async (req, res) => {
  */
 const getUserStatus = async (req, res) => {
   try {
-    const { userId } = req.params;
-    const userData = await userStatusModel.getUserStatus(userId);
-    const { userStatusExists, id, data } = userData;
-    let responseObject = { id, userId };
-    if (data) responseObject = { ...responseObject, ...data };
-    let statusCode;
-    if (userStatusExists) {
-      statusCode = 200;
-      responseObject.message = "User Status found successfully.";
+    let userId;
+    if (req.route.path === "/self") {
+      userId = req.userData.id;
     } else {
-      responseObject.message = "User Status couldn't be found.";
-      statusCode = 404;
+      userId = req.params.userId;
     }
-    return res.status(statusCode).json(responseObject);
+    if (userId) {
+      const userData = await userStatusModel.getUserStatus(userId);
+      const { userStatusExists, id, data } = userData;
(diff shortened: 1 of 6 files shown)


controllers/userStatus.js
```diff
@@ -84,28 +92,36 @@ const getAllUserStatus = async (req, res) => {
...
-      statusCode = 200;
+    let userId;
+    if (req.route.path === "/self") {
+      userId = req.userData.id;
     } else {
-      if (userStatusUpdated) {
-        responseObject.message = "User Status created successfully.";
-        statusCode = 201;
+      userId = req.params.userId;
+    }
```
yesyash, line 100: I see that this is being repeated in multiple controllers, would it be better to have a `getUserId` function that implements this logic and returns us the user id?

controllers/userStatus.js
```diff
@@ -84,28 +92,36 @@ const getAllUserStatus = async (req, res) => {
...
       } else {
-        responseObject.message = "User Status couldn't be created due to incomplete request body.";
-        statusCode = 400;
+        if (userStatusUpdated) {
+          responseObject.message = "User Status created successfully.";
+          statusCode = 201;
+        } else {
+          responseObject.message = "User Status couldn't be created due to incomplete request body.";
+          statusCode = 400;
+        }
```
yesyash, line 120: Same `responseObject` schema consistency.

Maheima:
I like this idea of route better than `user-status`. Will approve it once Yash's requested changes are resolved

controllers/userStatus.js
```diff
@@ -36,20 +36,28 @@ const deleteUserStatus = async (req, res) => {
...
+      let responseObject = { id, userId };
+      if (data) responseObject = { ...responseObject, ...data };
+      let statusCode;
+      if (userStatusExists) {
+        statusCode = 200;
+        responseObject.message = "User Status found successfully.";
+      } else {
+        responseObject.message = "User Status couldn't be found.";
+        statusCode = 404;
+      }
```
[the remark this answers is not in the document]
heyrandhir: I agree with your point but since I am not adding `data` as a direct key here instead I am destructuring the `data` object so that all the keys of `data` become the primary keys of `responseObject` as well. So if the `data` object exist then only those keys will be added if not then we are adding a relevant message `User Status couldn't be found.`

controllers/userStatus.js
heyrandhir: I agree with your point but since I am not adding `data` as a direct key here instead I am destructuring the `data` object so that all the keys of `data` become the primary keys of `responseObject` as well. So if the `data` object exist then only those keys will be added if not then we are adding a relevant message `User Status couldn't be found.`

controllers/userStatus.js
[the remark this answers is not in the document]
yesyash: I was thinking if we can have something like this. 

{ data: ...data} (where data is the key and we destructure the data obj from response) if the data in response obj does not exist we pass null instead along with the custom message.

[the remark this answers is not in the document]
heyrandhir: since `data` itself is an object. executing `{ data: ...data}` will throw syntax error. the other choice we have is to pass the entire object as `{data: data}` but it will nest the status object 1 level inside the `data` key instead of directly being accessible at the parent level.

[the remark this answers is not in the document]
heyrandhir: yeah, it would be better to add 1 level of nesting to have a consistent schema. Updated in the latest code.

controllers/userStatus.js
```diff
@@ -41,8 +41,8 @@ const getUserStatus = async (req, res) => {
     if (userId) {
       const userData = await userStatusModel.getUserStatus(userId);
       const { userStatusExists, id, data } = userData;
-      let responseObject = { id, userId, data: null };
-      if (data) responseObject = { ...responseObject, data };
+      const responseObject = { id, userId, data: null, message: null };
```
yesyash, line 44: Nitpick. 
Can the message be an empty string than null?

Example: message: ''

yesyash (approved):

controllers/userStatus.js
heyrandhir: yeah, empty string will be better. will update. Thanks.

yesyash (approved):

Maheima (approved):

poonammohata:
LGTM

PR #862, opened 2022-12-07
Status API patch refactor

heyrandhir:
This PR fixes a bug that earlier we were not allowing to create new user status with a partial object (i.e only if current status or monthly hours was present). Now we have removed this condition. This PR once merged will allow Users to create a new User Status Object with partial data (i.e either the key current status or monthly hours being present). 

This is a stacked PR and should be merged after https://github.com/Real-Dev-Squad/website-backend/pull/858

### DIFF
--- a/test/integration/userStatus.test.js
+++ b/test/integration/userStatus.test.js
@@ -73,9 +73,9 @@ describe("UserStatus", function () {
           expect(res).to.have.status(200);
           expect(res.body).to.be.a("object");
           expect(res.body.message).to.equal("User Status found successfully.");
-          expect(res.body.userId).to.equal(userId);
-          expect(res.body).to.have.property("monthlyHours");
-          expect(res.body).to.have.property("currentStatus");
+          expect(res.body.data.userId).to.equal(userId);
+          expect(res.body.data).to.have.property("monthlyHours");
+          expect(res.body.data).to.have.property("currentStatus");
           return done();
         });
     });
@@ -92,8 +92,8 @@ describe("UserStatus", function () {
           expect(res.body).to.be.a("object");
           expect(res.body.message).to.equal("User Status found successfully.");
           expect(res.body.userId).to.equal(userId);
-          expect(res.body).to.have.property("monthlyHours");
-          expect(res.body).to.have.property("currentStatus");
+          expect(res.body.data).to.have.property("monthlyHours");
+          expect(res.body.data).to.have.property("currentStatus");
           return done();
         });
     });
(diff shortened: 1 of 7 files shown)


test/integration/userStatus.test.js
```diff
@@ -41,11 +40,11 @@ describe("UserStatus", function () {
     await cleanDb();
   });
 
-  describe("GET /user-status", function () {
+  describe("GET /users/status", function () {
```
RitikJaiswal75, line 43: since we are using `/users/` it gives impression that it could be in user.js in test

test/integration/userStatus.test.js
```diff
@@ -75,14 +91,14 @@ describe("UserStatus", function () {
           expect(res.body).to.be.a("object");
           expect(res.body.message).to.equal("User Status found successfully.");
           expect(res.body.userId).to.equal(userId);
-          expect(res.body).to.have.property("monthlyHours");
-          expect(res.body).to.have.property("currentStatus");
+          expect(res.body.data).to.have.property("monthlyHours");
+          expect(res.body.data).to.have.property("currentStatus");
```
RitikJaiswal75, line 95: the test does not tell me that the data returned is for self. it can return for any person if I pass the cookie. lets say i pass in your cookie then i can get your details.

test/integration/userStatus.test.js
```diff
@@ -125,45 +159,44 @@ describe("UserStatus", function () {
         });
     });
 
-    it("Should return 401 for unauthorized request", function (done) {
+    it("Should update the User Status when requested by Super User", function (done) {
```
RitikJaiswal75, line 162: shouldn't users be able to change their ooo state why do we need a superuser to do so?

test/integration/userStatus.test.js
```diff
@@ -204,7 +237,7 @@ describe("UserStatus", function () {
     it("Should return 200 for deletion by Super User", function (done) {
```
RitikJaiswal75, line 237: can we add one more test for unauthorized i.e without passing any cookie?

test/integration/userStatus.test.js
heyrandhir: since the route is `/self`. we take the userId from req.userData hence it will always be the user status data for the person who has requested it.

test/integration/userStatus.test.js
heyrandhir: a user can update their user status while a super user can update anyone's user status.

test/integration/userStatus.test.js
heyrandhir: It's covered in the test case above if the user passes the incorrect cookie it will throw `Unauthorized User`

RitikJaiswal75 (approved):
LGTM

PR #863, opened 2022-12-08
Endpoint/username query

Muralidhar22:


prakashchoudhary07 (changes requested):

test/integration/users.test.js
```diff
@@ -11,14 +11,91 @@ const cleanDb = require("../utils/cleanDb");
...
+            item.value === "AN" ||
+            typeof item.value === "number" ||
+            item.value === ""
+          ) {
+            expect(res).to.have.status(200);
+            expect(res.body).to.be.a("object");
+            expect(res.body.message).to.equal("Users returned successfully!");
+            expect(res.body.users).to.be.a("array");
+          }
+
```
prakashchoudhary07, line 92: I think we are complicating things here by running it in the loop
Like conditional branching for each test case, 
This should be separately handled

controllers/users.js
```diff
@@ -64,7 +64,12 @@ const getUserById = async (req, res) => {
 const getUsers = async (req, res) => {
   try {
     const allUsers = await userQuery.fetchUsers(req.query);
-
+    if (!allUsers.length) {
+      return res.json({
+        message: "No users are present",
+        users: allUsers,
+      });
+    }
```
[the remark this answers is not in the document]
Muralidhar22: is it fine to send `"Users returned successfully!"` even when there are no users found with the params value?

models/users.js
```diff
@@ -119,16 +119,21 @@ const getSuggestedUsers = async (skill) => {
...
  */
 const fetchUsers = async (query) => {
   try {
-    const snapshot = await userModel
+    let dbQuery = userModel
       .limit(parseInt(query.size) || 100)
-      .offset((parseInt(query.size) || 100) * (parseInt(query.page) || 0))
-      .get();
-
+      .offset((parseInt(query.size) || 100) * (parseInt(query.page) || 0));
```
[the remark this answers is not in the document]
Muralidhar22: don't know will look into this

test/integration/users.test.js
```diff
@@ -11,14 +11,91 @@ const cleanDb = require("../utils/cleanDb");
...
+    it(`Test Scenario ${i}: ${item.desc}`, function (done) {
+      chai
+        .request(app)
+        .get("/users")
+        .query({ search: item.value })
+        .set("cookie", `${cookieName}=${jwt}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
```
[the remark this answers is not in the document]
Muralidhar22: okay will write individual tests.

test/integration/users.test.js
Muralidhar22: okay will write individual tests.

test/integration/users.test.js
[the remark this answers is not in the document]
Muralidhar22: done

models/users.js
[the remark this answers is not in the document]
Muralidhar22: will raise a new PR for this change

test/integration/users.test.js
Muralidhar22: done

controllers/users.js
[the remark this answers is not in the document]
Muralidhar22: changed the message, please review again

controllers/users.js
```diff
@@ -64,7 +64,9 @@ const getUserById = async (req, res) => {
 const getUsers = async (req, res) => {
   try {
     const allUsers = await userQuery.fetchUsers(req.query);
-
+    if (!allUsers.length) {
+      return res.boom.notFound("User doesn't exist");
+    }
     return res.json({
       message: "Users returned successfully!",
       users: allUsers,
```
prakashchoudhary07, line 72: I think here we should return an empty array if no users are found for the request with a status code of 200 ok, instead of returning 404, we are returning multiple users, so it should be like [ ] or [user1] or [user1, user2...]

test/integration/users.test.js
```diff
@@ -350,6 +350,134 @@ describe("Users", function () {
...
+        username: "23ankur",
+        first_name: "Ankur",
+        last_name: "Narkhede",
+        yoe: 0,
+        img: "./img.png",
+        linkedin_id: "ankurnarkhede",
+        github_id: "ankur1234",
+        github_display_name: "ankur-xyz",
+        phone: "1234567890",
+        email: "abc@gmail.com",
```
prakashchoudhary07, line 377: This is already available in fixtures, so we should be using it from there, If not available move this to fixtures

models/users.js
```diff
@@ -119,16 +119,26 @@ const getSuggestedUsers = async (skill) => {
...
+ * @param query { Object }: Filter for users
  * @return {Promise<userModel|Array>}
  */
 const fetchUsers = async (query) => {
+  const addSearchDbQuery = (dbQuery) => {
+    return dbQuery
+      .orderBy("username")
+      .startAt(query.search.toLowerCase().trim())
+      .endAt(query.search.toLowerCase().trim() + "\uf8ff");
+  };
```
prakashchoudhary07, line 131: Should this be in fetchUser function?

test/integration/users.test.js
```diff
@@ -350,6 +350,134 @@ describe("Users", function () {
...
+    beforeEach(async function () {
+      await addOrUpdate(userSearchParamsData[0]);
+      await addOrUpdate(userSearchParamsData[1]);
+    });
+
+    afterEach(async function () {
+      await cleanDb();
+    });
+
+    it(`${searchParamValues[0].desc}`, function (done) {
```
prakashchoudhary07, line 390: I don't know what the test does here, from its name.
We should write the test name, I feel it's just confusing to maintain a separate file for test names.

prakashchoudhary07 (changes requested):

controllers/users.js
Muralidhar22: removed it

test/integration/users.test.js
Muralidhar22: done

test/integration/users.test.js
Muralidhar22: done

models/users.js
Muralidhar22: moved to utils

test/fixtures/user/search.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Sample search values for tests
+ * @return  {object}
+ */
+
+module.exports = () => {
+  return ["an", "AN", "", "mu"];
+};
```
prakashchoudhary07, line 8: Is this adding value?

test/integration/users.test.js
```diff
@@ -350,6 +351,89 @@ describe("Users", function () {
...
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Users returned successfully!");
+          expect(res.body.users).to.be.a("array");
+
+          return done();
```
prakashchoudhary07, line 397: Here only checking if array or not, 
Can we check, if correct data is being returned

utils/users.js
```diff
@@ -93,10 +93,18 @@ function getLowestLevelSkill(skills) {
   return { skill, level };
 }
 
+const addSearchDbQuery = (dbQuery, requestQuery) => {
+  return dbQuery
+    .orderBy("username")
+    .startAt(requestQuery.search.toLowerCase().trim())
+    .endAt(requestQuery.search.toLowerCase().trim() + "\uf8ff");
+};
```
prakashchoudhary07, line 101: NIT: I don't think we needed to extract this

prakashchoudhary07 (changes requested):

utils/users.js
Muralidhar22: moved this seperately to make `model` less cumbersome

test/fixtures/user/search.js
Muralidhar22: changed into an object, i believe keeping all of the data used for testing at single place would be good

test/integration/users.test.js
Muralidhar22: done

models/users.js
```diff
@@ -119,16 +119,26 @@ const getSuggestedUsers = async (skill) => {
...
- * @param query { Object }: Filter for users data
+ * @param query { Object }: Filter for users
  * @return {Promise<userModel|Array>}
  */
 const fetchUsers = async (query) => {
+  const appendUsernamePrefixQuery = (dbQuery) => {
+    return dbQuery
+      .orderBy("username")
+      .startAt(query.search.toLowerCase().trim())
+      .endAt(query.search.toLowerCase().trim() + "\uf8ff");
```
vinayak-trivedi, line 130: can you just explain this, not able to get what you are doing here

Muralidhar22: if `search` param exists, then we will add this query

vinayak-trivedi, line 130: can you explain what this does?

vinayak-trivedi, line 130: can you explain this endAt line?

Muralidhar22: https://firebase.google.com/docs/database/rest/retrieve-data#range-queries

vinayak-trivedi: thanks

vinayak-trivedi (approved):
LGTM

PR #867, opened 2022-12-11
Removed Node modules & test files from production build.

### LINKED ISSUE #812: Adding Bundler for production build
- This would help in compiling multiple JS file into a single file.
- It creates a dependency relationship graph
- It improves the load time.
- The generated file would be present in `/dist`

vikhyat187:
Closes #812 

The task was to remove the dev dependencies from the production build & test cases.
1. I've used webpack to bundle the files and the output file would be stored in ```/dist``` folder.
2. The script to build is ```yarn run build```.
3. The script to run the bundle is ```yarn run startbuild```.

To analyze the bundle we can use https://chrisbateman.github.io/webpack-visualizer/

[stats1.txt](https://github.com/Real-Dev-Squad/website-backend/files/10239774/stats1.txt) **Before bundling the stats file size was 22MB.**

[stats.txt](https://github.com/Real-Dev-Squad/website-backend/files/10202818/stats.txt) **After removing node modules the size became 650KB**

Download the ```stats.txt``` file and rename to ```stats.json```, upload to above link to analyze bundle.

[Links referred](https://blog.bitsrc.io/6-tools-and-techniques-to-analyze-webpack-bundle-size-817337f8cf91)

before_node_modules

after_bundling

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,33 @@
+const path = require('path');
+const nodeExternals = require('webpack-node-externals');
+
+module.exports = {
+    entry:'./server.js',
+    target:'node',
+    output: {
+        filename: 'bundle.js',
+        path: path.resolve(__dirname,'dist')
+    },
+    resolve:{
+        modules:[
+            path.resolve(__dirname, 'node_modules'),
+        ],
+       
+    },
+    module:{
+        rules: [
+            {
+                test:/\.js$/,
+                exclude:[
+                    /node_modules/,
+                    /test/
+                ]
+            }
+        ]
+    },
+
+    externals:[
+        nodeExternals()
+    ]
+}
+
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,33 +1,24 @@
-const path = require('path');
-const nodeExternals = require('webpack-node-externals');
+const path = require("path");
+const nodeExternals = require("webpack-node-externals");
 
 module.exports = {
-    entry:'./server.js',
-    target:'node',
-    output: {
-        filename: 'bundle.js',
-        path: path.resolve(__dirname,'dist')
-    },
-    resolve:{
-        modules:[
-            path.resolve(__dirname, 'node_modules'),
-        ],
-       
-    },
-    module:{
-        rules: [
-            {
-                test:/\.js$/,
-                exclude:[
-                    /node_modules/,
-                    /test/
-                ]
-            }
-        ]
-    },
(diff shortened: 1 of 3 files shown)


vikhyat187:
> At what point are dev dependencies removed?

On building the bundled file.

MehulKChaudhari (approved):
LGTM.

prakashchoudhary07:
@vikhyat187 This look good, but would want to test the changes before pushing it to DEV

webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+  },
+  resolve: {
+    modules: [path.resolve(__dirname, "node_modules")],
+  },
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        exclude: [/node_modules/, /test/],
+      },
```
prakashchoudhary07, line 19: Do we remove tests as well,
Do we not run the tests after the build is done?

How is this generally do, like do tests run and then build takes place or do build run then test runs?

vikhyat187: Since the test supporting dependencies were dev and we were removing them while bundling, so test wouldn;t work.

vikhyat187:
> > @vikhyat187 This look good, but would want to test the changes before pushing it to DEV
> 
> Can we deploy it in a separate branch and test it out

Yes sure.

prakashchoudhary07:
How are we going to remove node modules when deploying in production?

vikhyat187:
> How are we going to remove node modules when deploying in production?

"webpack-node-externals": this helps in removing the dev dependencies.

prakashchoudhary07:
I think we should remove dev dependencies from the production, we will have to look into this, 
Either run some command to remove dev dependencies or only install dependencies in prod and then we can merge this PR

vikhyat187:
> I think we should remove dev dependencies from the production, we will have to look into this, Either run some command to remove dev dependencies or only install dependencies in prod and then we can merge this PR

Will check this out and get back.

vikhyat187:
> > I think we should remove dev dependencies from the production, we will have to look into this, Either run some command to remove dev dependencies or only install dependencies in prod and then we can merge this PR
> 
> Will check this out and get back.

https://docs.npmjs.com/cli/v8/commands/npm-prune this one doesn't install dev dependencies, lets try it out @prakashchoudhary07

vikhyat187:
Closing this PR as, this doesn't improve any performance in backend server.

Issue #850, opened 2022-12-01
Join Section | multiple data creation


RitikJaiswal75:
Every time the User sends in to submit a request a new document is created. 
We are sending a PUT request every time the user sends the data from the front end.

Solution:
We need to check if the data for userId is already present or not.
If already present fail the request saying data already present unless the profile is rejected by the superuser and the user is asked to re-fill the join data.

satyam73:
Hello @RitikJaiswal75 I want to work on this issue.

RitikJaiswal75:
Hello @satyam73 how is the work going on this? any blockers?

satyam73:
> Hello @satyam73 how is the work going on this? any blockers?

Can I have more context for this?

RitikJaiswal75:
please link your solution branch here

satyam73:
> Contributor
Did you mean PR?

satyam73:
> Contributor
Did you mean PR?

PR #870, opened 2022-12-12
Fixed join section multiple data creation issue

satyam73:
Hello @RitikJaiswal75 , I have fixed  the changes regarding the issue #850 .
closes: #850 
Now if the user makes put request at the route users/self/intro at first the data is created and the message is returned ( as in image below ).
image

And when the user has one join data present in the database it returns the message ( as in image below ).
image

Please look into it and tell me if something needs to be rectified or updated.
Thanks!

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -331,6 +331,14 @@ const rejectProfileDiff = async (req, res) => {
 const addUserIntro = async (req, res) => {
   try {
     const rawData = req.body;
+    const joinData = await userQuery.getJoinData(req.userData.id);
+
+    if (joinData.length === 1) {
+      return res.status(409).json({
+        message: "User data is already present!",
+      });
+    }
+
     const data = {
       userId: req.userData.id,
       biodata: {
--- a/test/integration/users.test.js
+++ b/test/integration/users.test.js
@@ -17,7 +17,6 @@ const config = require("config");
 const joinData = require("../fixtures/user/join");
 const { addJoinData, addOrUpdate } = require("../../models/users");
 const cookieName = config.get("userToken.cookieName");
-
 chai.use(chaiHttp);
 
 describe("Users", function () {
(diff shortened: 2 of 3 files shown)


controllers/users.js
```diff
@@ -331,6 +331,14 @@ const rejectProfileDiff = async (req, res) => {
 const addUserIntro = async (req, res) => {
   try {
     const rawData = req.body;
+    const joinData = await userQuery.getJoinData(req.userData.id);
+
+    if (joinData.length === 1) {
```
RitikJaiswal75, line 336: what happens if a person added 3 join data before your PR goes in?
wont the length === 1 fail?

RitikJaiswal75 (changes requested):
Please write tests for the same

RitikJaiswal75:
updated your description, added a `closes` keyword to automatically link it to the issue this makes sure as soon as the branch is merged the issue ticket gets closed.

controllers/users.js
satyam73: > what happens if a person added join data thrice before your PR goes in? won't the length === 1 fail?

If the user has previous records it will fetch one result which will be length === 1 and returns from there on.

test/integration/users.test.js
```diff
@@ -351,6 +350,24 @@ describe("Users", function () {
...
+          }
+          if (joinData()[1].length === 1) {
+            expect(res).to.have.status(409);
+            expect(res.body).to.be.a("object");
+            expect(res.body.message).to.equal("User data is already present!");
+          }
+          return done();
+        });
+    });
+
```
RitikJaiswal75, line 370: Where are you inserting the data? test could be adding data to db before this and then trying to add it again in describe

satyam73: Data insertion to db is in the other test case.
Should I add this test in the same test and add only the if condition?

RitikJaiswal75: lets discuss it over a call

also please try to take 2 reviews, not just mine.

satyam73: Okay👍🏻

yesyash (approved):

heyrandhir (approved):
LGTM

RitikJaiswal75 (approved):
LGTM

Issue #842, opened 2022-11-27
Add a max size in the `/users` query param `size`


heyrandhir:
the query param `size` should have an upper limit on the no of users requested from the `/users` endpoint

heyrandhir:
Hi @shmbajaj you are right this issue deals with pagination. Pagination is already implemented for `/users`. However, there is a query parameter `size` which currently has no upper cap. Someone can pass `/users?size=10000` which will end up consuming all our firebase reads. Hence we need an upper cap on the size param which can be requested. Hope I was able to make it clear.

shmbajaj:
@heyrandhir can you assign this to me....

heyrandhir:
> @heyrandhir can you assign this to me....

Done

shmbajaj:
[PR: Max size for size query param](https://github.com/Real-Dev-Squad/website-backend/pull/873)

PR #873, opened 2022-12-13
feat: users ep GET request, query-param size validation

shmbajaj:
## What is the change

- Min(1) and Max(100) limit on number of users requested using query-param size 
- Validation on query params _size_ and _page_ using middelware **joi**

```javascript
size: /^[1-9]\d?$|^100$/
page: /^\d$|^[1-9]\d?/
``` 

## Before the change

- No limit on number of users requested using query-param size 
- No validation on query params _size_ and _page_

### Issue
[Issue: 842](https://github.com/Real-Dev-Squad/website-backend/issues/842)

### Endpoint

```javascript
GET /users?size=<value>&page=<value>
```
<small>params: \*size, page\? </small>

#### Dev Tested

- [ ] Yes
- [x] No

#### Platforms

- [X] Postman

#### Info
- Failing Tests have nothing to do with this PR.
- [API Contract](https://github.com/Real-Dev-Squad/website-api-contracts/pull/102)

### DIFF
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -77,8 +77,51 @@ const validateJoinData = async (req, res, next) => {
     res.boom.badRequest(error.details[0].message);
   }
 };
+
+/**
+ * Validates getting users query
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ * @param next {Object} - Express middelware function
+ */
+async function getUsers(req, res, next) {
+  const schema = joi
+    .object()
+    .strict()
+    .keys({
+      size: joi
+        .string()
+        .required()
+        .pattern(/^[1-9]\d?$|^100$/)
+        .messages({
+          "string.base": "size should be a type of string",
+          "string.empty": "size must contain value in range 1-100",
+          "string.pattern.base": "size must be in range 1-100",
+          "any.required": "size is required",
+        }),
+      page: joi
+        .string()
+        .optional()
+        .pattern(/^\d$|^[1-9]\d?/)
+        .messages({
+          "string.base": "page should be a type of string",
+          "string.empty": "page must contain value in range 0-99",
+          "string.pattern.base": "page must be in range 0-99",
+        }),
+    });
+  try {
+    await schema.validateAsync(req.query);
+    next();
+  } catch (error) {
+    logger.error(`Error in getting users: ${error}`);
+    res.boom.badRequest(error.details[0].message);
+  }
+}
+
 module.exports = {
   updateUser,
   updateProfileURL,
   validateJoinData,
+  getUsers,
 };
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -95,19 +95,16 @@ async function getUsers(req, res, next) {
         .optional()
         .pattern(/^[1-9]\d?$|^100$/)
         .messages({
-          "string.base": "size should be a type of string",
           "string.empty": "size must contain value in range 1-100",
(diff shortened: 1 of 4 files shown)


middlewares/validators/user.js
```diff
@@ -77,8 +77,51 @@ const validateJoinData = async (req, res, next) => {
...
+    .keys({
+      size: joi
+        .string()
+        .optional()
+        .pattern(/^[1-9]\d?$|^100$/)
+        .messages({
+          "string.base": "size should be a type of string",
+          "string.empty": "size must contain value in range 1-100",
+          "string.pattern.base": "size must be in range 1-100",
+          "any.required": "size is required",
```
heyrandhir, line 101: I don't think we should make it mandatory for the user to pass `size` always. by default, the size is 100 if not provided by the user.

middlewares/validators/user.js
```diff
@@ -77,8 +77,51 @@ const validateJoinData = async (req, res, next) => {
...
+        }),
+      page: joi
+        .string()
+        .optional()
+        .pattern(/^\d$|^[1-9]\d?/)
+        .messages({
+          "string.base": "page should be a type of string",
+          "string.empty": "page must contain value in range 0-99",
```
heyrandhir, line 109: This will not work if we have more than 10k users. I think only checking if the page is a number should be sufficient.

heyrandhir:
can we have the test cases as well for this change.

middlewares/validators/user.js
alishvagadiya: If the size is not provided, then there should be a Default/ideal size; here, maybe we can say 10 records; 
we should not put max size as the default or ideal size. It will increase the unnecessary load. 
If the user wants more records, we respond to requests.

alishvagadiya (changes requested):

middlewares/validators/user.js
shmbajaj: Now page can be either 0 or any number >0.
cc: @alishvagadiya

middlewares/validators/user.js
shmbajaj: `size` is optional and default value is 10.

heyrandhir: I checked this point with Ankush today. Changing the default size would unnecessarily lead to more network calls. Hence we are going to stick with 100 as earlier

shmbajaj: Changed backed to 100.

heyrandhir:
can you update this change in api contracts as well if not updated.

shmbajaj:
> can you update this change in api contracts as well if not updated.

Hey please api-contracts, PR from here: [PR: /users api contract update](https://github.com/Real-Dev-Squad/website-api-contracts/pull/102)

test/integration/users.test.js
```diff
@@ -123,6 +123,76 @@ describe("Users", function () {
...
+        })
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Users returned successfully!");
+          expect(res.body.users).to.be.a("array");
```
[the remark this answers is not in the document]
shmbajaj: https://github.com/Real-Dev-Squad/website-backend/pull/873/commits/e0b9cb895bf375d457997217534ad733dea29822

heyrandhir (approved):
LGTM

yesyash (approved):

PR #879, opened 2022-12-17
reason not mandatory when user is OOO for less than 3 days 

heyrandhir:
- [x] reason not mandatory when the user is OOO for less than 3 days 
- [x] added a new test case to handle this case

### DIFF
--- a/test/fixtures/userStatus/userStatus.js
+++ b/test/fixtures/userStatus/userStatus.js
@@ -12,7 +12,7 @@ const userStsDataForNewUser = {
   },
 };
 
-const userStsDataForOooStateForShortDuration = {
+const oooUserStsDataForShortDuration = {
   currentStatus: {
     message: "",
     state: "OOO",
@@ -62,7 +62,7 @@ const invalidUserStsDataforUpdate = {
 module.exports = {
   userStsDataForNewUser,
   userStsDataForOooState,
-  userStsDataForOooStateForShortDuration,
+  oooUserStsDataForShortDuration,
   validUserStsDataforUpdate,
   invalidUserStsDataforUpdate,
 };
--- a/test/fixtures/userStatus/userStatus.js
+++ b/test/fixtures/userStatus/userStatus.js
@@ -12,6 +12,19 @@ const userStsDataForNewUser = {
   },
 };
 
+const userStsDataForOooStateForShortDuration = {
+  currentStatus: {
+    message: "",
+    state: "OOO",
+    updatedAt: 1673893800000,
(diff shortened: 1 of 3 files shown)


alishvagadiya (changes requested):

test/fixtures/userStatus/userStatus.js
```diff
@@ -12,6 +12,19 @@ const userStsDataForNewUser = {
   },
 };
 
+const userStsDataForOooStateForShortDuration = {
+  currentStatus: {
+    message: "",
```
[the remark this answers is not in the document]
heyrandhir: If the user is OOO for less than 3 days then adding a message is not mandatory. I am testing the same. Hence left the it blank.

alishvagadiya (changes requested):
Is there a test case when OOO duration is more than 3 days, and the reason is not mentioned, and failed/error Is the result, test cases are added?

If not please add test cases for that.

heyrandhir:
> Is there a test case when OOO duration is more than 3 days, and the reason is not mentioned, and failed/error Is the result, test cases are added?
> 
> If not please add test cases for that.

Hi @alishvagadiya . Added the test case for it in the stacked PR which will be merged post this. This test can be found [here](https://github.com/Real-Dev-Squad/website-backend/pull/884/files#diff-734be2449d7d802ef008723d52ed7c5fae9f44665f12d97b6f89c6aad6166bc3R381-R397)

rohan09-raj (approved):
Nice job in handling the edge cases 💯

test/fixtures/userStatus/userStatus.js
```diff
@@ -12,6 +12,19 @@ const userStsDataForNewUser = {
...
+    state: "OOO",
+    updatedAt: 1673893800000,
+    from: 1673893800000,
+    until: 1674066600000,
+  },
+  monthlyHours: {
+    updatedAt: 1668215609000,
+    committed: 40,
```
prakashchoudhary07, line 25: Why are we passing monthlyHours when setting OOO?
Am I missing something here

prakashchoudhary07 (changes requested):

test/fixtures/userStatus/userStatus.js
heyrandhir: this data I have used it here for testing if the endpoint is able to store the data in the collection document when requested by an authorized user or superuser. The test will work with both the partial status data as well as status data with monthly hours. Few of the tests are using complete data while few are using partial data.

PR #892, opened 2022-12-24
Created Cloudflare cache APIs

### LINKED ISSUE #292: Purge Cache for user specific Page
Right now `https://members.realdevsquad.com/${username}` page is served with Cloudflare. So it's cached copy is served every time. On changing, any user details it doesn't get reflected on the user page.

We needed an API that clears the cache at Cloudfare and we can see the updated Page.

We have created an endpoint `/cache/clear/self` that grabs a username from the cookies and clears the cache for the respective user.

### LINKED ISSUE #628: Create an endpoint for logs
- Create a GET endpoint for getting the logs of the user.
- It should give the logs of Cloudflare Cache Clear of the last 24 hours with log count

rohan09-raj:
Closes #292 
Closes #628 

- Created a GET API `/cache` that returns the latest meta data of Cloudflare cache purged in the last 24 hours of the logged-in user
- Created a POST API `/cache` that purges the cache of the specific page
- Added tests for the controller, model and util functions


utils/time.js
```diff
@@ -43,9 +43,22 @@ const getTimeInSecondAfter = ({ timestamp = Date.now(), days = 0, hours = 0, min
   return parseInt(timeInMilliseconds / 1000);
 };
 
+/**
+ * Returns time in seconds of timestamp before given duration
```
heyrandhir, line 47: are we returning time in `seconds` or `milliseconds` here

heyrandhir (approved):
LGTM

Maheima (dismissed):
LGTM

services/cloudflareService.js
```diff
@@ -0,0 +1,23 @@
+const { fetch } = require("../utils/fetch");
+
+const CLOUDFLARE_ZONE_ID = config.get("cloudflare.CLOUDFLARE_ZONE_ID");
+const CLOUDFLARE_PURGE_CACHE_API = `https://api.cloudflare.com/client/v4/zones/${CLOUDFLARE_ZONE_ID}/purge_cache`;
```
RitikJaiswal75, line 4: Cloudflare itself provides this, cant it be in the constants file? let's say v4 becomes v5 we will have to search and change it, keeping it in the constants file would make it easier.

services/cloudflareService.js
```diff
@@ -0,0 +1,23 @@
...
+
+async function purgeCache(files) {
+  const response = await fetch(
+    CLOUDFLARE_PURGE_CACHE_API,
+    "POST",
+    null,
+    { files: files },
```
RitikJaiswal75, line 11: isn't this for hosts/prefixes?

As per the documentation, there is a hosts parameter in the request
[Cloudflare API Documentation | Caching](https://developers.cloudflare.com/api/operations/zone-purge-files-by-cache-tags,-host,-or-prefix)

routes/cloudflareCache.js
```diff
@@ -0,0 +1,9 @@
+const express = require("express");
+const router = express.Router();
+const cloudflareCache = require("../controllers/cloudflareCache");
+const authenticate = require("../middlewares/authenticate");
+
+router.get("/", authenticate, cloudflareCache.fetchPurgedCacheMetadata);
+router.post("/", authenticate, cloudflareCache.purgeMembersCache);
```
RitikJaiswal75, line 7: when are we planning to purge the cache, is it automatic, or when someone hits the endpoint? If it is based on hitting the endpoint shouldn't it be done by a superuser as any normal user can easily overload the endpoint? As per the Cloudflare API, we only get 30,000 purge requests in 24 hours.

controllers/cloudflareCache.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+const purgeMembersCache = async (req, res) => {
+  try {
+    const { id, username } = req.userData;
+    const logs = await logsQuery.fetchCacheLogs(id);
+    const logsCount = logs.length;
+
+    const files = [`https://members.realdevsquad.com/${username}`];
+
+    if (logsCount < MAX_CACHE_PURGE_COUNT) {
```
RitikJaiswal75, line 21: What if a superuser is purging the cache?

services/cloudflareService.js
rohan09-raj: Yeah sure will make the change

services/cloudflareService.js
rohan09-raj: We are using our fetch function for this.
https://github.com/Real-Dev-Squad/website-backend/blob/develop/utils/fetch.js
image

routes/cloudflareCache.js
rohan09-raj: The whole idea behind this feature is, when someone is showing their page to some recruiter or someone else and if the page is cached of members site, then they can instantly purge the cache to see the updated page.
Also we have put the limit to user, in `24 hours` they are allowed to purge cache only 3 times.

Consider a scenario
- I have 3 requests available
- I have purged the cache 1 time now 2 more requests are remaining
- Now this 1 slot will be freed after 24 hours means from the current time after 24 hours I have again 3 remaining slots available to purge the cache.

controllers/cloudflareCache.js
rohan09-raj: Explained the reason above as users should not be reached to superuser everytime they need the updated members page of theirs.

utils/time.js
rohan09-raj: It's actually a firestore timestamp returned from the function and contains both seconds and milliseconds as well. Let me correct the description here

controllers/cloudflareCache.js
RitikJaiswal75: But the superuser must be able to purge the cache as many times as they want.

Discussed it today with ankush in morning

services/cloudflareService.js
RitikJaiswal75: Does it mean files serve as `options` here?

routes/cloudflareCache.js
RitikJaiswal75: Got it.

services/cloudflareService.js
rohan09-raj: `files` serves as `data` or you can say it's `options` only but a specified option here

controllers/cloudflareCache.js
rohan09-raj: So I think for superuser there should be a separate API to purge the cache and separate UI on dashboard site as `username` will be expected in the body to purge the cache of any user. Or I think we can modify this API to adapt the behaviour that will make this API a little bit complex probably

rohan09-raj: Implemented in the same controller function in my local and it will look like this
image

rohan09-raj: Not the best code as there a lot of repetition. I can simplify this by making a separate controller function which just clears the cache and put the restriction in a new controller function which will be eventually called by route

rohan09-raj: Simplified a bit
image

RitikJaiswal75: cant you identify the super user from the req.userData?
then in the same controller add a check if it's a normal user then use max count if the superuser then don't use max count.

rohan09-raj: That is what I have implemented actually, see the above screenshot

rohan09-raj: Added the changes

controllers/cloudflareCache.js
```diff
@@ -0,0 +1,92 @@
...
+const { MAX_CACHE_PURGE_COUNT } = require("../constants/cloudflareCache");
+const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
+
+/**
+ * Purges the Cache of Members Profile Page
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const purgeCacheByUserOrSuperUser = async (req, res) => {
```
RitikJaiswal75, line 13: can we rename it to just purgeCache, it it were a separate controller it could have been purgrCacheUser and purgeCacheSuperuser

controllers/cloudflareCache.js
```diff
@@ -0,0 +1,92 @@
...
+  try {
+    const { id, username, roles } = req.userData;
+    const logs = await logsQuery.fetchCacheLogs(id);
+    const logsCount = logs.length;
+
+    // Cache purged by a superuser without rate limits
+    if (roles.super_user) {
+      const { user } = req.body;
+      const files = [`https://members.realdevsquad.com/${user}`];
+      return purgeCache(res, id, files);
```
RitikJaiswal75, line 23: where can I find this purge cache function?

controllers/cloudflareCache.js
rohan09-raj: There's already a function present named as purgeCache

controllers/cloudflareCache.js
rohan09-raj: Just below this controller, there's another controller named as purgeCache

RitikJaiswal75:
As discussed if I enter a random username as a superuser the message I get is cache purged that should not happen. Reason:

Let's say I am the superuser and I am trying to clear randhir's cache his username is randhir but by mistake I type in as randir. In this case I will get the message as cache purged successfully but actually, it did not clear

rohan09-raj:
> As discussed if I enter a random username as a superuser the message I get is cache purged that should not happen. Reason:
> 
> Let's say I am the superuser and I am trying to clear randhir's cache his username is randhir but by mistake I type in as randir. In this case I will get the message as cache purged successfully but actually, it did not clear

[Screencast from 2023-01-05 09-10-48.webm](https://user-images.githubusercontent.com/78433013/210697224-1392f936-7f38-418a-a8ae-4b2c6e50f653.webm)

RitikJaiswal75 (changes requested):
I tried out the following steps:
1. Cleared cache for different users and self  3 times as a super user
2. tried clearing cache self (without passing user in body) 

now it showed max limit reached.
Can i say super user can not clear their cache more than 3 times?

RitikJaiswal75:
the Get for super user can have a user parameter to get the purged cache for specific user.

rohan09-raj:
> Can i say super user can not clear their cache more than 3 times?

Yes, that's the limitation if we use the same API

controllers/cloudflareCache.js
```diff
@@ -0,0 +1,102 @@
...
+    const logs = await logsQuery.fetchCacheLogs(id);
+    const logsCount = logs.length;
+
+    // Cache purged by a superuser without rate limits
+    if (req.body.user) {
+      if (roles.super_user) {
+        const { user } = req.body;
+        const userDetails = await userQuery.fetchUser({ username: user });
+        if (!userDetails.userExists) {
+          return res.boom.badRequest("Please provide a valid username");
```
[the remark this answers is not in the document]
rohan09-raj: Ok will update that

controllers/cloudflareCache.js
```diff
@@ -0,0 +1,102 @@
...
+          await logsQuery.addLog(
+            logType.CLOUDFLARE_CACHE_PURGED,
+            { userId: id, purgedFor: userDetails.user.id },
+            { message: "Cache Purged" }
+          );
+        }
+        return res.json({ message: "Cache purged successfully", purgedFor: userDetails.user.id, ...response.data });
+      } else {
+        return res.boom.unauthorized("You are not authorized to perform this action");
+      }
```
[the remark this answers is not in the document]
rohan09-raj: There are minor additional things for superuser which makes the abstraction a bit harder but will try to do that

rohan09-raj:
> the Get for super user can have a user parameter to get the purged cache history for specific users.

Here, the GET endpoint works a bit differently, it returns only the count and when it was last cleared within the time span of last 24 hours. Because these are only details needed on frontend to display.

Superuser can get the logs of it using the GET logs API as eventually the details for `cache` GET endpoints, comes from the logs itself

controllers/cloudflareCache.js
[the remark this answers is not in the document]
rohan09-raj: Updated

controllers/cloudflareCache.js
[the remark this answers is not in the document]
rohan09-raj: Done with the changes

middlewares/validators/userStatus.js
```diff
@@ -1,28 +1,41 @@
-const joi = require("joi");
+const Joi = require("joi");
 const { userState } = require("../../constants/userStatus");
+const threeDaysInMilliseconds = 172800000;
```
sohamsshah, line 3: since this is a constant, we can send it to `/constants` file. Also, it can be in capital letters spaced by `underscore` as general notation

models/userStatus.js
```diff
@@ -78,39 +79,136 @@ const getAllUserStatus = async (query) => {
...
+      const docRef = document.ref;
+      const newStatusData = { ...doc };
+      let toUpdate = false;
+      const { futureStatus, currentStatus } = doc;
+      const { state: futureState } = futureStatus;
+      if (futureState === "ACTIVE" || futureState === "IDLE") {
+        if (today >= futureStatus.from) {
+          newStatusData.currentStatus = { ...futureStatus, until: "", updatedAt: today };
+          newStatusData.futureStatus = {};
+          toUpdate = !toUpdate;
```
sohamsshah, line 179: because this is inside the`if block` it makes more sense to assign a boolean to `toUpdate` instead of toggling it. Applies to other places as well

models/userStatus.js
```diff
@@ -78,39 +79,136 @@ const getAllUserStatus = async (query) => {
...
+        }
+      }
+      if (toUpdate) {
+        batch.set(docRef, newStatusData);
+      }
+    });
+    if (batch._ops.length > 100) {
+      logger.info(
+        `Warning: More than 100 User Status documents to update. The max limit permissible is 500. Refer https://github.com/Real-Dev-Squad/website-backend/issues/890 for more details.`
```
sohamsshah, line 203: This can be made as a constant

sohamsshah (changes requested):
Overall code looks good. Tests are well added. Provided some comments, please address them

RitikJaiswal75:
Can you provide me with the links to API contract and the schema

test/integration/cloudflareCache.test.js
```diff
@@ -0,0 +1,200 @@
...
+        .get("/cache")
+        .set("cookie", `${cookieName}=${jwt}`)
+        .end((error, response) => {
+          if (error) {
+            return done(error);
+          }
+
+          expect(response).to.have.status(200);
+          expect(response.body).to.be.an("object");
+          expect(response.body.message).to.be.a("string");
```
RitikJaiswal75, line 51: where are we checking whether the correct message is passed sent or not the server can send anything of type string?

test/integration/cloudflareCache.test.js
```diff
@@ -0,0 +1,200 @@
...
+            return done(err);
+          }
+
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Cache purged successfully");
+          expect(res.body.success).to.equal(true);
+          expect(res.body.errors).to.deep.equal([]);
+          expect(res.body.messages).to.deep.equal([]);
+          expect(res.body.result).to.be.a("object");
```
RitikJaiswal75, line 131: why are we expecting empty arrays, i understand about errors to some extent but why do we need it in message ?

test/unit/models/logs.test.js
```diff
@@ -0,0 +1,40 @@
...
+  describe("Fetch purged cache logs", function () {
+    it("Should fetch the purged cache logs", async function () {
+      const { type, meta, body } = cacheData.cacheLogs[0];
+      const userId = meta.userId;
+
+      await logsQuery.addLog(type, meta, body);
+      const data = await logsQuery.fetchCacheLogs(userId);
+
+      expect(data[0].timestamp).to.be.an("object");
+      expect(data[0].timestamp._seconds).to.be.a("number");
```
RitikJaiswal75, line 22: nit: is it possible to enforce that number is between 0-30

test/integration/cloudflareCache.test.js
rohan09-raj: It's the response returned from the Cloudflare API itself

test/unit/models/logs.test.js
rohan09-raj: Couldn't get you, how can we enforce the seconds to be in the range of 0-30 and what's the benefit of doing it?

test/integration/cloudflareCache.test.js
rohan09-raj: Made the improvements

test/integration/cloudflareCache.test.js
rohan09-raj: https://api.cloudflare.com/#zone-purge-files-by-url

test/unit/models/logs.test.js
RitikJaiswal75: you are just checking that it's a number, it can be any number like 100 but seconds are only between 0-59 so.

The benefit would be better testing.

Also, it's an **NIT** comment.

test/integration/cloudflareCache.test.js
RitikJaiswal75: Can we keep the exact message? fix it in other places as well

Then we can get it merged bu today EOD

RitikJaiswal75 (approved):
LGTM

PR #900, opened 2023-01-11
Archive role check for queried Users State

heyrandhir:
- [x] Adds check for archive role before sending the list of queried users
- [x] Adds tests for the same

### DIFF
--- a/controllers/userStatus.js
+++ b/controllers/userStatus.js
@@ -70,17 +70,21 @@ const getUserStatus = async (req, res) => {
 const getAllUserStatus = async (req, res) => {
   try {
     const { allUserStatus } = await userStatusModel.getAllUserStatus(req.query);
+    const activeUsers = [];
     for (const status of allUserStatus) {
       //  fetching users from users collection by userID in userStatus collection
       const result = await fetchUser({ userId: status.userId });
-      status.full_name = `${result.user.first_name} ${result.user.last_name}`;
-      status.picture = result.user.picture;
-      status.username = result.user.username;
+      if (!result.user?.roles?.archived) {
+        status.full_name = `${result.user.first_name} ${result.user.last_name}`;
+        status.picture = result.user.picture;
+        status.username = result.user.username;
+        activeUsers.push(status);
+      }
     }
     return res.json({
       message: "All User Status found successfully.",
-      totalUserStatus: allUserStatus.length,
-      allUserStatus,
+      totalUserStatus: activeUsers.length,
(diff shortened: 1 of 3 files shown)


RitikJaiswal75 (commented):
Can we have a test for it to try to store a user with a role archive and call the API and check for returned value?

manish591 (commented):
LGTM

heyrandhir:
> Can we have a test for it to try to store a user with a role archive and call the API and check for returned value?

Yes done 🫡

controllers/userStatus.js
```diff
@@ -70,17 +70,21 @@ const getUserStatus = async (req, res) => {
...
+        status.username = result.user.username;
+        activeUsers.push(status);
+      }
     }
     return res.json({
       message: "All User Status found successfully.",
-      totalUserStatus: allUserStatus.length,
-      allUserStatus,
+      totalUserStatus: activeUsers.length,
+      allUserStatus: activeUsers,
```
Maheima, line 87: Good job handling it like this here. This way you saved the extra iterations of changing allUserStatus everywhere it is used

heyrandhir:
> LGTM! Good work Randhir 🎉

Thanks @Maheima 🙂

controllers/userStatus.js
heyrandhir: Yes changing it in backend was better than handling in frontend everywhere.

test/integration/userStatus.test.js
```diff
@@ -63,6 +63,44 @@ describe("UserStatus", function () {
...
+      const response = await chai.request(app).get("/users/status?state=IDLE");
+      expect(response).to.have.status(200);
+      expect(response.body.message).to.equal("All User Status found successfully.");
+      expect(response.body.totalUserStatus).to.be.a("number");
+      expect(response.body.totalUserStatus).to.equal(1);
+      expect(response.body.allUserStatus).to.be.a("array");
+      expect(response.body.allUserStatus.length).to.equal(1);
+    });
+
+    it("Should only get the idle userStatus only for active user when both archived and active users are there in DB", async function () {
```
RitikJaiswal75, line 91: can we change the statement to `should return only non-archived user`

the above test for `should not get archived user` can be removed as this test checks for both

RitikJaiswal75 (approved):
LGTM, 
Thankyou for always being open to changes and implementing them in a jiffy.

manish591 (approved):

shmbajaj (approved):

PR #902, opened 2023-01-12
added integration for new sign up flow in backend

vinayak-trivedi:
What does this PR do: 

- adds redirection to the new-signup page of my site in case the user has incomplete details

### DIFF
--- a/controllers/auth.js
+++ b/controllers/auth.js
@@ -12,7 +12,7 @@ const authService = require("../services/authService");
 const githubAuth = (req, res, next) => {
   let userData;
   const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
-  const authRedirectionUrl = req.query.state ?? rdsUiUrl;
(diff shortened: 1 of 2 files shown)


controllers/auth.js
```diff
@@ -12,7 +12,7 @@ const authService = require("../services/authService");
 const githubAuth = (req, res, next) => {
   let userData;
   const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
-  const authRedirectionUrl = req.query.state ?? rdsUiUrl;
+  let authRedirectionUrl = req.query.state ?? rdsUiUrl;
```
RitikJaiswal75, line 15: why did we change const to let over here?

vinayak-trivedi: we can change the redirection URL if we have incomplete user details, so we can redirect them to the new sign up flow

heyrandhir (approved):

RitikJaiswal75 (approved):
LGTM

Issue #905, opened 2023-01-15
RFC | Missing rate limiting


shmbajaj:
## What is Rate Limting?
Rate Limting is an strategy for limiting the rate at which requests are accepted by HTTP request handlers. 
It puts an cap on how often someone(human or bot) can repeat an action within a certain timeframe for instance, 
trying to log in to an account(sometimes with the intent of causing damage).

Without this, applications become vulnerable to attacks such as **DOS**. 
A rate-limiting middleware(in-app) should be used to prevent such attacks.

Broadly, rate limiting can be implemented using one of the two approaches:
- Using a Proxy
- Using In-App
<pre><small>In-App is suitable for our use case</small></pre>

<details>
  <summary>More on rate-limting</summary>
  
  1. `Rate` is numberOfRequests/timeframe
  2. `Limit` is constraint on whom rate-limting is applied. Which can be either global or on a   per-client basis(a client is user of a service, who is identified either by uid or ipaddress). 
  3. `Store` is a database which is used for storing info about which client requests to limit or how many requests are still acceptable. The store can be Redis, process Memory, and Hashmap.

</details>

## Source of Problem
- CodeQL code scanning reported missing rate-limiting middelware
![image](https://user-images.githubusercontent.com/29247011/212561584-815ab90c-0e69-4a75-9edb-9dc1e4e42653.png)

## How rate-limiting middelware is usable
- Limiting or slowing down unauthorized and/or unprevileged requests such as accessing resources limited to superuser or requires authorization for access.
- Tracking user activity(sometimes with the intent of causing damage).
- Reducing read/write costs for critical resources

## Methods for implement rate-limting middelware
1. Using third party packages
   - [Rate Limiter Flexible](https://github.com/animir/node-rate-limiter-flexible) <small>(can use either in-memory or redis store)</small>
   - [Express Rate Limit](https://www.npmjs.com/package/express-rate-limit)
2. Write a custom implementation<small>(for rate-limit algos, mentioned in order of least prefference)</small>
   - Fixed Window Counters
   - Sliding Window Counters
   - Sliding Window Log
   - Leaky Bucket
   - Token Bucket

## What needs to be done
- Write middelware for handling rate-limiting globally(across all routes) 
or on specific routes.

### Refferences
- [CodeQL doc on missing-rate-limiting](https://codeql.github.com/codeql-query-help/javascript/js-missing-rate-limiting/)<small>*</small>
[...]

PR #917, opened 2023-02-05
Rate Limiting Middelware By IP | HOLD | Requirements Not Collected and Cleared

shmbajaj:
Count and Limit requests by `IP-address`(key) with atomic increments, 
using process **in-memory store**.

### Before Change
- Requests were not rate-limited on critical or essential resource endpoints.
- Github CodeQL was complaining about [missing rate limiting](https://codeql.github.com/codeql-query-help/javascript/js-missing-rate-limiting/).
- Resources are prone to brute force attacks.
### After Change
- Requests are rate-limited on critical or essential resources endpoints, which limits
   - Fast brute force attack: 5 requests/30 seconds and on limit reach block for 10 minutes.
   - Slow brute force attack: 100 requests/1 day and on limit reach block for 1 day.
   - Key used for limit: IP-address
   - Storing info(store) : In Process Memory 

### Checks
- [X] Is TDD Followed
- [X] Is Commit Checks Resulted as Expected
- [X] Is Functions Description Added If Required or Critical

### Issue
- Closes [905](https://github.com/Real-Dev-Squad/website-backend/issues/905)

### Info
- `Package[used]:` [node-rate-limiter-flexible](https://github.com/animir/node-rate-limiter-flexible)
- `Store:` [Rate Limiter Memory](https://github.com/animir/node-rate-limiter-flexible/wiki/Memory) manages limits in current process memory.
      - Can block a key for max 24 days
      - Non-persistent storage, get's reset on application stop or failure.
      - Alternative: use persistent storage such as Redis.
- `Approach:` Block Source of requests by IP
      - Simple easy to implement
      - Easily scalable as per requirements and combinable with other limiters
- `TDD Coverage:`
Screenshot 2023-02-07 at 12 40 08 PM

### DIFF
--- a/constants/rateLimtingMiddelware.js
+++ b/constants/rateLimtingMiddelware.js
@@ -1,0 +1,8 @@
+export const MAX_WRONG_ATTEMPTS_BY_IP_PER_MINUTE = 5;
+export const MAX_WRONG_ATTEMPTS_BY_IP_PER_DAY = 100;
+export const MAX_REQUEST_WINDOW_BY_IP_PER_MINUTE = 30;
+export const MAX_REQUEST_WINDOW_BY_IP_PER_HOUR = 60 * 60 * 24;
+export const BLOCK_DURATION_WINDOW_BY_IP_PER_MINUTE = 60 * 10;
+export const BLOCK_DURATION_WINDOW_BY_IP_PER_HOUR = 60 * 60 * 24;
+export const MAX_WRONG_ATTEMPTS_BY_IP_PER_MINUTE_KEY_PREFIX = "login_fail_by_ip_per_minute";
+export const MAX_WRONG_ATTEMPTS_BY_IP_PER_DAY_KEY_PREFIX = "login_fail_by_ip_per_day";
--- a/constants/rateLimtingMiddelware.js
+++ b/constants/rateLimtingMiddelware.js
@@ -11,8 +11,10 @@ const RATE_LIMITER_FAST_BRUTE_BY_IP_OPTIONS = {
   duration: 30,
   blockDuration: 60 * 10,
 };
-
-const TOO_MANY_REQUESTS = "Too Many Requests";
+const TOO_MANY_REQUESTS = {
+  ERROR_TYPE: "Too Many Requests",
+  STATUS_CODE: 429,
+};
 
 module.exports = {
   RATE_LIMITER_FAST_BRUTE_BY_IP_OPTIONS,
--- a/constants/rateLimtingMiddelware.js
+++ b/constants/rateLimtingMiddelware.js
@@ -1,8 +1,21 @@
-export const MAX_WRONG_ATTEMPTS_BY_IP_PER_MINUTE = 5;
-export const MAX_WRONG_ATTEMPTS_BY_IP_PER_DAY = 100;
-export const MAX_REQUEST_WINDOW_BY_IP_PER_MINUTE = 30;
-export const MAX_REQUEST_WINDOW_BY_IP_PER_HOUR = 60 * 60 * 24;
-export const BLOCK_DURATION_WINDOW_BY_IP_PER_MINUTE = 60 * 10;
-export const BLOCK_DURATION_WINDOW_BY_IP_PER_HOUR = 60 * 60 * 24;
-export const MAX_WRONG_ATTEMPTS_BY_IP_PER_MINUTE_KEY_PREFIX = "login_fail_by_ip_per_minute";
-export const MAX_WRONG_ATTEMPTS_BY_IP_PER_DAY_KEY_PREFIX = "login_fail_by_ip_per_day";
+const RATE_LIMITER_SLOW_BRUTE_BY_IP_OPTIONS = {
+  points: 100,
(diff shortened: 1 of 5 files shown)


Maheima (approved):
LGTM!

constants/rateLimtingMiddelware.js
```diff
@@ -0,0 +1,23 @@
...
+  blockDuration: 60 * 60 * 24,
+  keyPrefix: "login_fail_by_ip_per_day",
+};
+
+const RATE_LIMITER_FAST_BRUTE_BY_IP_OPTIONS = {
+  keyPrefix: "login_fail_by_ip_per_minute",
+  points: 5,
+  duration: 30,
+  blockDuration: 60 * 10,
+};
```
heyrandhir, line 13: How did we arrive at these figures. Also should these numbers be same for user and super user. 🤔

constants/rateLimtingMiddelware.js
```diff
@@ -0,0 +1,23 @@
+const RATE_LIMITER_SLOW_BRUTE_BY_IP_OPTIONS = {
+  points: 100,
+  duration: 60 * 60 * 24,
+  blockDuration: 60 * 60 * 24,
+  keyPrefix: "login_fail_by_ip_per_day",
+};
```
heyrandhir, line 6: Instead of blocking the users for a day,  1/2 hour or 1 hour should be sufficient. But still check with @ankushdharkar once.

constants/rateLimtingMiddelware.js
shmbajaj: I just picked from the official wiki of this rate limiter package, for starting.
Thanks for pointing that out, I didn't think about this.

> I will add limits as per different roles.

constants/rateLimtingMiddelware.js
shmbajaj: I will check limit and block duration.

constants/rateLimtingMiddelware.js
```diff
@@ -0,0 +1,23 @@
+const RATE_LIMITER_SLOW_BRUTE_BY_IP_OPTIONS = {
+  points: 100,
+  duration: 60 * 60 * 24,
+  blockDuration: 60 * 60 * 24,
```
ivinayakg, line 4: add these values from environment variables.

constants/rateLimtingMiddelware.js
```diff
@@ -0,0 +1,23 @@
...
+  keyPrefix: "login_fail_by_ip_per_day",
+};
+
+const RATE_LIMITER_FAST_BRUTE_BY_IP_OPTIONS = {
+  keyPrefix: "login_fail_by_ip_per_minute",
+  points: 5,
+  duration: 30,
+  blockDuration: 60 * 10,
```
ivinayakg, line 12: add these values from environment variables.

ivinayakg (changes requested):
Amazing work 🥳🥳,
I just had a query, can't we do rate limiting without this package (I'm not sure about it.) and if this gives some features or something that sort do let me know.

shmbajaj:
> Amazing work 🥳🥳, I just had a query, can't we do rate limiting without this package (I'm not sure about it.) and if this gives some features or something that sort do let me know.

- We can do rate limiting without the package, I'll get back to you on this.
- Features: 
   - This packages manage store for us such as in-process-memory-stote and also provide other options such as redis and few others.
   - Gives options to combine different rate limiters and use as single interface for limiting requests.
   - Other like insurance for persistent storage like redis, MongoDB and mysql.

middlewares/rateLimiting.js
```diff
@@ -0,0 +1,67 @@
...
+      responseRateLimiterSlowBruteByIP.consumedPoints > RATE_LIMITER_SLOW_BRUTE_BY_IP_OPTIONS.points
+    ) {
+      retrySeconds = getRetrySeconds(responseRateLimiterSlowBruteByIP.msBeforeNext);
+    }
+    if (retrySeconds > 0) {
+      throw Error();
+    }
+    await Promise.all([rateLimiterFastBruteByIP.consume(ipAddress), rateLimiterSlowBruteByIP.consume(ipAddress)]);
+    return next();
+  } catch (error) {
```
prakashchoudhary07, line 52: Can we provide a way to pass the param for rate limiting, so that It can change depending on the routes? It can de dynamic 
What do you think?

prakashchoudhary07 (changes requested):

middlewares/rateLimiting.js
shmbajaj:
```js
// INFO: here params are options to middelware creating function

// some-random-route
const { getAuthRateLimitingMiddelware } = require('middleware-path');
const rateLimitingMiddelware = getRateAuthLimitingMiddelware(...params);
router.get("/", rateLimtingMiddelware, authorization, getSomething);

// another-random-route
const { getAuthRateLimitingMiddelware } = require('middleware-path');
const rateLimitingMiddelware = getAuthRateLimitingMiddelware(...params);
router.post("/", rateLimtingMiddelware, authorization, doSomething);
```

Issue #912, opened 2023-01-25
Discord Bot API


rohan09-raj:
## Issue
We need an API to store the data of discord users who runs `/verify` command to verify themselves so that when verification process redirects to any Real Dev Squad website, we need to get the Discord ID of the user to link it successfully through the given link.

## What to do?
- [ ] Create a POST endpoint which can verify the connection using JWT between Cloudflare Worker and Backend and can also store the data in the backend in the below format
```
token: String,
discordId: String,
generationTime: Number,
expiry: Number,
linkStatus: Boolean,
```
**PR that creates a POST endpoint :** #911 

- [ ] Create a GET endpoint to get the Discord ID from generated token.

**PR that creates a GET endpoint :** TO BE CREATED

PR #926, opened 2023-02-20
Created GET API for External Accounts

rohan09-raj:
## Issue
#912 

## What's the change
- Created a GET endpoint to get the data of the external account
- **GET** `/external-accounts/:token`
- Data returned from the API
- Added a duplicate token check in `POST` `/external-accounts` API
```json
{
  "message": "Data returned successfully",
  "attributes": "<object>"
}
```

### DIFF
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -1,10 +1,10 @@
 const externalAccountsModel = require("../models/external-accounts");
 
 const addExternalAccountData = async (req, res) => {
-  try {
-    const createdOn = Date.now();
-    const data = { ...req.body, createdOn };
+  const createdOn = Date.now();
 
+  try {
+    const data = { ...req.body, createdOn };
     await externalAccountsModel.addExternalAccountData(data);
 
     return res.status(201).json({ message: "Added external account data successfully" });
@@ -14,4 +14,24 @@ const addExternalAccountData = async (req, res) => {
   }
 };
 
-module.exports = { addExternalAccountData };
+const getExternalAccountData = async (req, res) => {
+  try {
+    const externalAccountData = await externalAccountsModel.fetchExternalAccountData(req.query, req.params.token);
+    if (externalAccountData.length === 0) {
+      return res.status(404).json({ message: "No data found" });
+    }
+
+    const attributes = externalAccountData[0].attributes;
+
+    if (attributes.expiry && attributes.expiry < Date.now()) {
+      return res.status(498).json({ message: "Token Expired. Please generate it again" });
+    }
+
+    return res.status(200).json({ message: "Data returned successfully", attributes: attributes });
+  } catch (error) {
+    logger.error(`Error getting external account data: ${error}`);
+    return res.boom.serverUnavailable("Something went wrong please contact admin");
+  }
+};
+
+module.exports = { addExternalAccountData, getExternalAccountData };
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -17,13 +17,13 @@ const addExternalAccountData = async (req, res) => {
 const getExternalAccountData = async (req, res) => {
   try {
     const externalAccountData = await externalAccountsModel.fetchExternalAccountData(req.query, req.params.token);
-    if (externalAccountData.length === 0) {
-      return res.status(404).json({ message: "No data found" });
+    if (!externalAccountData.id) {
+      return res.boom.notFound("No data found");
     }
 
-    const attributes = externalAccountData[0].attributes;
+    const attributes = externalAccountData.attributes;
     if (attributes.expiry && attributes.expiry < Date.now()) {
-      return res.status(498).json({ message: "Token Expired. Please generate it again" });
+      return res.boom.unauthorized("Token Expired. Please generate it again");
     }
 
     return res.status(200).json({ message: "Data returned successfully", attributes: attributes });
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -22,7 +22,6 @@ const getExternalAccountData = async (req, res) => {
     }
 
     const attributes = externalAccountData[0].attributes;
-
     if (attributes.expiry && attributes.expiry < Date.now()) {
       return res.status(498).json({ message: "Token Expired. Please generate it again" });
     }
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -5,6 +5,13 @@ const addExternalAccountData = async (req, res) => {
 
   try {
     const data = { ...req.body, createdOn };
+
+    // Check if token already exists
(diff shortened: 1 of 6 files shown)


models/external-accounts.js
```diff
@@ -11,4 +11,33 @@ const addExternalAccountData = async (data) => {
...
+    const userExternalAccountData = [];
+
+    let externalAccountData;
+    externalAccountData = await externalAccountsModel.where("token", "==", param);
+
+    Object.keys(query).forEach((key) => {
+      if (key === "type") {
+        externalAccountData = externalAccountData.where(key, "==", query?.key);
+      }
+    });
```
TanishqSingla, line 25: The `forEach` only runs for `key`  `type`, why do we need a loop here?

models/external-accounts.js
```diff
@@ -11,4 +11,33 @@ const addExternalAccountData = async (data) => {
...
+    externalAccountData = await externalAccountData.limit(1).get();
+
+    externalAccountData.forEach((data) => {
+      userExternalAccountData.push({
+        id: data.id,
+        ...data.data(),
+      });
+    });
+
+    return userExternalAccountData;
```
TanishqSingla, line 36: `externalAccountData` will always contain one entry, why `forEach` here?

models/external-accounts.js
```diff
@@ -11,4 +11,33 @@ const addExternalAccountData = async (data) => {
...
+
+    Object.keys(query).forEach((key) => {
+      if (key === "type") {
+        externalAccountData = externalAccountData.where(key, "==", query?.key);
+      }
+    });
+
+    externalAccountData = await externalAccountData.limit(1).get();
+
+    externalAccountData.forEach((data) => {
```
shmbajaj, line 29:
```js
const querySnapshot = await externalAccountData.limit(1).get();
if (!querySnapshot.empty) {
  // do something with the first matching document
} else {
  // handle case where there are no matching documents
}

```

models/external-accounts.js
```diff
@@ -11,4 +11,33 @@ const addExternalAccountData = async (data) => {
...
 
-module.exports = { addExternalAccountData };
+const fetchExternalAccountData = async (query, param) => {
+  try {
+    const userExternalAccountData = [];
+
+    let externalAccountData;
+    externalAccountData = await externalAccountsModel.where("token", "==", param);
+
+    Object.keys(query).forEach((key) => {
```
shmbajaj, line 21: It can be done without forEach this way
```js
if (query && query?.['type']) {
  externalAccountData = externalAccountData.where("type", "==", query?.['type']);
}
```

shmbajaj (changes requested):

controllers/external-accounts.js
```diff
@@ -14,4 +14,24 @@ const addExternalAccountData = async (req, res) => {
...
+      return res.status(404).json({ message: "No data found" });
+    }
+
+    const attributes = externalAccountData[0].attributes;
+
+    if (attributes.expiry && attributes.expiry < Date.now()) {
+      return res.status(498).json({ message: "Token Expired. Please generate it again" });
+    }
+
+    return res.status(200).json({ message: "Data returned successfully", attributes: attributes });
```
samyakshah3008, line 30: Can you please explain more on reason for using two if blocks? Can that be done with if and else if block?

samyakshah3008 (approved):
Looks good for approval ✅🚀

models/external-accounts.js
rohan09-raj: It was there because we can have more queries in future and then the if check can be removed. But yeah we can take it of it in future and I have removed the loop for now

models/external-accounts.js
rohan09-raj: Done with change

models/external-accounts.js
rohan09-raj: Thanks for the snippet, made that change

models/external-accounts.js
rohan09-raj: Done with change.

controllers/external-accounts.js
rohan09-raj: Two if block are I think more readable instead of a if else block as those are the edge cases being handled

samyakshah3008 (dismissed):
LGTM!

controllers/external-accounts.js
```diff
@@ -14,4 +14,23 @@ const addExternalAccountData = async (req, res) => {
...
+    if (externalAccountData.length === 0) {
+      return res.status(404).json({ message: "No data found" });
+    }
+
+    const attributes = externalAccountData[0].attributes;
+    if (attributes.expiry && attributes.expiry < Date.now()) {
+      return res.status(498).json({ message: "Token Expired. Please generate it again" });
+    }
+
+    return res.status(200).json({ message: "Data returned successfully", attributes: attributes });
```
shmbajaj, line 29: Hey it will be good to use either `status` or `boom`.

shmbajaj, line 29: Hey it will be good to use either `status` or `boom`.

controllers/external-accounts.js
```diff
@@ -14,4 +14,23 @@ const addExternalAccountData = async (req, res) => {
...
 
-module.exports = { addExternalAccountData };
+const getExternalAccountData = async (req, res) => {
+  try {
+    const externalAccountData = await externalAccountsModel.fetchExternalAccountData(req.query, req.params.token);
+    if (externalAccountData.length === 0) {
+      return res.status(404).json({ message: "No data found" });
+    }
+
+    const attributes = externalAccountData[0].attributes;
```
shmbajaj, line 24: please add a comment here, why array index `0` item is used instead of the whole array.

controllers/external-accounts.js
rohan09-raj: The `status` is already there

rohan09-raj: Updated the `returns` and used `boom` there

shmbajaj (approved):

controllers/external-accounts.js
```diff
@@ -1,10 +1,17 @@
...
 
 const addExternalAccountData = async (req, res) => {
+  const createdOn = Date.now();
+
   try {
-    const createdOn = Date.now();
     const data = { ...req.body, createdOn };
 
+    // Check if token already exists
+    const dataFound = await externalAccountsModel.fetchExternalAccountData("", data.token);
```
shmbajaj, line 10: Hey @rohan09-raj if possible can you give a brief on what's happening from line 10-12

rohan09-raj: We are generating a random token using `crpyto.subtle`, there are cases which people pointed that after certain time they experienced duplicate tokens. It's very very rare and the developers of that package resolved the issue too but just in case if we get a duplicate token generated. We should not be putting that request data in the database and throw `token already exists` message.

PR #931, opened 2023-02-27
feat: Add task request flow

### LINKED ISSUE #904: [RFC] Task Levelling new flow
### Current flow

- currently, the user marks their task as complete from status and they get options to whether they want a new task or not
- if the user clicks on don't assign task button then they are not assigned a new task and their task is marked as complete
- if they click on continue then the `tasks/assign/self` API is called which checks if the user has a status idle and then assigns them a task based on their skill levels

### How is it decided which task should be assigned to the user?
- we take the skills of  the user and then give them a task up to +2 level of that user's skill
- if the user doesn't have a skill then we get any task that is less than level 2 and assign them that task
- once the task is assigned we update user status to active

### What happens if there is no task available based on user skills or no task available which is less than level 2?

- The user has an option of fetching a task in the future for themselves by using a fetch button that is provided to them on my site only if they don't have an ongoing task, otherwise, the button will not be visible to them
- The super also has options of assigning tasks to users from the dashboard site, they also get suggested users based on the category of task that they have selected there

### current Flow diagram

Untitled Diagram drawio (2)

TanishqSingla:
related to closes #904
## Task Request Flow
This merge introduces api for task request feature. 
The features include
- API to fetch task requests to SUPERUSER
- API to approve task request by SUPERUSER
- API to request for a task for user.


TanishqSingla:
Waiting on: [Task Request API Contract](https://github.com/Real-Dev-Squad/website-api-contracts/pull/110)

controllers/tasks.js
```diff
@@ -258,12 +258,11 @@ const assignTask = async (req, res) => {
...
-    const { taskId } = await tasks.updateTask({ assignee: username, status: TASK_STATUS.ASSIGNED }, task.itemId);
-    if (taskId) {
-      // this will change once we start storing status in different collection
-      await addOrUpdate({ status: "active" }, userId);
-    }
-    return res.json({ message: "Task assigned", Id: task.itemId });
+    // before:
+    // const { taskId } = await tasks.updateTask({ assignee: username, status: TASK_STATUS.ASSIGNED }, task.itemId);
+    // after:
+    const taskRequest = await taskRequestModel({ taskId: task.itemId, userId });
```
vinayak-trivedi, line 264: shouldn't this be taskRequestModel.createRequest?

vinayak-trivedi:
> related to #904

try using closes #904, this will automatically close the issue ones this PR is merged

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
...
+    if (taskRequests && taskRequests.length > 0) {
+      return res.status(200).json({
+        message: "Task requests returned successfully",
+        taskRequests,
+      });
+    }
+
+    return res.status(400).json({
+      message: "Unable to fetch task requests",
+    });
```
vinayak-trivedi, line 16: should this be 400? if the taskRequest doesn't come for any other reasons? it iss not going to be only bad request I guess

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
...
+
+    if (taskRequestResponse.message.includes("updated")) {
+      return res.status(200).json(taskRequestResponse);
+    }
+
+    if (taskRequestResponse.message.includes("created")) {
+      return res.status(201).json(taskRequestResponse);
+    }
+
+    return res.status(400).json({ message: taskRequestResponse.message });
```
vinayak-trivedi, line 37: same as above, it can be some other reasons that we do not get the response

models/taskRequests.js
```diff
@@ -0,0 +1,127 @@
...
+};
+
+/**
+ * Creates a task request
+ *
+ * @param taskId { string }: id of task request
+ * @param userId { string }: userId of user whose creating the task request
+ * @return {Promise<{taskRequest: Object}>}
+ */
+const createTaskRequest = async (taskId, userId) => {
```
vinayak-trivedi, line 38: can we rename it to createOrUpdateRequest

models/taskRequests.js
```diff
@@ -0,0 +1,127 @@
...
+      return {
+        error: "User does not exists",
+      };
+    }
+
+    if (user.status === USER_STATUS.OOO || user.status === USER_STATUS.ACTIVE) {
+      return {
+        error: "User is unavailable",
+      };
+    }
```
vinayak-trivedi, line 105: we are not storing the user status on the user object now, connect with @heyrandhir regarding this

utils/taskRequests.js
```diff
@@ -0,0 +1,20 @@
...
+ * Validates task request object to create firestore document
+ * @params taskRequest { Object } task request object
+ * @returns taskRequestDocument { Object } task request object valid for firestore
+ */
+const toFirestoreData = (taskRequest) => {
+  const taskRequestDocument = {};
+
+  for (const key in taskRequest) {
+    if (taskRequest[key]) {
+      taskRequestDocument[key] = taskRequest[key];
```
vinayak-trivedi, line 11: generic object inection sink warning here, check if you can fix this

vinayak-trivedi (changes requested):
Thanks for the PR @TanishqSingla, the code is really clean which made it easy to review it, thanks for that as well, suggesting some changes please look at them

test/integration/taskRequests.test.js
```diff
@@ -0,0 +1,286 @@
...
+    await cleanDb();
+  });
+
+  afterEach(async function () {
+    sinon.restore();
+  });
+
+  describe("GET / - gets tasks requests", function () {
+    describe("When the user is super user", function () {
+      before(async function () {
```
shmbajaj, line 40: I believe here we don't require to call these functions like `verifyToken`, `generateAuthToken`, and `users-model`.

You can mock these functions, and get work done. 

In testing I believe we shouldn't be doing actual operations/db-transactions.

test/integration/taskRequests.test.js
```diff
@@ -0,0 +1,286 @@
...
+            expect(res.body.message).to.equal("Task requests returned successfully");
+            expect(res.body.taskRequests).to.be.a("Array");
+            expect(res.body.taskRequests.length).to.equal(1);
+            return done();
+          });
+      });
+    });
+
+    describe("When the user is not a super user", function () {
+      before(async function () {
```
shmbajaj, line 69: Same, we can do mocking here as well instead of employing/calling acutal function code.

test/integration/taskRequests.test.js
shmbajaj: For example mocking `authService`
```js
sinon.stub(authService, "generateAuthToken").returns("valid_token");
sinon.stub(authService, "verifyAuthToken").returns({ userId: superUserId });
```

models/taskRequests.js
shmbajaj: +1

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
...
+    const taskRequests = await taskRequestsModel.fetchTaskRequests();
+
+    if (taskRequests && taskRequests.length > 0) {
+      return res.status(200).json({
+        message: "Task requests returned successfully",
+        taskRequests,
+      });
+    }
+
+    return res.status(400).json({
```
shmbajaj, line 14: Why `400` if takskRequests is empty, doesn't it means no task-requests is been created?

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
+const taskRequestsModel = require("../models/taskRequests");
+
+const fetchTaskRequests = async (_, res) => {
+  try {
+    const taskRequests = await taskRequestsModel.fetchTaskRequests();
+
+    if (taskRequests && taskRequests.length > 0) {
```
shmbajaj, line 7: You won't need to add this check, if the response array is empty then it's still correct.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
...
+        taskRequests,
+      });
+    }
+
+    return res.status(400).json({
+      message: "Unable to fetch task requests",
+    });
+  } catch (err) {
+    logger.error("Error while fetching task requests", err);
+    throw err;
```
shmbajaj, line 19: Where your handling/catching this error?

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    if (taskRequestResponse.message.includes("created")) {
+      return res.status(201).json(taskRequestResponse);
+    }
+
+    return res.status(400).json({ message: taskRequestResponse.message });
+  } catch (err) {
+    logger.error("Error while creating task request");
+    throw err;
```
shmbajaj, line 40: Where your handling/catching this error?

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
...
+
+const approveTaskRequest = async (req, res) => {
+  try {
+    const { taskRequestId, userId } = req.body;
+    const response = await taskRequestsModel.approveTaskRequest(taskRequestId, userId);
+
+    return res.status(200).json(response);
+  } catch (err) {
+    logger.error("Error while approving task request", err);
+    throw err;
```
shmbajaj, line 52: Same as above comment.

controllers/tasksRequests.js
shmbajaj, line 37: If it's a failed request you also can throw an error and let the catch part handle it.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const createTaskRequest = async (req, res) => {
+  try {
+    const { taskId, userId } = req.body;
+
+    const taskRequestResponse = await taskRequestsModel.createTaskRequest(taskId, userId);
+
+    if (taskRequestResponse.message.includes("updated")) {
+      return res.status(200).json(taskRequestResponse);
```
shmbajaj, line 30: You dont need two if statements here.
```js
const isUpdateRequest = taskRequestResponse.message.includes("updated");
return res.status(isUpdateRequest ? 200 : 201).json()
```

shmbajaj (changes requested):
- Mock functions/model-functions instead of actually calling those functions and make changes in the db/firebase while testing.
------------

**PR Review: In__Progress**

models/taskRequests.js
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * @return {Object}
+ */
+const fetchTaskRequests = async () => {
+  try {
+    const taskRequests = await taskRequestsCollection.get();
+
+    const taskRequestsData = [];
+
+    taskRequests.forEach((taskRequest) => taskRequestsData.push({ id: taskRequest.id, ...taskRequest.data() }));
```
shmbajaj, line 21: `Alternative:`

```js
return taskRequests.docs.map(doc => ({id: doc.id, ...doc.data()}) )
```

models/taskRequests.js
```diff
@@ -0,0 +1,142 @@
...
+    const { userExists, user } = await userModel.fetchUser({ userId });
+
+    if (!userExists) {
+      return { userDoesNotExists: true };
+    }
+
+    if (taskRequestData) {
+      const currentRequestors = taskRequestData.requestors;
+      if (currentRequestors.length >= 1) {
+        const requestorExists = !!currentRequestors.find((requestor) => requestor === user.id);
```
shmbajaj, line 50: Rename from `requestorExists` to `requestor`

models/taskRequests.js
```diff
@@ -0,0 +1,142 @@
...
+      isNoteworthy: taskData.isNoteworthy,
+      priority: taskData.priority,
+      purpose: taskData.purpose || "",
+      requestors: [userId],
+      status: TASK_REQUEST_STATUS.WAITING,
+      title: taskData.title,
+      type: taskData.type,
+    };
+
+    await taskRequestsCollection.doc(taskId).set(newTaskRequest);
```
shmbajaj, line 80: How your creating/getting `taskId`?
Is `taskId` unique enough?

shmbajaj: For creating a new document preferred method is `add`, not use `set` because of possible conflicts of `id` for a document.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,95 @@
+const { INTERNAL_SERVER_ERROR, SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
+const taskRequestsModel = require("../models/taskRequests");
+
+const fetchTaskRequests = async (_, res) => {
+  try {
+    const taskRequests = await taskRequestsModel.fetchTaskRequests();
+
+    return res.status(200).json({
```
shmbajaj, line 8: Either use status or boom here to maintain consistency.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,95 @@
...
+      message: "Task requests returned successfully",
+      taskRequests,
+    });
+  } catch (err) {
+    logger.error("Error while fetching task requests", err);
+    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
```
shmbajaj, line 14: What's the value of catching the err, if error message is generic here.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    if (response.isUpdate) {
+      return res.status(200).json({
+        message: "Task request successfully updated",
+        requestors: response.requestors,
+      });
+    }
+
+    return res.status(201).json({
```
shmbajaj, line 47: Either use status or boom here to maintain consistency.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    if (response.isUserOOO) {
+      return res.boom.conflict("User is currently OOO");
+    }
+    if (response.isUserActive) {
+      return res.boom.conflict("User is currently active on another task");
+    }
+
+    return res.status(200).json({
```
shmbajaj, line 81: Either use status or boom here to maintain consistency.

shmbajaj (changes requested):

controllers/tasksRequests.js
TanishqSingla: `boom` is used for throwing error object, on a success response I am using `res.status`

controllers/tasksRequests.js
TanishqSingla: same as above

controllers/tasksRequests.js
TanishqSingla: same as above

controllers/tasksRequests.js
TanishqSingla: the catch here is for `exception` cases, rest of the errors are handled by returning an object from the models.

models/taskRequests.js
TanishqSingla: The `taskId` is the id of the task for which the user want to create task request for.
In the current setup `taskId` and the id of `taskRequest` document is going to be same, and `set` is only used for cases in which a new `taskRequest` object needs to be created.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,95 @@
...
+    const response = await taskRequestsModel.addOrUpdate(taskId, userId);
+    if (response.userDoesNotExists) {
+      return res.boom.conflict("User does not exist");
+    }
+    if (response.requestorExists) {
+      return res.boom.conflict("User is already requesting for the task");
+    }
+    if (response.taskRequestDoesNotExist) {
+      return res.boom.conflict("Task request already exist");
+    }
```
vinayak-trivedi: can you change this name please?

controllers/tasksRequests.js
```diff
@@ -0,0 +1,95 @@
...
+
+    const response = await taskRequestsModel.approveTaskRequest(taskRequestId, userId);
+
+    if (response.userDoesNotExists) {
+      return res.boom.conflict("User does not exists");
+    }
+    if (response.userStatusDoesNotExists) {
+      return res.boom.conflict("User status does not exists");
+    }
```
vinayak-trivedi, line 72: do we need to check this here, if the user doesn't have status we are not allowing them to create task request

models/taskRequests.js
vinayak-trivedi, line 21: can we also send user data here, also we can paginate this API as there are going to be a lot of task requests

vinayak-trivedi (changes requested):
requesting some changes please check

models/taskRequests.js
TanishqSingla: This is a boolean value since `!!` operator is being provided, so I thought `requestorExists` is more appropriate name than `requestor`.

shmbajaj: Oh, I missed this part.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,95 @@
...
+      return res.boom.badRequest("userId not provided");
+    }
+
+    const response = await taskRequestsModel.addOrUpdate(taskId, userId);
+    if (response.userDoesNotExists) {
+      return res.boom.conflict("User does not exist");
+    }
+    if (response.requestorExists) {
+      return res.boom.conflict("User is already requesting for the task");
+    }
```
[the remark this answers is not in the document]
TanishqSingla: Implemented for availability of user status on user

controllers/tasksRequests.js
```diff
@@ -0,0 +1,98 @@
...
+
+    if (!taskId) {
+      return res.boom.badRequest("taskId not provided");
+    }
+    if (!userId) {
+      return res.boom.badRequest("userId not provided");
+    }
+
+    const response = await taskRequestsModel.addOrUpdate(taskId, userId);
+    if (response.userDoesNotExist) {
```
shmbajaj, line 30: `Suggestion:` use switch statement here.

models/taskRequests.js
```diff
@@ -0,0 +1,171 @@
...
+      return { userDoesNotExists: true };
+    }
+    if (!userStatusExists) {
+      return { userStatusDoesNotExist: true };
+    }
+
+    if (taskRequestData) {
+      const currentRequestors = taskRequestData.requestors;
+      if (currentRequestors.length >= 1) {
+        const requestorExists = !!currentRequestors.find((requestor) => requestor === user.id);
```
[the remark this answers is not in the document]
TanishqSingla: Thank you for the suggestion

controllers/tasksRequests.js
TanishqSingla: Switch statement won't work here since the object being returned is structured like this `{ userDoesNotExist: true }`, hence switch won't be much convinent here.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,101 @@
...
+      message: "Task requests returned successfully",
+      data,
+    });
+  } catch (err) {
+    logger.error("Error while fetching task requests", err);
+    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
+  }
+};
+
+const addOrUpdate = async (req, res) => {
```
vinayak-trivedi, line 18: can we name it `addOrUpdateTaskRequest`? that seems more suggestive of what this is doing

controllers/tasksRequests.js
```diff
@@ -0,0 +1,101 @@
...
+      return res.boom.badRequest("userId not provided");
+    }
+
+    const response = await taskRequestsModel.addOrUpdate(taskId, userId);
+    if (response.userDoesNotExist) {
+      return res.boom.conflict("User does not exist");
+    }
+    if (response.userStatusDoesNotExists) {
+      return res.boom.conflict("User status does not exists");
+    }
```
vinayak-trivedi, line 35: Why are we only checking for `userStatus` not existing, user status might exist but that user may be with an active task, then also we shouldn't allow them to create a task request right?

controllers/tasksRequests.js
```diff
@@ -0,0 +1,101 @@
...
+    if (response.requestorExists) {
```
vinayak-trivedi, line 36: we discussed about changing it right? can we change this to `alreadyRequested`, `requestorExists` suggests me that another requestor is already exiting for this

models/taskRequests.js
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Creates a task request
+ *
+ * @param taskId { string }: id of task request
+ * @return {Promise<{taskRequest: Object}>}
+ */
+const addOrUpdate = async (taskId, userId) => {
+  try {
+    const taskRequestsSnapshot = await taskRequestsCollection.where("taskId", "==", taskId).limit(1).get();
```
vinayak-trivedi, line 63: do we need to use `limit(1)`, are we going to have multiple taskRequest object for one task? are we not creating a task request object once, and updating that if more people request that task?

controllers/tasksRequests.js
shmbajaj: Hmm, Correct!! 😵‍💫

vinayak-trivedi (commented):
Great work @TanishqSingla, had few question and NIT  comments please check

models/taskRequests.js
```diff
@@ -76,13 +77,13 @@ const addOrUpdate = async (taskId, userId) => {
     if (taskRequestData) {
       const currentRequestors = taskRequestData.requestors;
       if (currentRequestors.length >= 1) {
```
shmbajaj, line 79: Take the condition on line `79` out. It will help in improving readability and from 2-levels of if nesting.

shmbajaj (changes requested):

shmbajaj (dismissed):
LGTM ✅

models/taskRequests.js
TanishqSingla: Removed `limit(1)` in the latest  commit

controllers/tasksRequests.js
vinayak-trivedi: makes sense

models/taskRequests.js
```diff
@@ -0,0 +1,176 @@
...
+    const { taskData } = await tasksModel.fetchTask(taskId);
+    if (!taskData) {
+      return { taskDoesNotExist: true };
+    }
+
+    const newTaskRequest = {
+      isNoteworthy: taskData.isNoteworthy,
+      priority: taskData.priority,
+      purpose: taskData.purpose || "",
+      requestors: [userId],
```
ivinayakg, line 107: what if we want to get all the taksRequest for a particular user?
It will have to get all the taskRequest and then loop through the requestors array and then find these.

TanishqSingla: We'll be using query operator `arrayContains`

models/taskRequests.js
```diff
@@ -0,0 +1,119 @@
...
+
+  const taskRequestsSnapshot = (await taskRequestsCollection.get()).docs;
+
+  for (const taskRequestSnapshot of taskRequestsSnapshot) {
+    const taskRequest = taskRequestSnapshot.data();
+    const { taskData } = await tasksModel.fetchTask(taskRequest.taskId);
+    const users = await Promise.all(
+      taskRequest.requestors.map((requestor) => {
+        if (visitedUsers.includes(requestor)) {
+          return visitedUsersId.find((visitedUserId) => (visitedUserId.id = requestor));
```
iamitprakash, line 27: The bug is that the visitedUsersId variable is not being initialized. This means that the find() method will always return undefined. To fix this bug, you need to initialize the visitedUsersId variable to an empty array. You can do this by adding the following line to the top of the code:

models/taskRequests.js
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Creates a task request
+ *
+ * @param taskId { string }: id of task request
+ * @return {Promise<{taskRequest: Object}>}
+ */
+const addOrUpdate = async (taskId, userId) => {
+  try {
+    const taskRequestsSnapshot = await taskRequestsCollection.where("taskId", "==", taskId).get();
+    const [taskRequestRef] = taskRequestsSnapshot.docs;
```
iamitprakash, line 48: The bug is that the docs property of the taskRequestsSnapshot object is an array of documents, but the [0] index is being used to access a single document. This will cause an error if there are no documents in the array. To fix this bug, you need to use the find() method to find the document with the specified taskId. You can do this by adding the following line to the code:

iamitprakash (changes requested):

models/taskRequests.js
TanishqSingla: visitedUsersId is already defined as a `Set`, thanks for pointing this out though as I need to use `.has` method instead of `.includes`.

models/taskRequests.js
TanishqSingla: I get your point, but since this is javascript I would instead get a `undefined` as value instead of code breaking on this, and I am using this `undefined` result to check if the task request data exists.

controllers/tasksRequests.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    const { userExists } = await userModel.fetchUser({ userId });
+    if (!userExists) {
+      return res.boom.conflict("User does not exist");
+    }
+
+    const { userStatusExists, data: userStatus } = await userStatusModel.getUserStatus(userId);
+    if (!userStatusExists) {
+      return res.boom.conflict("User status does not exist");
```
DashDeipayan, line 39: instead of doing 3 such rigorous if checks, how about we create a separate validator function?

controllers/tasksRequests.js
```diff
@@ -0,0 +1,118 @@
...
+      return res.boom.badRequest("userId not provided");
+    }
+
+    const { userExists, user } = await userModel.fetchUser({ userId });
+    if (!userExists) {
+      return res.boom.conflict("User does not exists");
+    }
+
+    const { userStatusExists, data: userStatus } = await userStatusModel.getUserStatus(userId);
+    if (!userStatusExists) {
```
DashDeipayan, line 92: Same validator function suggestion as above

models/taskRequests.js
```diff
@@ -0,0 +1,126 @@
...
```
DashDeipayan, line 55: We must do an `isEmpty` check is case there is no document, or a `find()`. Although this code works due `undefined`( thanks to JavaScript (;  )

DashDeipayan (changes requested):
Good job with the PR. Lets get it done

controllers/tasksRequests.js
```diff
@@ -0,0 +1,118 @@
...
+      return res.boom.conflict("User status does not exists");
+    }
+    if (userStatus.currentStatus.state === userState.OOO) {
```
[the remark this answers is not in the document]
TanishqSingla: This constant has been there for a while, I believe its best to deal with this issue separately.

models/taskRequests.js
```diff
@@ -0,0 +1,110 @@
...
+const fetchTaskRequests = async () => {
+  const taskRequests = [];
+
+  const taskRequestsSnapshots = (await taskRequestsCollection.get()).docs;
+
+  for (const taskRequestsSnapshot of taskRequestsSnapshots) {
+    const taskRequestData = taskRequestsSnapshot.data();
+    const { requestors } = taskRequestData;
+
+    const { taskData } = await tasksModel.fetchTask(taskRequestData.taskId);
```
DashDeipayan, line 22: You can move this line out of the for-loop by creating an array of promises(not writing await statements inside a for loop) and outside the for-loop this will be resolved with a `promise.all()`

models/taskRequests.js
```diff
@@ -0,0 +1,110 @@
...
+    const users = await Promise.all(requestors.map((requestor) => userModel.fetchUser({ userId: requestor })));
```
DashDeipayan, line 23: Same suggestion as above for the users

DashDeipayan (changes requested):
There are few comments that still needs to be addressed(not every comment is a change request, some are just queries 🙂)

DashDeipayan (approved):
LGTM

middlewares/taskRequests.js
```diff
@@ -0,0 +1,46 @@
...
+    const { userExists, user } = await userModel.fetchUser({ userId });
+    if (!userExists) {
+      return res.boom.conflict("User does not exist");
+    }
+
+    const { userStatusExists, data: userStatus } = await userStatusModel.getUserStatus(userId);
+    if (!userStatusExists) {
+      return res.boom.conflict("User status does not exist");
+    }
+    if (userStatus.currentStatus.state === userState.OOO) {
```
iamitprakash, line 28: will return a 409 Conflict error if the user's current status is either OOO or ACTIVE. However, it is possible for a user to be in a state where they are neither OOO nor ACTIVE. For example, a user could be in a state where they are available but have not yet started working on a task.

models/taskRequests.js
```diff
@@ -0,0 +1,122 @@
...
+};
+
+/**
+ * Approves task request to user
+ *
+ * @param taskRequestId { string }: id of task request
+ * @param userId { Object }: user whose being approved
+ * @return {Promise<{approvedTo: string, taskRequest: Object}>}
+ */
+const approveTaskRequest = async (taskRequestId, user) => {
```
iamitprakash, line 95: here the task with the assignee and status fields, but it will not update the approvedTo field. This means that the approvedTo field will still be the old value, even though the task has been approved.

iamitprakash (changes requested):

middlewares/taskRequests.js
TanishqSingla: We only want to allow users with `IDLE` status to apply for a task. There may be a case where a user has started working on a task but that task has some blockers, then the task status would be marked as `Blocked` and the user status would be `IDLE` hence allowing the user to request for a new task.

models/taskRequests.js
TanishqSingla: The `task` collection won't be having `approvedTo` field, the `taskRequest` collection will contain it, the task request collection will only contain the data regarding the task, who requested for it and whom it got approved. For the assigned status we're storing it in the `task` collection

iamitprakash (changes requested):
please add test coverage stats

TanishqSingla:
Coverage stats are as follows
| file                                           |  %stmts| %Branch | %Funcs | %Lines | Uncovered Lines |
| -------------------------------------- | ----------- | ------------- | ----------- | ---------- | ------------------------ |
|constants/taskRequests.js  |        100 |      100 |     100 |     100 |                   
|controllers/tasksRequests.js                     |   89.18 |    85.71 |     100 |   89.18 | 56-57,75-76  |
|middlewares/taskRequests.js                      |   86.95 |       90 |     100 |   86.95 | 21,39-40|
|models/taskRequests.js                      |   92.45 |      100 |     100 |   92.15 | 83-84,113-114    |

TanishqSingla:
@iamitprakash Have commented the test coverage please check

models/taskRequests.js
iamitprakash: @RitikJaiswal75  same here

iamitprakash (approved):
Done

DashDeipayan (approved):
LGTM

PR #932, opened 2023-02-27
Added validations for external-accounts data

rohan09-raj:
## What is the change?
- Added `joi` validation for the attributes object in externals-accounts

## Why is the change?
- Validation has been added to prevent random values from getting added to DB
- Only the selected attributes are required to make the `POST` request success

### DIFF
--- a/middlewares/validators/external-accounts.js
+++ b/middlewares/validators/external-accounts.js
@@ -1,11 +1,19 @@
 const joi = require("joi");
 
 const externalAccountData = async (req, res, next) => {
-  const schema = joi.object().strict().keys({
-    type: joi.string().required(),
-    token: joi.string().required(),
-    attributes: joi.object().strict().required(),
-  });
+  const schema = joi
+    .object()
+    .strict()
+    .keys({
+      type: joi.string().required(),
+      token: joi.string().required(),
+      attributes: joi.object().strict().required().keys({
+        discordId: joi.number().optional(),
+        username: joi.string().optional(),
+        picture: joi.string().optional(),
+        expiry: joi.number().optional(),
+      }),
+    });
 
   try {
     await schema.validateAsync(req.body);


middlewares/validators/external-accounts.js
```diff
@@ -1,11 +1,19 @@
...
+    .strict()
+    .keys({
+      type: joi.string().required(),
+      token: joi.string().required(),
+      attributes: joi.object().required().keys({
+        discordId: joi.number().optional(),
+        username: joi.string().optional(),
+        picture: joi.string().optional(),
+        expiry: joi.number().optional(),
+      }),
```
heyrandhir, line 15: just wanted to confirm that since all individual keys are optional. does attributes need to be an empty object if none of the child keys are there.

heyrandhir (approved):
LGTM

middlewares/validators/external-accounts.js
rohan09-raj: No attributes can't be empty object as it will contain the data of external accounts, and child keys will vary according to the type of external account

rohan09-raj: - I needed a suggestion on naming the `discordId` key, should we name it `discordId` or `id`. 
- The usecase of this API is generic and can be used for any external service or account. 
- But on the other hand, user of this API may get confused with some other id while using it for discord bots

heyrandhir: I remember @RitikJaiswal75 having a discussion with Ankush on the same , I think Ankush said we can call it `socialId` or `thirdPartyId` or something on similar lines. @RitikJaiswal75 what do you think.

rohan09-raj: Thanks, will confirm it

heyrandhir: > No attributes can't be empty object as it will contain the data of external accounts, and child keys will vary according to the type of external account

In this case we can add a validation that at least 1 child key is required . something like 
```
.keys({
      discordId: joi.number().optional(),
      username: joi.string().optional(),
      picture: joi.string().optional(),
      expiry: joi.number().optional(),
}).min(1)
```

heyrandhir (changes requested):

shmbajaj (changes requested):
Please Resolve @heyrandhir [comment](https://github.com/Real-Dev-Squad/website-backend/pull/932/files#r1119432534)

middlewares/validators/external-accounts.js
rohan09-raj: Made the change, thanks for the suggestion

heyrandhir (approved):
Good work @rohan09-raj :)

shmbajaj (approved):
✅ LGTM

PR #937, opened 2023-03-03
Add redirectURL params in github/login route

prakashchoudhary07:
`/auth/github/login?redirectURL="any-realdevsquad-website"`
Adds `redirectURL` query params, to redirect to the specified URL

Closes: https://github.com/Real-Dev-Squad/website-backend/issues/1213

image

### DIFF
--- a/controllers/auth.js
+++ b/controllers/auth.js
@@ -1,6 +1,22 @@
 const passport = require("passport");
 const users = require("../models/users");
 const authService = require("../services/authService");
+const REALDEVSQUAD_HOSTNAME = config.get("services.hostName");
+
+/**
+ * Makes authentication call to GitHub statergy
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ * @param next {Function} - Express middleware function
+ */
+const githubAuthLogin = (req, res, next) => {
+  const redirectURL = req.query.redirectURL;
+  return passport.authenticate("github", {
+    scope: ["user:email"],
+    state: redirectURL,
+  })(req, res, next);
+};
 
 /**
  * Fetches the user info from GitHub and authenticates User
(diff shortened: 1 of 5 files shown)


RitikJaiswal75 (commented):
Please make the changes discussed in the call.

RitikJaiswal75 (approved):

sahsisunny (dismissed):

controllers/auth.js
```diff
@@ -9,11 +25,22 @@ const authService = require("../services/authService");
...
+const githubAuthCallback = (req, res, next) => {
   let userData;
   const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
-  let authRedirectionUrl = req.query.state ?? rdsUiUrl;
-
+  let authRedirectionUrl = rdsUiUrl;
+  if ("state" in req.query) {
+    try {
+      const redirectUrl = new URL(req.query.state);
+      if (redirectUrl.hostname.endsWith(REALDEVSQUAD_HOSTNAME)) {
```
shubham-y, line 35: I do not think using `endWith()` is a good idea.
Because if is pass `state = https://rrealdevsquad.com`(notice the extra r at the beginning) the condition will evaluate to true

controllers/auth.js
```diff
@@ -1,6 +1,22 @@
 const passport = require("passport");
 const users = require("../models/users");
 const authService = require("../services/authService");
+const REALDEVSQUAD_HOSTNAME = new URL(config.get("services.rdsUi.baseUrl")).hostname;
```
shubham-y, line 4: Adding this constant here seems redundant as we already store a similar value variable `rdsUiUrl` on line 30.
We can get the hostname from it.

shubham-y (changes requested):

shubham-y (approved):

RitikJaiswal75 (approved):

PR #950, opened 2023-03-14
fix: Incorrect return type to addUser causes lint warning

### LINKED ISSUE #949: `await addUser()` causes lint warnings
https://github.com/Real-Dev-Squad/website-backend/blob/c8f53da7970aea9685a50d7c2a622625b4aee881/test/utils/addUser.js#L9

## Current Behavior
Currently the function `addUser` gives a lint warning that says `await is unecessary` here when used with `await` keyword
E.g
```js
const userID = await addUser(user);
```

## Expected Behavior
`await` is actually required here and the lint warning should not be here

##  Cause
The return type of the function which is provided is wrong.

TanishqSingla:
fixes #949 

     addUser is an async function however its return type was not
     wrapped in a promise which is why the when using the function with
     `await` there was a lint warning that showed up that saying `await
     is not needed`, but without await the function won't work
     properly.

### DIFF
--- a/test/utils/addUser.js
+++ b/test/utils/addUser.js
@@ -6,7 +6,7 @@ const userData = require("../fixtures/user/user")();
 /**
  * File to be required in every test file where userId is required to generate the JWT
  *
- * @return {string} userId - userId for the added user
+ * @return {Promise<string>} userId - userId for the added user
  */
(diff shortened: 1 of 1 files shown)


prakashchoudhary07 (approved):
Thanks for this 🚀

PR #976, opened 2023-03-30
Test cases for the user filter endpoint based on status and skill tag

heyrandhir:
- [x] Setup tests for user management filter endpoint 
- [x] Currently filters based on status and skill tag

PR #978, opened 2023-04-01
Feature - Filter Users based on their state (OOO,ACTIVE,IDLE) and skills

heyrandhir:
- [x] Added a new endpoint to fetch user data based on a filter. Users can be filtered based on their skills and status i.e idle, active, or OOO.
- [x]  Used a combination of two models itemModel and userStatusModel to fetch the relevant user data.
- [x]  Added default filters for itemType i.e USER and tagType i.e SKILL in the calls.item query.
- [x]  Added a helper function in utils to check if two arrays have any common items.
- [x]  Added appropriate comments for the function and its parameters to help 

Note : The test for this PR is raised in a separate PR [here](https://github.com/Real-Dev-Squad/website-backend/pull/976)

### DIFF
--- a/constants/users.js
+++ b/constants/users.js
@@ -11,8 +11,8 @@ const USER_STATUS = {
 };
 
 const ALLOWED_FILTER_PARAMS = {
-  itemTag: ["levelId", "levelName", "levelNumber", "tagId"],
-  userState: ["state"],
+  ITEM_TAG: ["levelId", "levelName", "levelNumber", "tagId"],
+  USER_STATE: ["state"],
 };
 
 module.exports = { profileStatus, USER_STATUS, ALLOWED_FILTER_PARAMS };
--- a/constants/users.js
+++ b/constants/users.js
@@ -10,9 +10,9 @@ const USER_STATUS = {
   ACTIVE: "active",
 };
 
-const allowedFilterParams = {
+const ALLOWED_FILTER_PARAMS = {
   itemTag: ["levelId", "levelName", "levelNumber", "tagId"],
   userState: ["state"],
 };
 
-module.exports = { profileStatus, USER_STATUS, allowedFilterParams };
+module.exports = { profileStatus, USER_STATUS, ALLOWED_FILTER_PARAMS };
--- a/constants/users.js
+++ b/constants/users.js
@@ -11,7 +11,7 @@ const USER_STATUS = {
 };
 
 const ALLOWED_FILTER_PARAMS = {
-  ITEM_TAG: ["levelId", "levelName", "levelNumber", "tagId"],
+  ITEM_TAG: ["levelId", "levelName", "levelValue", "tagId"],
   USER_STATE: ["state"],
 };
 
--- a/constants/users.js
+++ b/constants/users.js
@@ -10,4 +10,9 @@ const USER_STATUS = {
   ACTIVE: "active",
 };
 
-module.exports = { profileStatus, USER_STATUS };
+const allowedFilterParams = {
+  itemTag: ["levelId", "levelName", "levelNumber", "tagId"],
+  userState: ["state"],
+};
+
+module.exports = { profileStatus, USER_STATUS, allowedFilterParams };
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -417,6 +417,31 @@ const addDefaultArchivedRole = async (req, res) => {
   }
 };
 
+/**
+ * Returns the lists of users who match the specified query params
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const filterUserBasedOnQuery = async (req, res) => {
+  try {
+    if (!Object.keys(req.query).length) {
+      return res.boom.badRequest("filter for item not provided");
+    }
+    // const { allUsers } = await userQuery.getUsersBasedOnFilter(req.query);
+    const allUsers = await userQuery.getUsersBasedOnFilter(req.query);
+
+    return res.json({
+      message: "Users found successfully!",
+      users: allUsers,
+    });
+  } catch (error) {
+    logger.error(`Error while fetching all users: ${error}`);
+    return res.boom.serverUnavailable("Something went wrong please contact admin");
+  }
+};
+
 module.exports = {
   verifyUser,
   generateChaincode,
@@ -435,4 +460,5 @@ module.exports = {
   getUserIntro,
   addDefaultArchivedRole,
   getUserSkills,
+  filterUserBasedOnQuery,
 };
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -424,7 +424,7 @@ const addDefaultArchivedRole = async (req, res) => {
  * @param res {Object} - Express response object
  */
 
-const filterUserBasedOnQuery = async (req, res) => {
+const filterUsers = async (req, res) => {
   try {
     if (!Object.keys(req.query).length) {
       return res.boom.badRequest("filter for item not provided");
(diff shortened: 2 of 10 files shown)


models/users.js
```diff
@@ -292,6 +294,66 @@ const fetchUserSkills = async (id) => {
...
+ * @param {Array} query.levelName - Array of levelNames to filter the users on
+ * @param {Array} query.levelNumber - Array of levelNumbers to filter the users on
+ * @param {Array} query.tagId - Array of tagIds to filter the users on
+ * @param {Array} query.state - Array of states to filter the users on
+ * @return {Promise<Array>} - Array of user documents that match the filter criteria
+ */
+
+const getUsersBasedOnFilter = async (query) => {
+  const itemTagsKeys = ["levelId", "levelName", "levelNumber", "tagId"];
+  const userStateKeys = ["state"];
```
Maheima, line 311: Can these be added to a constants file?

constants/users.js
```diff
@@ -10,4 +10,9 @@ const USER_STATUS = {
   ACTIVE: "active",
 };
 
-module.exports = { profileStatus, USER_STATUS };
+const allowedFilterParams = {
+  itemTag: ["levelId", "levelName", "levelNumber", "tagId"],
+  userState: ["state"],
```
Maheima, line 15: nit: can you make the name of the variables snake-cased uppercase characters?

Maheima (dismissed):

utils/array.js
```diff
@@ -18,6 +18,30 @@ function chunks(array, size = 1) {
...
 
+/**
+ * Checks if two arrays have any common items
+ * @param array1 {Array} - first array
+ * @param array2 {Array} - second array
+ * @returns {boolean} - true if the arrays have at least one common item, false otherwise
+ */
+
+function arraysHaveCommonItem(array1, array2) {
+  if (!Array.isArray(array1) || !Array.isArray(array2) || array1.length > 100 || array2.length > 100) {
```
ankushdharkar, line 29: Why does this function care if the array lengths are more than a 100?

constants/users.js
```diff
@@ -10,4 +10,9 @@ const USER_STATUS = {
   ACTIVE: "active",
 };
 
-module.exports = { profileStatus, USER_STATUS };
+const ALLOWED_FILTER_PARAMS = {
+  ITEM_TAG: ["levelId", "levelName", "levelNumber", "tagId"],
```
ankushdharkar, line 14: This should be `levelValue` instead of `levelNumber`.

controllers/users.js
```diff
@@ -417,6 +417,31 @@ const addDefaultArchivedRole = async (req, res) => {
...
 };
 
+/**
+ * Returns the lists of users who match the specified query params
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const filterUserBasedOnQuery = async (req, res) => {
```
ankushdharkar, line 427: Does `BasedOnQuery` suffix in the name deliver additional value?

utils/array.js
```diff
@@ -18,6 +18,30 @@ function chunks(array, size = 1) {
...
+  }
+
+  for (let i = 0; i < array1.length; i++) {
+    for (let j = 0; j < array2.length; j++) {
+      /* eslint-disable security/detect-object-injection */
+      if (array1[i] === array2[j]) {
+        return true;
+      }
+    }
+  }
```
ankushdharkar, line 40: Can this be done using some proper array methods (and maybe using sets/maps)?

ankushdharkar (changes requested):

controllers/users.js
heyrandhir: `BasedOnQuery` suffix is not providing additional value to the function name. I will remove it.

utils/array.js
heyrandhir: I had added it for safety check but since we have a joi validator, this makes no sense, I will remove it.

utils/array.js
heyrandhir: Implemented.

controllers/users.js
```diff
@@ -417,6 +417,31 @@ const addDefaultArchivedRole = async (req, res) => {
   }
 };
 
+/**
```
vvaibhavdesai, line 420: I don't think we need to add comment for the `req` and `res`

utils/array.js
```diff
@@ -18,6 +18,21 @@ function chunks(array, size = 1) {
...
 
+/**
+ * Checks if two arrays have any common items
+ * @param array1 {Array} - first array
+ * @param array2 {Array} - second array
+ * @returns {boolean} - true if the arrays have at least one common item, false otherwise
+ */
+
+function arraysHaveCommonItem(array1, array2) {
+  if (!Array.isArray(array1) || !Array.isArray(array2) || !array1.length || !array2.length) {
```
vvaibhavdesai, line 29: what are we trying to check here?

heyrandhir: If one of the array don't have any length or is not an array then I am returning false instead of looping and finding common items.

utils/array.js
```diff
@@ -18,6 +18,21 @@ function chunks(array, size = 1) {
   return result;
 }
 
+/**
+ * Checks if two arrays have any common items
+ * @param array1 {Array} - first array
```
shmbajaj, line 23: Add Array<item> type into the description, as this function will be reused.
Also add this method is for singular values only such as strings and numbers.

models/users.js
```diff
@@ -292,6 +295,65 @@ const fetchUserSkills = async (id) => {
...
+ * @param {Array} query.levelId - Array of levelIds to filter the users on
+ * @param {Array} query.levelName - Array of levelNames to filter the users on
+ * @param {Array} query.levelNumber - Array of levelNumbers to filter the users on
+ * @param {Array} query.tagId - Array of tagIds to filter the users on
+ * @param {Array} query.state - Array of states to filter the users on
+ * @return {Promise<Array>} - Array of user documents that match the filter criteria
+ */
+
+const getUsersBasedOnFilter = async (query) => {
+  const { ITEM_TAG, USER_STATE } = ALLOWED_FILTER_PARAMS;
```
shmbajaj, line 311: **Suggestion:** Move this line to outside of the function, as this is constant and will be reused again.

**PS:** Won't make a big change/benefit.

shmbajaj (commented):
LGTM ✅

utils/array.js
vvaibhavdesai: do we need to check Array.isArray ?

can we just check if the array1?.length property exists on the array if yes then it will return false since 0 is falsy in case array is empty & if its not it will continue to execute your written logic

heyrandhir: sure will do it.

vvaibhavdesai (approved):

models/users.js
```diff
@@ -292,6 +295,64 @@ const fetchUserSkills = async (id) => {
...
+ * @param {Array} query.levelId - Array of levelIds to filter the users on
+ * @param {Array} query.levelName - Array of levelNames to filter the users on
+ * @param {Array} query.levelNumber - Array of levelNumbers to filter the users on
+ * @param {Array} query.tagId - Array of tagIds to filter the users on
+ * @param {Array} query.state - Array of states to filter the users on
+ * @return {Promise<Array>} - Array of user documents that match the filter criteria
+ */
+
+const getUsersBasedOnFilter = async (query) => {
+  const allQueryKeys = Object.keys(query);
```
ivinayakg, line 311: can't we run Object.entries() once and save it to use at other places.

heyrandhir: again after running the object.entries() we would need to extract the `keys` from that entries as the `arraysHaveCommonItem` function only excepts an array of strings .

Issue #983, opened 2023-04-02
Fix the route in integration tests, for API `users/userId/:id` changing to `users/?id=abc`


vikhyat187:
As we are changing the API route the test cases are also to be updated as per the new route `users/?id=abc`

PR #984, opened 2023-04-03
Fix the route in integration tests, for API users/userId/:id changing to users/?id=abc #983

endeavourmonk:


prakashchoudhary07:
Tests are failing?
Please have a look

Issue #981, opened 2023-04-02
Remove tag and level attributes from itemTag object


DashDeipayan:
# Task
Remove attributes like `tagname`, `tagtype`, `levelname` and `levelvalue` from `itemTag` object.
# Why was it added?
- Avoid two more API calls to get data of `level` and `tag` based on their respective foreign keys
- Ease of data usage in the front end.

# Why is it to be removed?
- In case of changes are made to the attribute values of a specific `tag` or `level`, there will be inconsistency across collections as the newly made changes won't be available in the existing documents in the `ItemTags` collection

PR #986, opened 2023-04-03
Fixes for itemtags

DashDeipayan:
# Closes: #981

### DIFF
--- a/controllers/items.js
+++ b/controllers/items.js
@@ -9,9 +9,9 @@ const ItemTagsModel = require("../models/items");
 
 const addTagsToItem = async (req, res) => {
   try {
-    const { itemId } = await ItemTagsModel.addTagsToItem(req.body);
-    return res.json({
-      message: "Tags added successfully!",
+    const { itemId, isNewTag } = await ItemTagsModel.addTagsToItem(req.body);
+    return res.status(isNewTag ? 201 : 400).json({
+      message: isNewTag ? "Tags added successfully!" : "Tags already exists!",
       itemId,
     });
   } catch (err) {
--- a/models/items.js
+++ b/models/items.js
@@ -1,8 +1,6 @@
 const firestore = require("../utils/firestore");
 
-const itemTagsModel = firestore.collection("itemTags");
-const tagModel = firestore.collection("tags");
-const levelModel = firestore.collection("levels");
+const itemTagsModel = firestore.collection("items");
 
 /**
  *
@@ -15,17 +13,17 @@ const addTagsToItem = async (itemData) => {
     const { itemId, itemType, tagPayload } = itemData;
     const batch = firestore.batch();
     for (const tag of tagPayload) {
-      const tagData = await tagModel.doc(tag.tagId).get();
-      const levelData = await levelModel.doc(tag.levelId).get();
+      const itemData = await itemTagsModel
+        .where("itemId", "==", itemId)
+        .where("tagId", "==", tag.tagId)
+        .where("levelId", "==", tag.levelId)
+        .get();
+      if (!itemData.empty) continue;
       const itemTag = {
         itemId,
         itemType: itemType.toUpperCase(),
         tagId: tag.tagId,
-        tagName: tagData.data().name,
-        tagType: tagData.data().type,
         levelId: tag.levelId,
-        levelName: levelData.data().name,
-        levelValue: levelData.data().value,
       };
       const docid = itemTagsModel.doc();
       batch.set(docid, itemTag);
--- a/models/items.js
+++ b/models/items.js
@@ -12,6 +12,7 @@ const addTagsToItem = async (itemData) => {
   try {
     const { itemId, itemType, tagPayload } = itemData;
     const batch = firestore.batch();
+    let isNewtag = false;
     for (const tag of tagPayload) {
       const itemData = await itemTagsModel
         .where("itemId", "==", itemId)
@@ -19,6 +20,7 @@ const addTagsToItem = async (itemData) => {
         .where("levelId", "==", tag.levelId)
         .get();
       if (!itemData.empty) continue;
+      isNewtag = true;
       const itemTag = {
         itemId,
         itemType: itemType.toUpperCase(),
@@ -29,7 +31,7 @@ const addTagsToItem = async (itemData) => {
       batch.set(docid, itemTag);
     }
     await batch.commit();
-    return { itemId };
+    return { itemId, isNewtag };
   } catch (err) {
     logger.error("Error in creating Item", err);
     throw err;


models/items.js
```diff
@@ -14,24 +12,26 @@ const addTagsToItem = async (itemData) => {
...
+      isNewtag = true;
       const itemTag = {
         itemId,
         itemType: itemType.toUpperCase(),
         tagId: tag.tagId,
-        tagName: tagData.data().name,
-        tagType: tagData.data().type,
         levelId: tag.levelId,
-        levelName: levelData.data().name,
-        levelValue: levelData.data().value,
```
vinayak-trivedi, line 28: have we missed fetching these details on the get API call for getting the item based on filter, should we not fetch all the details of tag and levels and send it when we are fetching the item

vinayak-trivedi (changes requested):
other than the request changes everything else looks good to me

models/items.js
```diff
@@ -14,24 +12,26 @@ const addTagsToItem = async (itemData) => {
...
+        .where("levelId", "==", tag.levelId)
+        .get();
+      if (!itemData.empty) continue;
+      isNewtag = true;
       const itemTag = {
         itemId,
         itemType: itemType.toUpperCase(),
         tagId: tag.tagId,
-        tagName: tagData.data().name,
-        tagType: tagData.data().type,
```
RitikJaiswal75, line 25: I guess the tag type should exist. It will help in filtering in future to fetch users with specific tag and tasks with specific tags

models/items.js
```diff
@@ -14,24 +12,26 @@ const addTagsToItem = async (itemData) => {
...
     for (const tag of tagPayload) {
-      const tagData = await tagModel.doc(tag.tagId).get();
-      const levelData = await levelModel.doc(tag.levelId).get();
+      const itemData = await itemTagsModel
+        .where("itemId", "==", itemId)
+        .where("tagId", "==", tag.tagId)
+        .where("levelId", "==", tag.levelId)
+        .get();
+      if (!itemData.empty) continue;
+      isNewtag = true;
```
RitikJaiswal75, line 23: if the data is not empty then we are marking the new tag as true? shouldn't it be the opposite if the item data of the specified tag is empty i.e. not exists then it is a new tag?

models/items.js
DashDeipayan: I have added fetching the tag data only based on ID according to the current requirements. We can discuss this with Ankush.

models/items.js
DashDeipayan: Sorry didn't get you. If the `if` statement is executed, the `isNewTag` value never changes. It changes only when `if` isn't true i.e. a new tag. Hence, `isNewTag = true`

models/items.js
DashDeipayan: If we remove these values, we can only filter based on tags and levels(tagId and levelId specifically). Other types of filtering wont work as of now

vinayak-trivedi (approved):
LGTM

models/items.js
```diff
@@ -14,24 +14,26 @@ const addTagsToItem = async (itemData) => {
...
       const itemTag = {
         itemId,
         itemType: itemType.toUpperCase(),
         tagId: tag.tagId,
-        tagName: tagData.data().name,
-        tagType: tagData.data().type,
```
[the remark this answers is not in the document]
DashDeipayan: done

controllers/items.js
```diff
@@ -9,9 +9,9 @@ const ItemTagsModel = require("../models/items");
 
 const addTagsToItem = async (req, res) => {
   try {
-    const { itemId } = await ItemTagsModel.addTagsToItem(req.body);
-    return res.json({
-      message: "Tags added successfully!",
+    const { itemId, isNewTag } = await ItemTagsModel.addTagsToItem(req.body);
+    return res.status(isNewTag ? 201 : 400).json({
```
ankushdharkar, line 13: Should this be a 403 instead of 400? (since 400 represents malformed request)

ankushdharkar, line 13: Could this be cleaned up a bit please using an early return for `if(!isNewTag) { ... }`?

models/items.js
```diff
@@ -1,8 +1,8 @@
 const firestore = require("../utils/firestore");
 
-const itemTagsModel = firestore.collection("itemTags");
-const tagModel = firestore.collection("tags");
-const levelModel = firestore.collection("levels");
+const itemTagsModel = firestore.collection("items");
```
ankushdharkar, line 3: Is this name correct?

models/items.js
```diff
@@ -14,24 +14,26 @@ const addTagsToItem = async (itemData) => {
   try {
     const { itemId, itemType, tagPayload } = itemData;
```
ankushdharkar, line 15: Instead of `tagPayload`, can we call it `tags` please?

models/items.js
```diff
@@ -14,24 +14,26 @@ const addTagsToItem = async (itemData) => {
...
     const batch = firestore.batch();
+    let isNewtag = false;
     for (const tag of tagPayload) {
-      const tagData = await tagModel.doc(tag.tagId).get();
-      const levelData = await levelModel.doc(tag.levelId).get();
+      const itemData = await itemTagsModel
```
ankushdharkar, line 19: Please use `Promise.all` and parallelize these calls to speed up fetching.

ankushdharkar (changes requested):

controllers/items.js
```diff
@@ -9,8 +9,9 @@ const ItemTagsModel = require("../models/items");
 
 const addTagsToItem = async (req, res) => {
   try {
-    const { itemId } = await ItemTagsModel.addTagsToItem(req.body);
-    return res.json({
+    const { itemId, wasSuccess } = await ItemTagsModel.addTagsToItem(req.body);
```
pallabez, line 12: We don't know what is `req.body`. This would cause trouble when migrating to Typescript and when refactoring.

models/items.js
```diff
@@ -14,24 +14,41 @@ const addTagsToItem = async (itemData) => {
...
+          .where("levelId", "==", tag.levelId)
+          .get()
+          .then((doc) => {
+            isNewTag[tag.tagId] = doc.empty;
+          })
+      );
     }
+
+    await Promise.all(hasItemTagsPromises);
+    const wasSuccess = Object.values(isNewTag).some((value) => value === true);
```
pallabez, line 33: What are we checking here?

models/items.js
```diff
@@ -14,24 +14,51 @@ const addTagsToItem = async (itemData) => {
...
+    /**
+     * In the above code, we are looping across every tag that is to be set and creating an Object that will
+     * finally be used to check if atleast ONE of the tags is a new tag. Because if every tag is an existing
+     * tag, it is unnecessary to proceed any further. The `isNewTag` Object will look like
+     * {
+     *    some_tag_id_1: false,
+     *    some_tag_id_2: true
+     * }
+     * `some_tag_id_2:true` represent a new tag and false in the vice-versa.
+     */
```
vinayak-trivedi, line 40: thanks for this

vinayak-trivedi (approved):
Thanks for the changes @DashDeipayan, LGTM

pallabez (approved):
LGTM

models/items.js
```diff
@@ -14,24 +14,51 @@ const addTagsToItem = async (itemData) => {
...
+
+    await Promise.all(hasItemTagsPromises);
+    const wasSuccess = Object.values(isNewTag).some((value) => value === true); // Atleast one new tag
+
+    if (!wasSuccess) return { itemId, wasSuccess }; // Returns if none of the tags are new
+
+    tagPayload.forEach((tag) => {
```
iamitprakash, line 47: The isNewTag object is not being updated, so the if statement will always be true. This means that all of the tags will be added to the database, even if they are not new. To fix this, you need to update the isNewTag object after each iteration of the loop

iamitprakash (changes requested):

models/items.js
DashDeipayan: the `isNewTag` Object has a structure like
```
{
    some_tag_id_1: false,
    some_tag_id_2: true,
    .
    .
    .
}
```
Before adding a new document, it checks if the value of the tagId is true or not, then and only then it adds. 

Maybe I am missing your point😅

models/items.js
```diff
@@ -76,14 +103,37 @@ const getItemBasedOnFilter = async (query) => {
     const items = [];
     const data = await call.get();
 
-    data.forEach((doc) => {
```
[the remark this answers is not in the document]
DashDeipayan: That was the aim actually. In a for loop, awaiting code can cause a performance bottleneck, that's why I awaited it out of the for loop. In line 136, that's why `addTagsAndLevelsData` will only have the data after this line.

PR #991, opened 2023-04-05
Fix the route in integration tests, for API users/userId/:id changing to users/?id=abc #983

endeavourmonk:
Fix the route in integration test


test/integration/users.test.js
```diff
@@ -462,11 +462,11 @@ describe("Users", function () {
     });
   });
 
-  describe("GET /users/userId/id", function () {
+  describe.only("GET /users/?id", function () {
```
vikhyat187, line 465: @endeavourmonk please remove .only it was only to test in local

PR #1001, opened 2023-04-06
filter members option in user API

### LINKED ISSUE #1000: Add filter for members on `/users` route
- when member=true query params is passed to the `/users` route as query params return all the RDS members who are not archieved

vinayak-trivedi:
closes #1000

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -130,6 +130,25 @@ const fetchUsers = async (query) => {
   try {
     // INFO: default user size set to 100
     // INFO: https://github.com/Real-Dev-Squad/website-backend/pull/873#discussion_r1064229932
+    if (query.member === "true") {
+      const membersArray = [];
+      const snapshot = await userModel.where("roles.member", "==", true).where("roles.archived", "==", false).get();
+      if (!snapshot.empty) {
+        snapshot.forEach((doc) => {
+          membersArray.push({
+            id: doc.id,
+            ...doc.data(),
+            phone: undefined,
+            email: undefined,
+            tokens: undefined,
+            chaincode: undefined,
+          });
+        });
+      }
+      return {
+        allUsers: membersArray,
+      };
+    }
     const size = parseInt(query.size) || 100;
     const doc = (query.next || query.prev) && (await userModel.doc(query.next || query.prev).get());
     let dbQuery = (query.prev ? userModel.limitToLast(size) : userModel.limit(size)).orderBy("username");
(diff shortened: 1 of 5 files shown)


models/users.js
```diff
@@ -130,6 +130,25 @@ const fetchUsers = async (query) => {
...
+      const snapshot = await userModel.where("roles.member", "==", true).where("roles.archived", "==", false).get();
+      if (!snapshot.empty) {
+        snapshot.forEach((doc) => {
+          membersArray.push({
+            id: doc.id,
+            ...doc.data(),
+            phone: undefined,
+            email: undefined,
+            tokens: undefined,
+            chaincode: undefined,
```
RitikJaiswal75, line 144: can we not send them instead of setting them undefined?

test/integration/users.test.js
```diff
@@ -357,6 +358,20 @@ describe("Users", function () {
...
+
+      expect(response).to.have.status(200);
+      expect(response.body).to.be.a("object");
+      expect(response.body.message).to.equal("Users returned successfully!");
+      expect(response.body).to.have.property("links");
+      expect(response.body.links).to.have.property("next");
+      expect(response.body.links).to.have.property("prev");
+
+      expect(response.body.users[0].roles.member).to.be.equal(true);
+      expect(response.body.users[0].roles.archived).to.be.equal(false);
```
RitikJaiswal75, line 373: can we also test the length of the response returned?

models/users.js
vinayak-trivedi: we do not want to send these details

test/integration/users.test.js
vinayak-trivedi: I think we do not need to test it

models/users.js
```diff
@@ -130,6 +130,25 @@ const fetchUsers = async (query) => {
...
+          });
+        });
+      }
+      return {
+        allUsers: membersArray,
+      };
+    }
```
prakashchoudhary07, line 151: Please handle this in the bottom part, where users are being fetched. We are just repeating code here.
There is no pagination here, even though user API has pagination.

And instead of having a key where member=true, let's generic field like a filter or role
Where we pass filterBy=MEMBERS, filterBy=AO or role=ARCHIVED, role=NEW, with paginations

prakashchoudhary07 (changes requested):

prakashchoudhary07:
Can you please share the data model and API contracts for it?

PR #1022, opened 2023-04-11
Fix: update fetchAllUsers to fetch only filtered users from firebase 

khalatevarun:


models/users.js
```diff
@@ -178,16 +178,16 @@ const fetchPaginatedUsers = async (query) => {
   }
 };
 
-const fetchAllUsers = async () => {
+const fetchFilteredUsers = async (usernames = []) => {
   try {
     const dbQuery = userModel;
 
-    const snapshot = await dbQuery.get();
+    const snapshot = await dbQuery.where("github_id", "in", usernames).get();
```
ankushdharkar, line 185: Is there a limit to how many usernames can exist in the `in` clause?

khalatevarun: Update: 
image

Check for an example, `in` supports 30 comparison values.

Thanks for pointing this out, will handle this case.

models/users.js
```diff
@@ -178,26 +178,39 @@ const fetchPaginatedUsers = async (query) => {
...
 
-const fetchAllUsers = async () => {
+const fetchFilteredUsers = async (usernames = []) => {
   try {
     const dbQuery = userModel;
+    const filterdUsersWithDetails = [];
 
-    const snapshot = await dbQuery.get();
+    const groups = [];
+    const batchSize = 30; // since only 30 comparisons are allowed with the 'in' clause
```
ankushdharkar, line 187: @khalatevarun Would be nice to move this to a "Firestore Constants" file, so we have a single source of truth for all knowledge and usage to things related to Firestore specifics. That way, in future, let's say batch size changes to 50 in a new firestore version, then we can just update the config constant in one place, and it will apply throughout the codebase.

khalatevarun: sure, makes sense
have moved the constant to firebase.js under constants directory

models/users.js
```diff
@@ -178,26 +179,38 @@ const fetchPaginatedUsers = async (query) => {
...
+    snapshots.forEach((snapshot) => {
+      snapshot.forEach((doc) => {
+        filterdUsersWithDetails.push({
+          id: doc.id,
+          ...doc.data(),
+          phone: undefined,
+          email: undefined,
+          tokens: undefined,
+          chaincode: undefined,
+        });
```
ankushdharkar, line 208: In future, we should create a util or a DAO to ensure we can fetch user data with given explicit access clearance based on [data classifications](https://docs.aws.amazon.com/whitepapers/latest/data-classification/data-classification-models-and-schemes.html)

models/users.js
```diff
@@ -178,26 +179,38 @@ const fetchPaginatedUsers = async (query) => {
...
+
+    snapshots.forEach((snapshot) => {
+      snapshot.forEach((doc) => {
+        filterdUsersWithDetails.push({
+          id: doc.id,
+          ...doc.data(),
+          phone: undefined,
+          email: undefined,
+          tokens: undefined,
+          chaincode: undefined,
```
ankushdharkar, line 207: Good handling here!

models/users.js
```diff
@@ -178,26 +179,38 @@ const fetchPaginatedUsers = async (query) => {
   }
 };
 
-const fetchAllUsers = async () => {
+const fetchFilteredUsers = async (usernames = []) => {
```
ankushdharkar, line 182: I feel this function name should just be `fetchUsers(providedUsernames)`

Why should this function be aware that the list was 'filtered' down somewhere else?

It's contract is simple: Give it a list of usernames, and it will give you back the data for all those usernames.

ankushdharkar (changes requested):

utils/users.js
```diff
@@ -121,24 +121,14 @@ function getPaginationLink(query, cursor, documentId) {
  * @param allUsers {Array} - list of total users from firebase
  *
  */
-function getFilteredUsers(allPRs, allUsers) {
+function getFilteredUsernames(allPRs) {
```
ankushdharkar, line 124: `getFilteredUsernames` doesn't tell me much. This is more like a list of users from PRs. What can we name this to make more sense later.

If we were to filter a list from discord messages tomorrow, we won't call that function `getFilteredUsernames2`, right?

khalatevarun: `getUsernamesFromPRs` can be apt here?

utils/users.js
```diff
@@ -121,24 +121,14 @@ function getPaginationLink(query, cursor, documentId) {
...
-  uniqueUsersInOrder.forEach((username) => {
-    const userDetails = allUsers.find((user) => user.github_id === username);
-
-    if (userDetails) {
-      usersWithDetails.push(userDetails);
-    }
-  });
-
-  return usersWithDetails;
+  return Array.from(uniqueUsersInOrder);
```
ankushdharkar, line 131: `uniqueUsersInOrder` is misleading variable name, as it's a Set and no order is guaranteed in it. We should rename it.

utils/users.js
ankushdharkar: We give it a list of PRs, and the function extracts a list of unique usernames.

So, how about `getContributingUsernames`?

models/users.js
```diff
@@ -178,26 +179,38 @@ const fetchPaginatedUsers = async (query) => {
...
+          email: undefined,
+          tokens: undefined,
+          chaincode: undefined,
+        });
       });
     });
+
     return {
-      allUsers,
+      filterdUsersWithDetails,
```
prakashchoudhary07, line 213: We should rename this to something  like `users` instead of `filteredUsersWithDetails`, or to something else

utils/users.js
```diff
@@ -148,5 +143,5 @@ module.exports = {
   getParticipantUsernames,
   getLowestLevelSkill,
   getPaginationLink,
-  getFilteredUsers,
+  getUsernamesFromPRs,
```
ankushdharkar, line 146: Why did we not change any corresponding tests? Are we missing unit tests?

prakashchoudhary07 (changes requested):

ankushdharkar (approved):

models/users.js
ankushdharkar: Good point Prakash. Unfortunately, I jumped the gun and merged 😢

ankushdharkar: Will get someone to fix it here: https://github.com/Real-Dev-Squad/website-backend/issues/1026

PR #1035, opened 2023-04-18
Fix/new user redirection

vinayak-trivedi:
- fixes redirection to new signup flow for new users

### DIFF
--- a/test/integration/auth.test.js
+++ b/test/integration/auth.test.js
@@ -6,6 +6,8 @@ const passport = require("passport");
 
 const app = require("../../server");
 const cleanDb = require("../utils/cleanDb");
+const { addUserToDBForTest } = require("../../utils/users");
+const userData = require("../fixtures/user/user")();
 
 chai.use(chaiHttp);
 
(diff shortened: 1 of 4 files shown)


utils/users.js
```diff
@@ -1,11 +1,22 @@
 const { fetchUser } = require("../models/users");
+const firestore = require("../utils/firestore");
+const userModel = firestore.collection("users");
+
+const addUserToDBForTest = async (userData) => {
+  try {
+    await userModel.add(userData);
+  } catch (err) {
+    // console.log(err);
```
bhtibrewal, line 9: why aren't you handling error here?

vinayak-trivedi: as discussed in the call, removing try catch

bhtibrewal (approved):
looks good to merge

test/integration/auth.test.js
```diff
@@ -19,28 +21,38 @@ describe("auth", function () {
...
-      .end((err, res) => {
-        if (err) {
-          return done(err);
-        }
-        expect(res).to.have.status(302);
-        expect(res.headers.location).to.equal(rdsUiUrl);
-
-        return done();
-      });
+      .query({ code: "codeReturnedByGithub", state: rdsUiUrl, shouldConsole: true })
```
prakashchoudhary07, line 52: What is `shouldConsole: true` here

prakashchoudhary07 (dismissed):

test/integration/auth.test.js
```diff
@@ -19,28 +21,38 @@ describe("auth", function () {
     sinon.restore();
   });
 
-  it("should redirect the request to the goto page on successful login", function (done) {
+  it("should redirect the user to new sign up flow if they are have incomplte user details true", async function () {
+    const redirectURL = "https://my.realdevsquad.com/new-signup";
+    sinon.stub(passport, "authenticate").callsFake((strategy, options, callback) => {
+      callback(null, "accessToken", githubUserInfo[0]);
+      return (req, res, next) => {};
```
heyrandhir, line 28: why are we returning an empty function here ? I think this can be omitted.

test/integration/auth.test.js
```diff
@@ -19,28 +21,38 @@ describe("auth", function () {
     sinon.restore();
   });
 
-  it("should redirect the request to the goto page on successful login", function (done) {
+  it("should redirect the user to new sign up flow if they are have incomplte user details true", async function () {
```
heyrandhir, line 24: typo `incomplte` !

heyrandhir, line 24: why are we not adding user to db unlike the next assertion 🤔

Issue #1004, opened 2023-04-07
Issue running tests on local


fakhruddinkw:
**What is the issue?**
Tests are not running on the local.

**What is the reason?**

1. Firestore emulator not running.
2. Java version is not above 11
3. When we run yarn test, it runs both the unit and integration tests (in this order). So after the unit tests are done, the java process is not killed automatically and when our integration test run it gives error.

**Possible solution** (in order)

1. add '--project=/your firestor app name/ ' flag to both the test scripts.
2. Java version above 11 is needed for firebase tool version >= 11
3. Either manually kill the java process after unit tests are done or run both the tests separately by running the test commands.

PR #1042, opened 2023-04-23
Fix for running tests locally

shubham-y:
Issue #1004 (point 1)

Changes made:

- Updated the package `firebase-tools` version to the latest(i.e. 11.28.0) to make it compatible with node v18.x
- Updated both test scripts to get the value of `project_id` from firestore config and pass it as `--project=$project_id`(https://firebase.google.com/docs/emulator-suite/install_and_configure)


scripts/tests/testIntegration.sh
```diff
@@ -3,5 +3,9 @@
 # set 'test' environment
 export NODE_ENV='test'
 
+# get project_id value from firestore config
+json=$(node -e "console.log(require('config').get('firestore'))")
+project_id=$(echo $json | grep -o '"project_id":[^,}]*' | cut -d':' -f2 | tr -d '"' | tr -d '[:space:]')
```
shubham-y, line 8: Brief explanation of the above line

`grep -o '"project_id":[^,}]*`: This command uses grep to search for the pattern "project_id": followed by any number of characters that are not , or }. The -o flag tells grep to only output the matched text.

`cut -d':' -f2`: This command uses cut to extract the second field from the output of grep, using : as the delimiter. Since the output of grep is "project_id": "real-dev-squad-backend-abda3", the second field is "real-dev-squad-backend-abda3".

`tr -d '"'`: This command uses tr to delete any occurrences of " characters from the output of cut.

`tr -d '[:space:]'`: This command uses tr to delete any whitespace characters from the output of the previous command.

RitikJaiswal75 (approved):

yarn.lock
RitikJaiswal75, line 1: why has this yarn lock changed?

shubham-y: Because we have updated the version of ` firebase-tools` to latest

PR #1060, opened 2023-04-28
Feature: inDiscord

bhtibrewal:


controllers/users.js
```diff
@@ -24,7 +24,11 @@ const verifyUser = async (req, res) => {
     logger.error(`Error while verifying user: ${error}`);
     return res.boom.serverUnavailable(SOMETHING_WENT_WRONG);
   }
-  fetch(process.env.IDENTITY_SERVICE_URL, "POST", null, { userId }, { "Content-Type": "application/json" });
+  fetch(process.env.IDENTITY_SERVICE_URL, {
+    method: "POST",
+    body: { userId },
+    headers: { "Content-Type": "application/json" },
+  });
```
bhtibrewal, line 31: chore: removed the custom fetch so made a minor change in the syntax here

models/users.js
```diff
@@ -412,6 +414,34 @@ const getUsersBasedOnFilter = async (query) => {
...
+        id: doc.id,
+        ...doc.data(),
+      });
+    });
+
+    return filteredUsers.filter((user) => !user.roles?.archived);
+  }
+  if (verifiedQuery === "true") {
+    const filteredUsers = [];
+    const snapshot = await userModel.where("discordId", "!=", null).get();
```
bhtibrewal, line 435: Info: Firestore doesn't explicitly provide a method to check if a field exists so adding any kind of check to the field returns the docs containing that field

constants/roles.js
```diff
@@ -4,6 +4,7 @@ const ROLES = {
   APPOWNER: "app_owner",
   MEMBER: "member",
   ARCHIVED: "archived",
+  INDISCORD: "in_discord",
```
DashDeipayan, line 7: I guess it will be better to use a camel cased literal rather than a snake case

models/users.js
```diff
@@ -412,6 +414,34 @@ const getUsersBasedOnFilter = async (query) => {
     const filteredUserDocs = userDocs.filter((doc) => !doc.roles?.archived);
     return filteredUserDocs;
   }
+
+  const roleQuery = query.role;
```
DashDeipayan, line 418: We can use destructuring here. Shorter code 😃

models/users.js
```diff
@@ -360,6 +360,8 @@ const getRdsUserInfoByGitHubUsername = async (githubUsername) => {
  * @param {Array} query.levelNumber - Array of levelNumbers to filter the users on
  * @param {Array} query.tagId - Array of tagIds to filter the users on
  * @param {Array} query.state - Array of states to filter the users on
+ * @param {String} query.role - filter the users on
+ * @param {String} query.verified - filter the users on
```
DashDeipayan, line 364: What does this mean? I guess you forgot completing the sentence 😂

models/users.js
```diff
@@ -412,6 +414,34 @@ const getUsersBasedOnFilter = async (query) => {
     const filteredUserDocs = userDocs.filter((doc) => !doc.roles?.archived);
     return filteredUserDocs;
   }
+
+  const roleQuery = query.role;
+  const verifiedQuery = query.verified;
```
DashDeipayan, line 419: What does this verified check? If this is for checking if a particular user in out discord bot verified or not, we already do have a middleware for that. If not, then let's discuss.

test/fixtures/user/user.js
```diff
@@ -21,8 +21,10 @@ module.exports = () => {
       isMember: true,
       phone: "1234567890",
       email: "abc@gmail.com",
+      joined_discord: "2023-01-13T18:21:09.278000+00:00",
       roles: {
         member: true,
+        in_discord: true,
```
DashDeipayan, line 27: Camel casing 🙂

test/fixtures/user/user.js
```diff
@@ -105,13 +107,15 @@ module.exports = () => {
       github_display_name: "Ankush Dharkar",
       phone: "1234567890",
       email: "ad@amazon.com",
+      joined_discord: "2023-01-13T18:21:09.278000+00:00",
```
DashDeipayan, line 110: I guess this must also be camel cased

DashDeipayan (changes requested):

constants/roles.js
bhtibrewal: super_user and app_owner were already present so I had to follow the same rule here

controllers/users.js
```diff
@@ -24,7 +24,11 @@ const verifyUser = async (req, res) => {
...
   return res.json({
```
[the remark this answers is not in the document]
bhtibrewal: right, thanks for pointing out

models/users.js
bhtibrewal: yes , thought would complete it later

models/users.js
bhtibrewal: done

controllers/users.js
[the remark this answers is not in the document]
bhtibrewal: I haven't worked on this API so I think this issue shouldn't be addressed in this PR, it might break something I am not aware of

controllers/users.js
heyrandhir: why have we replaced our custom fetch with the native fetch.

models/users.js
```diff
@@ -412,6 +414,33 @@ const getUsersBasedOnFilter = async (query) => {
...
+    const filteredUsers = [];
+    const snapshot = await userModel.where("discordId", "!=", null).get();
+    snapshot.forEach((doc) => {
+      filteredUsers.push({
+        id: doc.id,
+        ...doc.data(),
+      });
+    });
+
+    return filteredUsers.filter((user) => !user.roles?.archived);
```
heyrandhir, line 442: what if `role` and `verified` both are passed in the query. 🤔

middlewares/validators/user.js
```diff
@@ -172,6 +172,8 @@ async function validateUserQueryParams(req, res, next) {
           joi.array().items(joi.string().valid("IDLE", "OOO", "ACTIVE"))
         )
         .optional(),
+      role: joi.string().valid("in_discord", "member").optional(),
```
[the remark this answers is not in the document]
DashDeipayan: Good catch randhir

models/users.js
DashDeipayan: According to the code, `role` has a higher precedence over `verified`.

controllers/users.js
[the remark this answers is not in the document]
heyrandhir: I believe @bhtibrewal concern is that if we modify this code, it could potentially impact the frontend code that consumes this API. Specifically, there might be code in the frontend that checks the status or relies on this specific implementation, and any changes made here could potentially impact or break the code there.

models/users.js
heyrandhir: or have we missed the case in which both are present.

middlewares/validators/user.js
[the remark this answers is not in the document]
bhtibrewal: if I make role to array here I will have to handle the case of multiple entries in the array, which will go against the API contract and the requirements of this PR , so I think it would be better to do that when we implement search based on multiple roles

models/users.js
bhtibrewal: I have not missed the case where both are passed; it just was not the requirement currently and it didn't make sense to increase the code complexity unnecessarily. I had discussed with Ankush that the current search function here is not scalable and requires a lot of if-else statements, so just implementing the requirements here.

I also considered the case here what if status and role both are passed.

middlewares/validators/user.js
```diff
@@ -172,6 +172,8 @@ async function validateUserQueryParams(req, res, next) {
...
+      verified: joi.string().optional(),
```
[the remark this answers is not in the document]
bhtibrewal: a min(1 ) is already present in the function on line 162

controllers/users.js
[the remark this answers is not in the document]
DashDeipayan: Okay, sounds good, but we need to take such things in account. No worries @bhtibrewal

DashDeipayan (approved):
LGTM

models/users.js
heyrandhir: > According to the code, `role` has a higher precedence over `verified`.

ok so as @DashDeipayan mentioned in that case then `role` gets a higher precedence over `verified`.

DashDeipayan: We can sit and discuss ways of improving it 🙂 if delivering this feature isn't urgent/we have some time

PR #1067, opened 2023-05-03
Feature: Search issues using GitHub Search API

bharati-21:
# Feature: Search Issues using GitHub search API 

This PR closes the following [issue to search for issues across RDS repositories](https://github.com/Real-Dev-Squad/todo-action-items/issues/144)

- Users can search for specific issues using search strings that are passed as request query parameter
- Depending on the value of the query parameter, the relevant API from GitHub is called.
  - If the query param is empty, then the issues are fetched from GitHub issues API
  - Else, the issues are fetched from the GitHub search API

- Existing service to fetch issues is leveraged with an additional parameter for search string.

### DIFF
--- a/test/unit/services/githubService.test.js
+++ b/test/unit/services/githubService.test.js
@@ -145,7 +145,7 @@ describe("githubService", function () {
         searchString: "website",
       });
       expect(response).to.be.equal(
-        "https://api.github.com/search/issues?q=website+org%3AReal-Dev-Squad+type%3Aissue+is%3Aopen+created%3A%3E%3D2023-01-01&sort=created&order=desc&per_page=100&page=1"
+        "https://api.github.com/search/issues?q=website+org%3AReal-Dev-Squad+type%3Aissue+is%3Aopen&sort=created&per_page=100&page=1"
       );
     });
   });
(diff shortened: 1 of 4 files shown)


test/unit/services/githubService.test.js
```diff
@@ -138,4 +138,15 @@ describe("githubService", function () {
...
     });
   });
+
+  describe("fetchOpenIssues with search query param", function () {
+    it("Should generate the correct url to fetch open issues with search param", async function () {
+      const response = await githubService.fetchOpenIssues({
+        searchString: "website",
+      });
+      expect(response).to.be.equal(
+        "https://api.github.com/search/issues?q=website+org%3AReal-Dev-Squad+type%3Aissue+is%3Aopen+created%3A%3E%3D2023-01-01&sort=created&order=desc&per_page=100&page=1"
```
RitikJaiswal75, line 148: can we use interpolation here and give the URL a proper structure so that it is easy to write tests for something similar in the future?

RitikJaiswal75 (dismissed):
LGTM

test/unit/services/githubService.test.js
bharati-21: Although this has been the convention which was already followed, we can use the GitHub URL and Org name from the config. Other's are not common to every test. It's common to this particular test.

bharati-21: Done.

bharati-21:
Failing tests have also been fixed.

RitikJaiswal75 (approved):
LGTM

PR #1071, opened 2023-05-06
Added JOI Validation for progresses api

heyrandhir:
Added input validation for the /progresses API endpoint using Joi.  The following fields are now validated:

• type - Must be "user" or "task" 
• completed - string (mandatory)
• planned - string (mandatory)
• blockers - string (mandatory allowed to be empty) 
• taskId - Required when type is "task"


heyrandhir:
Added this in a different PR

Issue #1076, opened 2023-05-11
Create progresses Collection Routes 


heyrandhir:
Create below three API endpoints related to tracking progress for a particular user or task. Here's a brief description of each:

POST /progresses: This endpoint will allow us to create a new progress entry.
GET /progresses: This endpoint will allow us to retrieve progress entries based on the provided parameters.
GET /progresses/range: This endpoint will allows us to retrieve progress records for a particular user or task within the specified date range.

It has following constraints: 
Progress record before 6 am of the next day should be counted as progress report for today.
Only 1 Progress record is allowed to be created by per user per day

heyrandhir:
The child tickets are as follows : 
- #1074 
- https://github.com/Real-Dev-Squad/website-data-models/pull/36
- https://github.com/Real-Dev-Squad/website-api-contracts/pull/124

heyrandhir:
TODO : 

Create a composite index as follows for the progresses collection
-

| Field | taskId | Ascending |
|-------|--------|-----------|
| Field | date   | Ascending |

| Field | userId | Ascending |
|-------|--------|-----------|
| Field | date   | Ascending |

PR #1088, opened 2023-05-15
Remove the test skips for progresses API (TDD approach)

heyrandhir:
This pull request follows a Test-Driven Development (TDD) approach to remove the test skips for the progresses API. Additionally, this PR introduces the use of the INTERNAL_SERVER_ERROR_MESSAGE constant. This constant allows for consistent error handling by providing appropriate 500 error responses with relevant messages, while avoiding the exposure of exact error details.

The Parent ticket can be found [here](https://github.com/Real-Dev-Squad/website-backend/issues/1076)

### DIFF
--- a/controllers/progresses.js
+++ b/controllers/progresses.js
@@ -1,6 +1,6 @@
 const { Conflict, NotFound } = require("http-errors");
 const { createProgressDocument, getProgressDocument, getRangeProgressData } = require("../models/progresses");
-const { RESPONSE_MESSAGES } = require("../constants/progresses");
+const { RESPONSE_MESSAGES, INTERNAL_SERVER_ERROR_MESSAGE } = require("../constants/progresses");
 const { PROGRESS_DOCUMENT_RETRIEVAL_SUCCEEDED, PROGRESS_DOCUMENT_CREATED_SUCCEEDED } = RESPONSE_MESSAGES;
 
 /**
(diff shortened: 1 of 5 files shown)


controllers/progresses.js
```diff
@@ -112,8 +113,9 @@ const getProgress = async (req, res) => {
         message: error.message,
       });
     }
-    return res.status(400).json({
-      message: error.message,
+    logger.error(error.message);
+    return res.status(500).json({
+      message: INTERNAL_SERVER_ERROR_MESSAGE,
     });
   }
```
Pratiyushkumar, line 120: why have we written  this twice
logger.error(error.message);
    return res.status(500).json({
      message: INTERNAL_SERVER_ERROR_MESSAGE,
    });
  }

Pratiyushkumar (commented):
rest everything LGTM

controllers/progresses.js
heyrandhir: logger will log it in our logs. while the `res.status(500).json({`will send the reponse.

sahsisunny (approved):

controllers/progresses.js
Pratiyushkumar: you this answer gives me the meaning of the logger,
my doubt was we have mentioned twice the  line in the code why was that needed

heyrandhir: I am unable to get you. which line has been mentioned twice. can you tell the line no's which you think are repeated.

heyrandhir: As discussed in the call @Pratiyushkumar  I am resolving this.

Issue #1073, opened 2023-05-09
Deliverables, ETAs, Milestones - Rate Limting Middleware | Parent Ticket | 


shmbajaj:
## Problem
- The Express.js backend API code is missing a rate-limiting feature, which can lead to server overload, slow response times, and poor user experience.
- The absence of rate-limiting can cause issues such as excessive requests, automated bots abuse, and security threats like brute-force attacks.
- The consequences of missing rate-limiting are numerous, including system vulnerability to DDoS attacks, poor API performance, and reduced user functionality and trust.

## Solution

- Our proposed solution is to implement a custom rate-limiting middleware using the "node-rate-limiter-flexible" package, which can be easily integrated with the existing Express.js backend.
- The middleware will enforce rate limits on all incoming requests to the API, preventing abuse and improving overall performance.
- We will use a Redis Hash database to store rate-limiting rules and deploy the application on the "railway.app" platform to ensure scalability and reliability.
- Different rate-limiting rules will be set for users with different roles, and IP address resolution will be implemented to resolve public IP and proxy IP issues. (and more)

`Info:`
- For More Check the Design Document, referenced below.

## Deliverables

| Feature Name | Feature ETA | Feature Status | Feature Description | Feature PR/ Discussion | Prod Ready |
|--------------|-------------|----------------|---------------------|------------|------------|
| Rate Limiting Rules: Research, Analysis, and Identifying Acceptable Rates | 14/05/2023 | IN_PROGRESS | This feature involves researching, identifying, and coming up with rate-limiting rules for common rate limiters and a few routes for testing and debugging while developing. _Already build solutions references will be taken for identifying how they came up with the solution._ | [Proposed: Using `newrelic` to come up with rate limit parameters and their values ](https://github.com/Real-Dev-Squad/website-backend/issues/1073#issuecomment-1546757318) | N | 
| Route-Based Rate Limiting: Common Rate Limiter Middleware | 18/05/2023 | IN_PROGRESS | This feature involves: setting up the `node-rate-limiter-flexible` package, using in-process memory store(temp), getting rate limiting rules from variables stored in process memory (temp). | [PR: Setting Up Rate Limitier](https://github.com/Real-Dev-Squad/website-backend/pull/1090) | N |
[...]

PR #1090, opened 2023-05-15
Feature/rlm common rate limiter

shmbajaj:
This pull request introduces rate-limiting middleware for the backend APIs. The rate-limiting middleware is implemented using the `node-rate-limiter-flexible` package. The code follows a Test-Driven Development (TDD) approach and includes unit tests using the sinon library.

Closes [Feature#2: Route-Based Rate Limiting: Common Rate Limiter Middleware](https://github.com/Real-Dev-Squad/website-backend/issues/1073)

## Before Change
- The code did not have rate-limiting middleware implemented.

## After Change
- Added rate-limiting middleware using the `node-rate-limiter-flexible` package.
- Implemented rate-limiting rules using temporary variables stored in process memory.
- Created unit tests for the rate-limiting middleware using `sinon`.

### More
- The rate-limiting middleware is temporarily using an in-process memory store for simplicity and speed. The plan is to later replace it with a Redis cache service.
- The code ensures that the next middleware is called when the request count is under the limit.
- If the request count exceeds the limit, the code returns a 429 status code with the message "Too many requests."
- The code resets the request count after a specific duration has passed and ensures that subsequent requests are not rate-limited.

## Checks Done
- [x] TDD Followed
- [x] Commits Checked Passed as Expected
- [x] Is Prod Ready? 

**NOTE:** 
The temporary variables used for rate-limiting rules will be replaced with a database in the future.

### DIFF
--- a/constants/rateLimiting.js
+++ b/constants/rateLimiting.js
@@ -1,0 +1,8 @@
+const TOO_MANY_REQUESTS = {
+  ERROR_TYPE: "Too Many Requests",
+  STATUS_CODE: 429,
+};
+
+module.exports = {
+  TOO_MANY_REQUESTS,
+};
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "newrelic": "^9.0.0",
     "passport": "^0.6.0",
     "passport-github2": "^0.1.12",
+    "rate-limiter-flexible": "^2.4.1",
     "winston": "^3.3.3"
   },
   "devDependencies": {
--- a/middlewares/rateLimiting.js
+++ b/middlewares/rateLimiting.js
@@ -1,0 +1,52 @@
+const { RateLimiterMemory } = require("rate-limiter-flexible");
+const { TOO_MANY_REQUESTS } = require("../constants/rateLimiting");
+const { getRetrySeconds } = require("../utils/rateLimiting");
+
+// INFO: temporarily added here, will be take from env-var/config
+const opts = {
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "newrelic": "^9.0.0",
     "passport": "^0.6.0",
     "passport-github2": "^0.1.12",
+    "rate-limiter-flexible": "^2.4.1",
```
nitiiink, line 37: try excluding the caret symbol in package json

shmbajaj: hey @nitiiink what's the need of excluding, do you want to manage updates manually?

nitiiink: hey @shmbajaj yes, usually there might be vulnerabilities associated with the codebase if we keep automatic updates on,hence suggested this.

prakashchoudhary07 (approved):

package.json
shmbajaj: Thanks @nitiiink, I will implement in the subsequent feature PR.

constants/rateLimiting.js
```diff
@@ -0,0 +1,8 @@
+const TOO_MANY_REQUESTS = {
+  ERROR_TYPE: "Too Many Requests",
+  STATUS_CODE: 429,
+};
+
+module.exports = {
+  TOO_MANY_REQUESTS,
+};
```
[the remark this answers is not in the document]
shmbajaj: Thanks for pointoing out, I will add these to other error messages in next feature PR.
As it's a `NIT` and small change.

PR #1103, opened 2023-05-20
Adds tests for tracked progress route

heyrandhir:
This PR adds tests for the Tracked Progress Collection. The tests cover various scenarios such as creating tracked progress documents, handling errors, retrieving progress documents, and asserting response data.

With this we also make the queries more specific in the util file by specifying the `type`.

Please find the parent ticket [here](https://github.com/Real-Dev-Squad/website-backend/issues/1076)

PR #1074, opened 2023-05-10
Add progress tracking API endpoints

heyrandhir:
These three API endpoints are related to tracking progress for a particular user or task. Here's a brief description of each:

1. `POST /progresses`: This endpoint allows us to create a new progress entry. 
2. `GET /progresses`: This endpoint allows you to retrieve progress entries based on the provided parameters. 
3. `GET /progresses/range`: This endpoint allows you to retrieve progress records for a particular user or task within the specified date range.

The Issue ticket can be found [here](https://github.com/Real-Dev-Squad/website-backend/issues/1076)
The Data Model for these APIs can be found  [here](https://github.com/Real-Dev-Squad/website-data-models/tree/main/progress)
The API Contract can be found [here](https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/progress)
The test for the API can be found [here](https://github.com/Real-Dev-Squad/website-backend/pull/1068)

PR #1111, opened 2023-05-23
Add GET , POST , PATCH method to the `/monitor` route for managing tracked progress.

heyrandhir:
This PR adds the endpoints for handling GET , POST , PATCH method to the `/monitor` route for managing tracked progress.

The Data model for the same can be found [here](https://github.com/Real-Dev-Squad/website-data-models/tree/main/tracked-progresses)

The Tests can be found [here](https://github.com/Real-Dev-Squad/website-backend/pull/1103)

The API Contract can be found [her](https://github.com/Real-Dev-Squad/website-api-contracts/pull/131)

The Parent ticket for this can be found [here](https://github.com/Real-Dev-Squad/website-backend/issues/1076)


utils/monitor.js
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Builds a Firestore query to check if a document exists based on the provided query parameters.
+ *
+ * @param {Object} queryParams - The query parameters for checking if a document exists.
+ * @param {string} queryParams.userId - The ID of the user (optional).
+ * @param {string} queryParams.taskId - The ID of the task (optional).
+ * @returns {Firestore.Query} - A Firestore query to check if a document exists.
+ */
+const buildQueryToCheckIfDocExists = (queryParams) => {
```
RitikJaiswal75, line 15: what does this function do? does it check for valid user and task id?

utils/monitor.js
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * @param {Object} queryParams - The query parameters for fetching a specific tracked progress document.
+ * @param {string} queryParams.userId - The userId of the tracked progress document
+ * @param {string} queryParams.taskId - The taskId of the tracked progress document
+ * @returns {Firestore.Query} - A Firestore query for fetching a specific tracked progress document.
+ */
+const buildQueryToFetchTrackedDoc = (queryParams) => {
+  const { userId, taskId } = queryParams;
+  if (userId) {
+    return trackedProgressesCollection.where("userId", "==", userId);
```
RitikJaiswal75, line 52: this is a direct query which should be in models again (My opinion)

middlewares/validators/monitor.js
```diff
@@ -0,0 +1,129 @@
...
+        "any.only": "'frequency' field must be equal to 1 for type 'user'",
+      }),
+    taskId: joi
+      .string()
+      .trim()
+      .when("type", {
+        is: "task",
+        then: joi.required().messages({
+          "any.required": "Required field 'taskId' is missing.",
+          "string.trim": "taskId must not have leading or trailing whitespace",
```
RitikJaiswal75, line 45: should we show this error? cant we trim trailing spaces in backend?

models/monitor.js
```diff
@@ -0,0 +1,101 @@
...
+  const updatedData = { type, [`${type}Id`]: typeId };
+  const query = buildQueryToCheckIfDocExists(updatedData);
+  const existingDocumentSnapshot = await query.get();
+  if (existingDocumentSnapshot.empty) {
+    throw new NotFound(RESOURCE_NOT_FOUND);
+  }
+  const doc = existingDocumentSnapshot.docs[0];
+  const docId = doc.id;
+  const docData = { ...req.body, updatedAt: new Date().toISOString() };
+  await trackedProgressesCollection.doc(docId).update(docData);
```
RitikJaiswal75, line 70: wont we lose some keys here?

utils/monitor.js
heyrandhir: it simply builds a Firestore query to fetch the documents based on taskId or userId. the description needs to be updated. Thanks for pointing this out will update it.

utils/monitor.js
heyrandhir: Since this generic query can be reused at multiple places it was suggested by few of the reviewers in few of my earlier PRs to abstract in a util function. https://github.com/Real-Dev-Squad/website-backend/pull/1074#discussion_r1190106864. Let me know your thoughts. In the current code as well few of the queries have been used at multiple places.

utils/monitor.js
```diff
@@ -0,0 +1,81 @@
...
+ * @param {string} queryParams.type - The type of tracked progress documents.
+ * @param {boolean} queryParams.monitored - The flag indicating if the documents are marked (optional).
+ * @returns {Firestore.Query} - A Firestore query for fetching tracked progress documents of a specific type.
+ */
+const buildQueryForFetchingDocsOfType = (queryParams) => {
+  const { type, monitored } = queryParams;
+  let query = trackedProgressesCollection.where("type", "==", type);
+  if (monitored !== undefined) {
+    query = query.where("monitored", "==", JSON.parse(monitored));
+  }
```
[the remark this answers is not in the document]
heyrandhir: Since this generic query can be reused at multiple places it was suggested by few of the reviewers in few of my earlier PRs to abstract in a util function. https://github.com/Real-Dev-Squad/website-backend/pull/1074#discussion_r1190106864. Let me know your thoughts. In the current code as well few of the queries have been used at multiple places.

middlewares/validators/monitor.js
heyrandhir: can be handled at backend.

constants/monitor.js
```diff
@@ -0,0 +1,15 @@
+const RESOURCE_CREATED_SUCCEEDED = "Resource created successfully.";
+const RESOURCE_UPDATED_SUCCEEDED = "Resource updated successfully.";
+const RESOURCE_RETRIEVAL_SUCCEEDED = "Resource retrieved successfully.";
```
Pratiyushkumar, line 3: why cannot we put the constant

heyrandhir: I didn't get you. Can you reword your doubt again.

controllers/monitor.js
```diff
@@ -0,0 +1,187 @@
...
+ * @param {TrackedProgressRequestBody} req.body - The Request body object.
+ * @param {Express.Response} res - The Express response object.
+ * @returns {Promise<void>} - A Promise that resolves when the response has been sent.
+ */
+
+const createTrackedProgressController = async (req, res) => {
+  try {
+    const data = await createTrackedProgressDocument({ ...req.body });
+    return res.status(201).json({
+      message: RESOURCE_CREATED_SUCCEEDED,
```
Pratiyushkumar, line 49: instead of createTrackedProgressController, won't it be good to use createMonitoredProgressController

Pratiyushkumar (changes requested):
code looks good to me but naming conventions don't match with routes to me. if possible please look into it

constants/monitor.js
bhtibrewal:
```suggestion
const RESOURCE : {
CREATED_SUCCEEDED: "Resource created successfully.";
UPDATED_SUCCEEDED : "Resource updated successfully.";
RETRIEVAL_SUCCEEDED : "Resource retrieved successfully.";
}
```

middlewares/validators/monitor.js
```diff
@@ -0,0 +1,120 @@
...
+      is: "user",
+      then: joi.required().messages({
+        "any.required": "Required field 'userId' is missing.",
+      }),
+      otherwise: joi.optional(),
+    }),
+  })
+  .messages({ "object.unknown": "Invalid field provided." });
+
+const validateCreateTrackedProgressRecords = async (req, res, next) => {
```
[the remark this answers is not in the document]
heyrandhir: Your'e right. Its used for single record update. will remove the extra s at the end.

constants/monitor.js
heyrandhir: @bhtibrewal @Pratiyushkumar if you see below I am already exporting as `RESPONSE_MESSAGES.RESOURCE_CREATED_SUCCEEDED` , `RESPONSE_MESSAGES. RESOURCE_UPDATED_SUCCEEDED`. Let me know if you still feel it needs to be changed.

controllers/monitor.js
heyrandhir: "Tracked" and "Monitored" essentially convey the same meaning in this context. Considering that our collection name is still `trackedProgresses`, I don't believe changing the name to `createMonitoredProgressController` would make a significant difference. Let me know if you feel otherwise.

constants/monitor.js
Pratiyushkumar: Randhir don't you think we need to put variable names in present tense instead oof past tense for eg. :-  instead of  RESOURCE_CREATED_SUCCEEDED, can't we write RESOURCE_CREATED_SUCCESSFULLY

Pratiyushkumar (commented):
Rest looks good to me, will wait for others comments as well

constants/monitor.js
bhtibrewal: Yes randhir I saw that export at the end, but I still think we don't need to write RESOURCE with every constant here so making a nested object like I suggested above, removes that repeating word.

bhtibrewal (approved):
LGTM

constants/monitor.js
heyrandhir: @bhtibrewal then in that case it will be `RESPONSE_MESSAGES.RESOURCE.CREATED_SUCCEEDED`. I think just for a simple message we are adding unnecessary 2 levels of nesting.

bhtibrewal (approved):

Pratiyushkumar (approved):
Looks good to me

PR #972, opened 2023-03-28
Feat/100ms setup

prerana1821:
## Description:
100ms setup with RDS backend for live screen share feat.

## More details:
Fixes #969

Issue #969, opened 2023-03-27
[Live-Site] Setup backend with 100ms SDK


prerana1821:
## Parent Story
https://github.com/Real-Dev-Squad/website-www/issues/343

## Overview
This PR provides API contracts for live site - https://github.com/Real-Dev-Squad/website-www/issues/343 and https://github.com/Real-Dev-Squad/website-www/issues/427.

PR #47, opened 2020-09-26
Upgraded config library

ankurnarkhede:
- Upgraded config library to v3.3.2
- Mapped config keys to the environment variables

PR #1113, opened 2023-05-24
Tests/100ms setup

prerana1821:
## Description:

Tests for 100ms setup with RDS backend for live screen share feat.

## Main PR
https://github.com/Real-Dev-Squad/website-backend/pull/972

## More details:
https://github.com/Real-Dev-Squad/website-backend/issues/969

### DIFF
--- a/services/EventAPIService.js
+++ b/services/EventAPIService.js
@@ -1,4 +1,5 @@
 const axios = require("axios");
+const { API_100MS_BASE_URL } = require("../constants/events");
 
 // A service class for all REST API operations
 class EventAPIService {
@@ -7,7 +8,7 @@ class EventAPIService {
   constructor(tokenService) {
     // Set Axios baseURL to 100ms API BaseURI
     this.#axiosInstance = axios.create({
-      baseURL: "https://api.100ms.live/v2",
+      baseURL: API_100MS_BASE_URL,
       timeout: 3 * 60000,
     });
     this.#tokenServiceInstance = tokenService;
@@ -16,6 +17,10 @@ class EventAPIService {
 
   // Add Axios interceptors to process all requests and responses
   #configureAxios() {
+    const shouldRetryWithRefreshedToken = (error, originalRequest) => {
+      return (error.response?.status === 403 || error.response?.status === 401) && !originalRequest._retry;
+    };
+
     this.#axiosInstance.interceptors.request.use(
       (config) => {
         // Add Authorization on every request made using the Management token
@@ -35,7 +40,7 @@ class EventAPIService {
       (error) => {
         logger.error("Error in making API call", { response: error.response?.data });
         const originalRequest = error.config;
-        if ((error.response?.status === 403 || error.response?.status === 401) && !originalRequest._retry) {
+        if (shouldRetryWithRefreshedToken(error, originalRequest)) {
           logger.info("Retrying request with refreshed token");
           originalRequest._retry = true;
 
--- a/services/EventAPIService.js
+++ b/services/EventAPIService.js
@@ -6,7 +6,6 @@ class EventAPIService {
   #tokenServiceInstance;
   constructor(tokenService) {
     // Set Axios baseURL to 100ms API BaseURI
-    logger.info({ axios });
     this.#axiosInstance = axios.create({
       baseURL: "https://api.100ms.live/v2",
       timeout: 3 * 60000,
@@ -63,7 +62,6 @@ class EventAPIService {
 
   // A method for POST requests using the configured Axios instance
   async post(path, payload) {
-    logger.info({ path, payload });
     const res = await this.#axiosInstance.post(path, payload || {});
     logger.info(`post call to path - ${path}, status code - ${res.status}`);
     return res.data;
--- a/services/EventAPIService.js
+++ b/services/EventAPIService.js
@@ -1,0 +1,73 @@
+const axios = require("axios");
+
+// A service class for all REST API operations
+class EventAPIService {
+  #axiosInstance;
+  #tokenServiceInstance;
+  constructor(tokenService) {
+    // Set Axios baseURL to 100ms API BaseURI
+    logger.info({ axios });
+    this.#axiosInstance = axios.create({
+      baseURL: "https://api.100ms.live/v2",
+      timeout: 3 * 60000,
+    });
+    this.#tokenServiceInstance = tokenService;
+    this.#configureAxios();
+  }
+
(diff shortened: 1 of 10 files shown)


services/EventTokenService.js
```diff
@@ -0,0 +1,68 @@
...
+class EventTokenService {
+  static #app_access_key = process.env.LIVE_APP_ACCESS_KEY || config.get("Event100ms.APP_ACCESS_KEY");
+  static #app_secret = process.env.LIVE_APP_SECRET || config.get("Event100ms.APP_SECRET");
+  #managementToken;
+  constructor() {
+    this.#managementToken = this.getManagementToken(true);
+  }
+
+  // A private method that uses JWT to sign the payload with APP_SECRET
+  #signPayloadToToken(payload) {
```
iamitprakash, line 15: in the line return !exp || exp + buffer < currTimeSeconds;. This line is checking to see if the token has expired. However, if the exp value is null or undefined, then the || operator will always evaluate to true, even if exp + buffer < currTimeSeconds is false. This means that the function will always return true, even if the token is not expired.

services/EventAPIService.js
```diff
@@ -0,0 +1,76 @@
...
+      },
+      (error) => {
+        logger.error("Error in making API call", { response: error.response?.data });
+        const originalRequest = error.config;
+        if (shouldRetryWithRefreshedToken(error, originalRequest)) {
+          logger.info("Retrying request with refreshed token");
+          originalRequest._retry = true;
+
+          // Force refresh Management token on error making API call
+          this.axios.defaults.headers.common.Authorization =
```
iamitprakash, line 48: in the line this.axios.defaults.headers.common.Authorization = "Bearer " + this.#tokenServiceInstance.getManagementToken(true);. This line is setting the Authorization header to the value of the refreshed token. However, the Authorization header is already set in the config object that is passed to the response interceptor. This means that the value of the Authorization header will be overwritten when the response interceptor is called.

iamitprakash (changes requested):

services/EventAPIService.js
```diff
@@ -0,0 +1,76 @@
...
+      baseURL: API_100MS_BASE_URL,
+      timeout: 3 * 60000,
+    });
+    this.#tokenServiceInstance = tokenService;
+    this.#configureAxios();
+  }
+
+  // Add Axios interceptors to process all requests and responses
+  #configureAxios() {
+    const shouldRetryWithRefreshedToken = (error, originalRequest) => {
```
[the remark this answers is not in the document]
prerana1821: Yes, added the check for WWW-Authenticate header in the shouldRetryWithRefreshedToken method. Thank you for your review.

services/EventTokenService.js
prerana1821: As discussed, we don't need to check for the exp while returning. Thank you for the comment. I have removed the check.

services/EventAPIService.js
prerana1821: As discussed, we don't need to again set the token since we have already set in the config interceptor. Thank you.

services/EventTokenService.js
```diff
@@ -0,0 +1,68 @@
...
+    const token = jwt.sign(payload, EventTokenService.#app_secret, {
+      algorithm: "HS256",
+      expiresIn: "24h",
+      jwtid: crypto.randomUUID({ disableEntropyCache: true }),
+    });
+    return token;
+  }
+
+  // A private method to check if a JWT token has expired or going to expire soon
+  #isTokenExpired(token) {
```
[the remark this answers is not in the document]
prerana1821: As discussed, we don't need to check for the exp while returning. Thank you for the comment. I have removed the check.

services/EventTokenService.js
```diff
@@ -0,0 +1,68 @@
...
+      const currTimeSeconds = Math.floor(Date.now() / 1000);
+      return !exp || exp + buffer < currTimeSeconds;
+    } catch (err) {
+      logger.info("error in decoding token", err);
+      return true;
+    }
+  }
+
+  // Generate new Management token, if expired or forced
+  getManagementToken(forceNew) {
```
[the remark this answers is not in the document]
prerana1821: As discussed, I have fixed the isTokenExpired method. Thank you for your comment.

services/EventTokenService.js
```diff
@@ -0,0 +1,68 @@
...
+        iat: Math.floor(Date.now() / 1000),
+        nbf: Math.floor(Date.now() / 1000),
+      };
+      this.#managementToken = this.#signPayloadToToken(payload);
+    }
+    return this.#managementToken;
+  }
+
+  // Generate new Auth token for a peer
+  getAuthToken({ roomId, userId, role }) {
```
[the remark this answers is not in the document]
prerana1821: Hey @iamitprakash,
As the token is been consumed at the same time, keeping an nbf in itself won't be fruitful because the token will be consumed by that time. As soon as the connection to 100ms is sent, we also add the management token. So, the token is authenticated as soon as it is generated. We can keep the time in nanoseconds but we are not sure about when the token will reach the server. 
If we keep the nbf 2-3 seconds later, users will face the latency which we don't want to happen or joining in the session after 2-3 seconds
If we keep the nbf the 100ms server won't be able to authenticate in case the request goes in & everytime the user has to create a new token. Everytime if the nbf is 2-3 seconds later then the user won't be able to authenticate to 100ms. (Basically, it would delay the token's validity and create unnecessary authentication delays for users.)

So, the usecase in which we keep the nbf to few seconds later won't be a valid case here.

services/EventTokenService.js
```diff
@@ -0,0 +1,66 @@
...
+  // Generate new Management token, if expired or forced
+  getManagementToken(forceNew) {
+    if (forceNew || this.#isTokenExpired(this.#managementToken)) {
+      const payload = {
+        access_key: EventTokenService.#app_access_key,
+        type: "management",
+        version: 2,
+        iat: Math.floor(Date.now() / 1000),
+      };
+      this.#managementToken = this.#signPayloadToToken(payload);
```
iamitprakash, line 46: line #47 even if the forceNew parameter is set to false and the current management token is not expired. This is because the #signPayloadToToken() function will always create a new token, regardless of the value of the forceNew parameter.

To fix this bug, you can change the following line of code:
if (forceNew) {
  this.#managementToken = this.#signPayloadToToken(payload);
}

prerana1821: Yes, I have fixed it. Thank you.

vvaibhavdesai (approved):

PR #1115, opened 2023-05-25
Added a route  `/scanQr/:user_id` to verify the status

### LINKED ISSUE #942: Feature Request : GET api call to get the status
Create a GET API call that will verify those tokens and give the status in the response whether it is verified or not.

Satyam005:
# Fixes Issue
Closes #942 

## Changes proposed
GET API call that will verify whether the user has scanned the QR or not by sending user_id in the body. If the user has scanned the QR then it will return the device info and the user Id, if not it will return the message that the user has not scanned the QR yet.

**_Response_** 

- If the user **has scanned qr**
image

- If the user has **not scanned qr**
image

### DIFF
--- a/controllers/scanQr.js
+++ b/controllers/scanQr.js
@@ -1,0 +1,24 @@
+const { INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
+const { fetchUserDeviceInfo } = require("../models/scanQr");
+
+/**
+ * Fetch the user device info data if user has scanned qr
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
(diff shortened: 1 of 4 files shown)


isVivek99:
i will be taking forward the api's, here is the api contract for reference [here](https://github.com/Real-Dev-Squad/website-api-contracts/pull/134/files)

isVivek99:
> i will be taking forward the api's, here is the api contract for reference [here](https://github.com/Real-Dev-Squad/website-api-contracts/pull/134/files)

@Satyam005 , thank you for working on the api's if you want to help please reach out in discord, or else reply here itself

PR #1122, opened 2023-05-27
changed dependsId to dependsOn

### LINKED ISSUE #1121: Changed dependsId to dependency

vinit717:
close #1121


sahsisunny (approved):

RitikJaiswal75 (dismissed):

models/tasks.js
RitikJaiswal75, line 1: What about the tests?

vinit717: The test is not breaking integration test is for checking req and response and response earlier it is dependsOn so it is not breaking

RitikJaiswal75 (approved):

heyrandhir (approved):

shubhamsigdar1:
You are also changing variable dependsId to dependency

shubhamsigdar1 (approved):

Issue #1025, opened 2023-04-12
Unit tests for user utils


ankushdharkar:
Why did we not change any corresponding tests? Are we missing unit tests?

_Originally posted by @ankushdharkar in https://github.com/Real-Dev-Squad/website-backend/pull/1022#discussion_r1164520785_

fakhruddinkw:
What is expected to be done here? @ankushdharkar

PR #1123, opened 2023-05-28
test: unit test for user util - getUsernamesFromPRs

khalatevarun:
Fixes - #1025 

Add unit test for user utils - `getUsernamesFromPRs()`


test/unit/utils/users.test.js
```diff
@@ -65,4 +65,145 @@ describe("users", function () {
...
+        {
+          title: "Added a link on logo ",
+          username: "kushvahasumit",
+          state: "closed",
+          createdAt: "2023-04-29T18:01:24Z",
+          updatedAt: "2023-05-28T17:58:53Z",
+          repository: "website-www",
+          url: "https://github.com/Real-Dev-Squad/website-www/pull/465",
+          labels: [],
+          assignees: [],
```
prakashchoudhary07, line 93: What do you think about moving the mock data into fixtures?

test/unit/utils/users.test.js
```diff
@@ -65,4 +66,31 @@ describe("users", function () {
...
     });
   });
+
+  describe("getUsernamesFromPRs", function () {
+    it("returns an array of unique usernames from the filtered PRs/Issues response", function () {
+      // Mock the allPRs data
+      const allPRs = filteredPRs;
+
+      // Execute the function and store the result
+      const usernames = usersUtils.getUsernamesFromPRs(allPRs);
```
shubhamsigdar1, line 76: getUsernamesFromPrs is a function that return all the usernames in an array?? Correct me if I am wrong??

shubhamsigdar1 (commented):
great work @khalatevarun

test/unit/utils/users.test.js
khalatevarun: yes

PR #1124, opened 2023-05-29
add tests for blocking archived users from discord actions

RitikJaiswal75:
Add tests for the feature to block archived users from using discord actions.

Test coverage

image

image


iamitprakash (changes requested):
in progress.js file #250

const progressDocumentData = { ...progressData, createdAt: createdAtTimestamp, date: progressDateTimestamp };. The problem is that the function is not checking if the date property is already set in the progressData object. If the date property is already set, then the function will overwrite the value with the current date. This could result in the progress document being created with an incorrect date.

RitikJaiswal75:
> in progress.js file https://github.com/Real-Dev-Squad/website-backend/pull/250

> const progressDocumentData = { ...progressData, createdAt: createdAtTimestamp, date: progressDateTimestamp };. The problem is that the function is not checking if the date property is already set in the progressData object. If the date property is already set, then the function will overwrite the value with the current date. This could result in the progress document being created with an incorrect date.

hey @iamitprakash I dont think this comment belongs to this pr.

prakashchoudhary07:
Do we also need to write test for, not blocking active users?

RitikJaiswal75:
> Do we also need to write test for, not blocking active users?

Good point let me add that.

test/integration/discord.test.js
```diff
@@ -0,0 +1,53 @@
...
+    chai
+      .request(app)
+      .post("/discord-actions/groups")
+      .set("Cookie", `${cookieName}=${jwt}`)
+      .send(roleData)
+      .end((err, res) => {
+        if (err) {
+          return done(err);
+        }
+        expect(res).to.have.status(403);
```
heyrandhir, line 35: Nit : we can also check the error message.

test/integration/discord.test.js
```diff
@@ -0,0 +1,53 @@
...
+  it("returns 403 for archived users get method", function (done) {
+    chai
+      .request(app)
+      .get("/discord-actions/groups")
+      .set("Cookie", `${cookieName}=${jwt}`)
+      .end((err, res) => {
+        if (err) {
+          return done(err);
+        }
+        expect(res).to.have.status(403);
```
heyrandhir, line 49: Nit : we can also check the error message.

heyrandhir (approved):

test/integration/discord.test.js
RitikJaiswal75: i dont think it is required as 403 is sent from the middleware

test/integration/discord.test.js
RitikJaiswal75: i dont think it is required as 403 is sent from the middleware

iamitprakash (approved):

heyrandhir (approved):

heyrandhir (commented):
Why so many changes are there in yarn.lock file 👀

RitikJaiswal75:
> Why so many changes are there in yarn.lock file eyes

These are integrity changes and the packages have been moved from npmjs to yarnpkgjs

heyrandhir (dismissed):

heyrandhir (approved):

prakashchoudhary07 (approved):

Issue #1085, opened 2023-05-14
Handle Assignee not found in PATCH `/tasks/:id` API


bhtibrewal:
### Bug
Currently, if we send an assignee username in the PATCH `/tasks/:id` API  payload and it doesn't find a user with the user name, it still gives the status 204

### Expectation
if no user is found with the username it should throw an error.

<img width="725" alt="image" src="https://github.com/Real-Dev-Squad/website-backend/assets/42600164/431b33a4-76b6-4d3d-a524-2fe1b51635fc">

khushi818:
Can you assign me this task?

khushi818:
I will send pr by tomorrow

PR #1125, opened 2023-05-29
fixed the bug for assignee patch request

khushi818:
Issue(#1085)

This code checks if a task has an assignee
then you search for it in user db
if the user does not exist you return a 404 saying user dosen't exist

image


RitikJaiswal75 (changes requested):
Please write tests for it

khushi818:
> Please write tests for it

It is almost done. will send the changes by tomorrow

controllers/tasks.js
```diff
@@ -177,7 +178,12 @@ const updateTask = async (req, res) => {
     if (!task.taskData) {
       return res.boom.notFound("Task not found");
     }
-
+    if (req.body.assignee) {
```
TanishqSingla, line 181: are we sure if we are getting the `assignee` field, if not I would recommend use optional chaining (i.e `?` operator) in the `if` block as it would avoid unexpected error.

TanishqSingla (approved):

vinit717 (approved):

shubhamsigdar1 (approved):
LGTM

RitikJaiswal75 (approved):
LGTM

heyrandhir (approved):

PR #1126, opened 2023-05-30
Update task status

prakashchoudhary07:
There are duplicate task status, removing those

ref: #1127


khalatevarun (approved):

test/integration/tasks.test.js
```diff
@@ -396,7 +396,7 @@ describe("Tasks", function () {
     it("Should return 404 when username is invalid", function (done) {
       chai
         .request(app)
-        .get("/tasks/dummyUser?status=active")
+        .get("/tasks/dummyUser?status=in_progress")
```
vikhyat187, line 399: one question @prakashchoudhary07, from where are we getting this small case `active` in the constants we have `IN_PROGRESS`

prakashchoudhary07: while passing it in the URL

PR #1132, opened 2023-06-01
Remove /members/moveToMembers/:username, and members/archiveMembers/:username and create a generic route 

### LINKED ISSUE #1131: creating a generic route for adding and removing roles from a user.
Currently, we have the following two routes for adding member roles to users, and achieving them:

- `members/moveToMembers/:username`: to add a member role to the users
- `members/archiveMembers/:username`: to add archive role to the users

## Problem with the approach used above
- If we want to remove those roles from the users, we need to create two more endpoints for removing those roles from the user ( basically making the value of the field false )

## Solution

- we create a generic route for adding and removing roles from a user:
   - `/users/:id/temporary/data`: it will be a patch request, in the payload we will send an object which contains an object with role as a key and a boolean value which will update the `role` according to that boolean value
   - For now this generic route is applicable for `member` and `archieved` role

shubhamsigdar1:
### What is the change?

- create generic route in user.js file and write validator, controller and model  

### Is Development Tested?

- [X] Yes
- [ ] No

API contract: https://github.com/Real-Dev-Squad/website-api-contracts/pull/132

### DIFF
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -1,4 +1,3 @@
-const { BadRequest, Unauthorized } = require("http-errors");
 const ROLES = require("../constants/roles");
 const members = require("../models/members");
 const tasks = require("../models/tasks");
@@ -100,40 +99,9 @@ const archiveMembers = async (req, res) => {
   }
 };
 
-const updateRoles = async (req, res) => {
-  try {
-    const userId = req.params.id;
-    const user = await fetchUser({ userId });
-    if (user?.userExists) {
-      const dataToUpdate = req.body;
-      const successObject = await members.updateRoles(user.user.id, dataToUpdate);
-      if (successObject.isRoleUpdated) {
-        return res.status(204).json({
-          message: "role updated successfully!",
-        });
-      }
-    }
-    return res.boom.notFound("User not found");
-  } catch (error) {
-    if (error instanceof BadRequest) {
-      return res.status(400).json({
-        message: "Invalid role",
-      });
-    } else if (error instanceof Unauthorized) {
-      return res.status(401).json({
-        message: "Unauthenticated User",
-      });
-    }
-    return res.status(503).json({
-      message: "Something went wrong please contact admin",
-    });
-  }
-};
-
 module.exports = {
   archiveMembers,
   getMembers,
   getIdleMembers,
   moveToMembers,
-  updateRoles,
 };
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -1,3 +1,4 @@
+const { NotFound, Unauthorized, ServiceUnavailable } = require("http-errors");
 const ROLES = require("../constants/roles");
 const members = require("../models/members");
 const tasks = require("../models/tasks");
@@ -62,6 +63,7 @@ const moveToMembers = async (req, res) => {
     const result = await fetchUser({ username });
     if (result.userExists) {
       const successObject = await members.moveToMembers(result.user.id);
+      // console.log(successObject);
       if (successObject.isAlreadyMember) {
         return res.boom.badRequest("User is already a member");
       }
@@ -99,9 +101,43 @@ const archiveMembers = async (req, res) => {
   }
 };
 
+const updateRoles = async (req, res) => {
+  try{
+    const userId = req.params.id;
+    const user = await fetchUser({ userId });
+    if (user?.userExists) {
+      const dataToUpdate = req.body;
+      const successObject = await members.updateRoles(user.user.id, dataToUpdate);
+      let statusCode;
+      if (successObject.isRoleUpdated) {
+        statusCode = 204;
+        responseObject.message = "role updated successfully!";
+      }
+      return res.status(statusCode).json(responseObject);
+    }
+  } catch (error) {
+    if (error instanceof NotFound) {
+      return res.status(404).json({
+        message: "User not found",
+      });
+    }else if(error instanceof Unauthorized) {
+      return res.status(401).json({
+        message: "Unauthenticated User",
+      });
+    }else if(error instanceof ServiceUnavailable) {
+      return res.status(503).json({
+        message: "Unauthenticated User",
+      });
+    }
+    logger.error(error.message);
+    return res.boom.badRequest("Invalid Request");
+  }
+}
+
 module.exports = {
   archiveMembers,
   getMembers,
   getIdleMembers,
   moveToMembers,
+  updateRoles,
 };
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -102,12 +102,13 @@ const archiveMembers = async (req, res) => {
 };
 
 const updateRoles = async (req, res) => {
-  try{
+  try {
     const userId = req.params.id;
     const user = await fetchUser({ userId });
     if (user?.userExists) {
       const dataToUpdate = req.body;
       const successObject = await members.updateRoles(user.user.id, dataToUpdate);
+      const responseObject = { message: "" };
       let statusCode;
       if (successObject.isRoleUpdated) {
         statusCode = 204;
@@ -115,24 +116,27 @@ const updateRoles = async (req, res) => {
       }
       return res.status(statusCode).json(responseObject);
     }
+    return res.boom.notFound("role updated failed.");
   } catch (error) {
     if (error instanceof NotFound) {
       return res.status(404).json({
         message: "User not found",
       });
-    }else if(error instanceof Unauthorized) {
+    } else if (error instanceof Unauthorized) {
       return res.status(401).json({
         message: "Unauthenticated User",
       });
-    }else if(error instanceof ServiceUnavailable) {
+    } else if (error instanceof ServiceUnavailable) {
       return res.status(503).json({
         message: "Unauthenticated User",
       });
     }
     logger.error(error.message);
-    return res.boom.badRequest("Invalid Request");
+    return res.status(400).json({
+      message: "Invalid role",
+    });
   }
-}
+};
 
 module.exports = {
   archiveMembers,
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -1,4 +1,4 @@
-const { NotFound, Unauthorized, ServiceUnavailable } = require("http-errors");
+const { BadRequest, Unauthorized } = require("http-errors");
 const ROLES = require("../constants/roles");
 const members = require("../models/members");
 const tasks = require("../models/tasks");
@@ -63,7 +63,6 @@ const moveToMembers = async (req, res) => {
     const result = await fetchUser({ username });
     if (result.userExists) {
       const successObject = await members.moveToMembers(result.user.id);
-      // console.log(successObject);
       if (successObject.isAlreadyMember) {
         return res.boom.badRequest("User is already a member");
       }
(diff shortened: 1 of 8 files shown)


controllers/members.js
```diff
@@ -62,6 +63,7 @@ const moveToMembers = async (req, res) => {
     const result = await fetchUser({ username });
     if (result.userExists) {
       const successObject = await members.moveToMembers(result.user.id);
+      // console.log(successObject);
```
vikhyat187, line 66: Please remove this if its commented.

controllers/members.js
```diff
@@ -99,9 +101,47 @@ const archiveMembers = async (req, res) => {
   }
 };
 
+const updateRoles = async (req, res) => {
+  try {
+    const userId = req.params.id;
+    const user = await fetchUser({ userId });
+    if (user?.userExists) {
+      const dataToUpdate = req.body;
+      const successObject = await members.updateRoles(user.user.id, dataToUpdate);
```
vikhyat187, line 110: I'm thinking why are we passing only userId and fetching the user again, can we pass the user object, so that we don't fetch it again.
Let others comment @RitikJaiswal75 @vinayak-trivedi

models/members.js
```diff
@@ -115,9 +115,40 @@ const addArchiveRoleToMembers = async (userId) => {
   }
 };
 
+const updateRoles = async (userId, newRoles) => {
+  try {
+    const userDoc = await userModel.doc(userId).get();
+    const user = userDoc.data();
```
vikhyat187, line 121: Fetching the user again here

models/members.js
```diff
@@ -115,9 +115,40 @@ const addArchiveRoleToMembers = async (userId) => {
...
+        }
+      } else if (newRoles.member === false) {
+        if (user?.roles?.member === false) {
+          return { isRoleUpdated: false };
+        }
+      }
+      roles = user.roles ? { ...user.roles, member: newRoles.member } : { member: newRoles.member };
+    } else if (Object.keys(newRoles).includes("archived")) {
+      if (user?.roles && user.roles[ROLES.ARCHIVED]) return { isRoleUpdated: false };
+      roles = { ...user.roles, archived: newRoles.archived };
```
vikhyat187, line 136: Here also I suggest checking if the role exists similar to the 
```suggestion
      roles = user.roles ? { ...user.roles, archived: newRoles.archived } : { archived: newRoles.archived };
```

vikhyat187 (commented):
Added a few minor comments, rest everything is good.
Good work @shubhamsigdar1

models/members.js
shubhamsigdar1: You asked question that why I fetch user data again?? or I have to fetch like this `const user = await fetchUser({ userId });`

models/members.js
shubhamsigdar1: But everytime when new user created in backend... `archived =false` is already present

controllers/members.js
shubhamsigdar1: okk I will ping once @vinayak-trivedi for this on discord

routes/users.js
```diff
@@ -21,6 +22,7 @@ router.get("/:userId/intro", authenticate, authorizeRoles([SUPERUSER]), users.ge
 router.put("/self/intro", authenticate, userValidator.validateJoinData, users.addUserIntro);
 router.get("/:id/skills", users.getUserSkills);
 router.get("/:id/badges", getUserBadges);
+router.patch("/:userId/roles", authenticate, authorizeRoles([SUPERUSER]), updateRoles);
```
heyrandhir, line 25: in the route to stay consistent with other routes we can use `id`in place of `userId`

heyrandhir (changes requested):
can you write the test for this first please. There are few errors which you will find once you write the tests

routes/users.js
shubhamsigdar1: If you see there is also used :userId in routes/user.js file -> https://github.com/Real-Dev-Squad/website-backend/blob/develop/routes/users.js#L20

models/members.js
shubhamsigdar1: Done

controllers/members.js
vinayak-trivedi: don't make it this generic please, add better names for this

vinayak-trivedi (changes requested):
@shubhamsigdar1 thanks for the PR adding few comments please check and let me know in case of any doubts

models/members.js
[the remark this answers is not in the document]
vikhyat187: +1

controllers/members.js
shubhamsigdar1: > don't make it this generic please, add better names for this

You are talking about change name `successObject` to better variable name

controllers/members.js
```diff
@@ -99,9 +100,40 @@ const archiveMembers = async (req, res) => {
...
+      if (successObject.isRoleUpdated) {
+        return res.status(204).json({
+          message: "role updated successfully!",
+        });
+      }
```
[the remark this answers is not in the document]
shubhamsigdar1: Done...Can you check once on latest changes(also have doubt)

models/members.js
```diff
@@ -115,9 +115,40 @@ const addArchiveRoleToMembers = async (userId) => {
...
+      } else if (newRoles.member === false) {
+        if (user?.roles?.member === false) {
+          return { isRoleUpdated: false };
+        }
+      }
+      roles = user.roles ? { ...user.roles, member: newRoles.member } : { member: newRoles.member };
+    } else if (Object.keys(newRoles).includes("archived")) {
+      if (user?.roles && user.roles.archived) return { isRoleUpdated: false };
+      roles = user.roles ? { ...user.roles, archived: newRoles.archived } : { archived: newRoles.archived };
+    }
```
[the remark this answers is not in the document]
shubhamsigdar1: I want to do discuss regarding this

models/members.js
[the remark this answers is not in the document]
shubhamsigdar1: Done

models/users.js
```diff
@@ -450,6 +450,36 @@ const getUsersBasedOnFilter = async (query) => {
...
+      if (newRoles.member === true) {
+        if (user?.roles?.member) {
+          return { isRoleUpdated: false };
+        }
+      } else if (newRoles.member === false) {
+        if (user?.roles?.member === false) {
+          return { isRoleUpdated: false };
+        }
+      }
+      roles = user.roles ? { ...user.roles, member: newRoles.member } : { member: newRoles.member };
```
heyrandhir, line 468: can we initialize `roles` in the begging itself with ` ...user.roles` so that here we just need to add an extra property

routes/users.js
```diff
@@ -21,6 +21,13 @@ router.get("/:userId/intro", authenticate, authorizeRoles([SUPERUSER]), users.ge
 router.put("/self/intro", authenticate, userValidator.validateJoinData, users.addUserIntro);
 router.get("/:id/skills", users.getUserSkills);
 router.get("/:id/badges", getUserBadges);
+router.patch(
+  "/:userId/roles",
```
heyrandhir, line 25: should we call the path param `:id` to stay consistent with the previous lines

routes/users.js
```diff
@@ -21,6 +21,13 @@ router.get("/:userId/intro", authenticate, authorizeRoles([SUPERUSER]), users.ge
...
+  authenticate,
+  userValidator.validateUpdateRoles,
+  authorizeRoles([SUPERUSER]),
```
heyrandhir, line 28: super user validator can be done before validateUpdateRoles so that we need not validate the body if the user is not a super user

test/integration/users.test.js
```diff
@@ -1071,4 +1076,156 @@ describe("Users", function () {
...
+            if (err) {
+              return done(err);
+            }
+
+            expect(res).to.have.status(400);
+            expect(res.body.message).to.be.equal("Invalid role");
+            return done();
+          });
+      });
+    });
```
heyrandhir, line 1166: in this case we should throw 409 conflict

test/integration/users.test.js
```diff
@@ -1071,4 +1076,156 @@ describe("Users", function () {
...
```
heyrandhir, line 1187: same here

heyrandhir (changes requested):

controllers/members.js
vikhyat187: > regarding your suggestion @vikhyat187, it will definitely save us few reads in the database, I think that this will not be a good to do, so let's take a hypothetical example, where in future we have two super users and we both of them make a requests to make the user a member at the same time, so ideally if the user is already a member we should return a response that user is already a member from backend, which will be possible only if we fetch the user details, I might not be clear with my description, please let me know if you want to discuss this

I think we can add optimistic locking, to only allow one update from a set of parallel updates, @iamitprakash please share your point of view to tackle this.

@vinayak-trivedi Let's connect today around 9:30PM IST

controllers/members.js
shubhamsigdar1: Done

routes/users.js
shubhamsigdar1: If you see there is also used :userId in routes/user.js file -> https://github.com/Real-Dev-Squad/website-backend/blob/develop/routes/users.js#L20

models/users.js
shubhamsigdar1: You mean I have to write like this:- 
`roles = {...userData.roles} ? { ...userData.roles, archived: newRoles.archived } : { archived: newRoles.archived };`

models/users.js
```diff
@@ -450,6 +450,36 @@ const getUsersBasedOnFilter = async (query) => {
...
+    if (Object.keys(newRoles).includes("member")) {
+      if (newRoles.member === true) {
+        if (user?.roles?.member) {
+          return { isRoleUpdated: false };
+        }
+      } else if (newRoles.member === false) {
+        if (user?.roles?.member === false) {
+          return { isRoleUpdated: false };
+        }
+      }
```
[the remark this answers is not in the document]
shubhamsigdar1: You means I have to write this way
```
if (newRoles.member === true || newRoles.member === false) {
        if (userData?.roles?.member === true || userData?.roles?.member === false) {
          return { isRoleUpdated: false };
        }
      }
```

controllers/members.js
shubhamsigdar1: Done

models/members.js
shubhamsigdar1: fixed

routes/users.js
shubhamsigdar1: Done

controllers/users.js
```diff
@@ -495,6 +495,26 @@ const filterUsers = async (req, res) => {
   }
 };
 
+const updateRoles = async (req, res) => {
+  try {
+    const user = await userQuery.fetchUser({ userId: req.params.userId });
+    if (user?.userExists) {
+      const dataToUpdate = req.body;
+      const successObject = await userQuery.updateRoles(user.user.id, dataToUpdate);
```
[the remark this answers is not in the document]
shubhamsigdar1: Done please check in latest commit

middlewares/validators/user.js
```diff
@@ -192,10 +192,26 @@ async function validateUserQueryParams(req, res, next) {
   }
 }
 
+async function validateUpdateRoles(req, res, next) {
+  const schema = joi.object().strict().keys({
+    member: joi.boolean(),
+    archived: joi.boolean(),
+  });
```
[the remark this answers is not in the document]
shubhamsigdar1: Done please check in latest changes

models/users.js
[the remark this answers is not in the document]
shubhamsigdar1: Did discussion with randhir he said not to change

models/users.js
```diff
@@ -450,6 +450,36 @@ const getUsersBasedOnFilter = async (query) => {
   return [];
 };
 
+const updateRoles = async (userId, newRoles) => {
+  try {
+    const userDoc = await userModel.doc(userId).get();
+    const user = userDoc.data();
```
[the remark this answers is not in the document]
shubhamsigdar1: Done

routes/users.js
shubhamsigdar1: Done please check in latest changes

test/integration/users.test.js
shubhamsigdar1: Done please check in latest changes

test/integration/users.test.js
shubhamsigdar1: Done please check in latest changes

vikhyat187:
@shubhamsigdar1 they are some new changes pushed in the repo please resolve the merge conflicts

shubhamsigdar1:
#1147

PR #1136, opened 2023-06-05
Fix bug with adding `assigneeRdsInfo` to tasks

heyrandhir:
Issue ticket : https://github.com/Real-Dev-Squad/website-status/issues/607

This PR fixes a bug where `assigneeRdsInfo` was not being added to the `tasks` object correctly. The issue was caused by asynchronous code, where tasks were being fetched using `tasks.fetchTasks()`, which returns a list of Promises. However, the code was not waiting for all Promises to resolve before adding `assigneeRdsInfo` to the tasks.

To fix this, I updated the code to use `Promise.all` to wait for all Promises to resolve before adding `assigneeRdsInfo` to the tasks.

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -51,7 +51,8 @@ const addNewTask = async (req, res) => {
  */
 const fetchTasks = async (req, res) => {
   try {
-    const allTasks = await tasks.fetchTasks();
+    const allTasksPromises = await tasks.fetchTasks();
+    const allTasks = await Promise.all(allTasksPromises);
     const fetchTasksWithRdsAssigneeInfo = allTasks.map(async (task) => {
       /*
        If the issue has a "github.issue" inner object and a property "assignee",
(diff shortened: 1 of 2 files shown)


heyrandhir:
**Before** 

image


**After** 

image

controllers/tasks.js
```diff
@@ -51,7 +51,8 @@ const addNewTask = async (req, res) => {
  */
 const fetchTasks = async (req, res) => {
   try {
-    const allTasks = await tasks.fetchTasks();
+    const allTasksPromises = await tasks.fetchTasks();
```
RitikJaiswal75, line 54: Didn't await make the code wait for fetch tasks?

heyrandhir: await is returning a list of promises.

bharati-21 (approved):
LGTM

prakashchoudhary07 (approved):

PR #1142, opened 2023-06-06
Return anrchived user only

sahsisunny:


Pratiyushkumar (approved):
LGTM

models/users.js
```diff
@@ -165,14 +165,16 @@ const fetchPaginatedUsers = async (query) => {
...
+      if (!doc.data().roles.archived) {
+        allUsers.push({
+          id: doc.id,
+          ...doc.data(),
+          phone: undefined,
+          email: undefined,
+          tokens: undefined,
+          chaincode: undefined,
+        });
+      }
```
heyrandhir, line 177: since `doc.data()` is used at multiple places we can store it in a variable and use it at both places.

heyrandhir (changes requested):
tests please !

heyrandhir (changes requested):

models/users.js
```diff
@@ -165,14 +165,17 @@ const fetchPaginatedUsers = async (query) => {
...
-        phone: undefined,
-        email: undefined,
-        tokens: undefined,
-        chaincode: undefined,
-      });
+      const userData = doc.data();
+      if (!userData.roles.archived) {
+        allUsers.push({
+          id: doc.id,
+          ...doc.data(),
```
RitikJaiswal75, line 172: this won't work db query already has 100 users some of them might be archived, lets say 20, now we are just returning 80 users, expectation was to return 100 unarchived users

RitikJaiswal75 (changes requested):
Where is the test for this feature?

models/users.js
sahsisunny: done

models/users.js
```diff
@@ -165,14 +165,17 @@ const fetchPaginatedUsers = async (query) => {
...
-        phone: undefined,
-        email: undefined,
-        tokens: undefined,
-        chaincode: undefined,
-      });
+      const userData = doc.data()
+      if (!userData.roles.archived) {
+        allUsers.push({
+          id: doc.id,
+          ...doc.data(),
```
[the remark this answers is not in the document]
sahsisunny: done

sahsisunny:
> Where is the test for this feature?

added

TanishqSingla (approved):

models/users.js
```diff
@@ -143,6 +143,8 @@ const fetchPaginatedUsers = async (query) => {
     const size = parseInt(query.size) || 100;
     const doc = (query.next || query.prev) && (await userModel.doc(query.next || query.prev).get());
     let dbQuery = (query.prev ? userModel.limitToLast(size) : userModel.limit(size)).orderBy("username");
+    dbQuery = dbQuery.where("roles.archived", "==", false);
```
RitikJaiswal75, line 146: still, the same issue that I mentioned first you will get 100 users then you will be filtering from them so it would return 100-n users but we expect 100 users not 100-n

RitikJaiswal75 (changes requested):

models/users.js
sahsisunny: done

RitikJaiswal75 (approved):
LGTM

Pratiyushkumar (approved):
LGTM

khushi818 (approved):
LGTM

heyrandhir (approved):
LGTM

PR #1145, opened 2023-06-07
Add a service to get members from discord

RitikJaiswal75:
I'm excited to inform you that I've developed a new method to obtain details regarding the individuals in our Discord community. This feature has proven incredibly useful in monitoring activity levels and identifying those less engaged. If there's anything specific you'd like me to investigate, please don't hesitate to ask!

We are not writing tests for it as it is just a fetch call and fetch calls are already tested.

we are not throwing an error in service as it will be present in the file we will consume the service.


services/discordService.js
```diff
@@ -0,0 +1,30 @@
...
+    const response = await fetch(`${DISCORD_BASE_URL}/discord-members`, {
+      method: "GET",
+      headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+    })
+      .then((response, err) => {
+        return response.json();
+      })
+      .then((data) => {
+        return data;
+      });
```
heyrandhir, line 21: can you please confirm if this `.then` is needed. as it is just returning the data which it is receiving. Removing this should still make the code work.

services/discordService.js
```diff
@@ -0,0 +1,30 @@
...
+    return response;
+  } catch (err) {
+    return [{ message: "Oops an internal error occured" }];
```
heyrandhir, line 24: why are we returning error message in a array 🤔

heyrandhir (changes requested):

services/discordService.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  try {
+    const response = await fetch(`${DISCORD_BASE_URL}/discord-members`, {
+      method: "GET",
+      headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+    })
+      .then((response, err) => {
+        return response.json();
+      })
```
heyrandhir, line 18: Nit : we can use async/await instead of promise chaining

heyrandhir (changes requested):

services/discordService.js
RitikJaiswal75: yes i tried it normally but id did not work as our cfw sends nested response, which wasn't accessible normally.

Also, could you provide reference where it sayswe should not use `.then`?

services/discordService.js
RitikJaiswal75: to maintain consistency in data transfer object

services/discordService.js
RitikJaiswal75: tried, but wasn't working

services/discordService.js
heyrandhir: Certainly! While there are no strict rules about using `async` and `await` in place of promise chaining,  using async await makes code easier to understand. If you see this mdn doc [here](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Promises) it says 
```note
The async and await keywords make it easier to build an operation from a series of consecutive asynchronous function calls, avoiding the need to create explicit promise chains, and allowing you to write code that looks just like synchronous code.
```

heyrandhir (approved):

services/discordService.js
```diff
@@ -0,0 +1,30 @@
...
+const DISCORD_BASE_URL = config.get("services.discordBot.baseUrl");
+
+const getDiscordMembers = async () => {
+  const authToken = jwt.sign({}, config.get("rdsServerlessBot.rdsServerLessPrivateKey"), {
+    algorithm: "RS256",
+    expiresIn: config.get("rdsServerlessBot.ttl"),
+  });
+
+  try {
+    const response = await fetch(`${DISCORD_BASE_URL}/discord-members`, {
```
DashDeipayan, line 12: Why was await and then used together, use either await(preferred) or then

services/discordService.js
DashDeipayan, line 24: We must log the error

PR #1152, opened 2023-06-10
add an endpoint to sync discord users

RitikJaiswal75:
Consume a pre-created function and create an endpoint to sync discord users and add their joined at date.

### Coverage report:
lines for controller written by me are covered

image

### Linked tickets:
- #1155

### DIFF
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -49,6 +49,7 @@ const syncExternalAccountData = async (req, res) => {
     const discordUserData = await getDiscordMembers();
     const rdsUserData = await getDiscordUsers();
     const rdsUserDataMap = {};
+    const userDataUpdatorPromises = [];
 
     rdsUserData.forEach((rdsUser) => {
       rdsUserDataMap[rdsUser.discordId] = {
@@ -57,7 +58,7 @@ const syncExternalAccountData = async (req, res) => {
       };
     });
 
-    discordUserData.forEach(async (discordUser) => {
+    discordUserData.forEach((discordUser) => {
       const mappedRdsUser = rdsUserDataMap[discordUser.user.id];
       if (mappedRdsUser) {
         const userData = {
@@ -67,16 +68,19 @@ const syncExternalAccountData = async (req, res) => {
             in_discord: true,
           },
         };
-        await addOrUpdate(userData, mappedRdsUser.id);
+        userDataUpdatorPromises.push(addOrUpdate(userData, mappedRdsUser.id));
       }
     });
 
+    await Promise.all(userDataUpdatorPromises);
+
     return res.json({
       rdsUsers: rdsUserData.length,
+      discordUsers: discordUserData.length,
       message: "Data Sync Complete",
     });
   } catch (err) {
-    logger.error("Error in syncind users discord joined at", err);
+    logger.error("Error in syncing users discord joined at", err);
     return res.status(500).json({ message: INTERNAL_SERVER_ERROR });
   }
 };
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -1,5 +1,5 @@
 const externalAccountsModel = require("../models/external-accounts");
-const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
+const { SOMETHING_WENT_WRONG, INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
 const { getDiscordMembers } = require("../services/discordService");
 const { getDiscordUsers, addOrUpdate } = require("../models/users");
 const logger = require("../utils/logger");
@@ -54,7 +54,8 @@ const syncExternalAccountData = async (req, res) => {
           const userData = {
             discordJoinedAt: discordUser.joined_at,
             roles: {
-              inDiscord: true,
+              ...rdsUser.roles,
+              in_discord: true,
             },
           };
           await addOrUpdate(userData, rdsUser.id);
@@ -63,12 +64,11 @@ const syncExternalAccountData = async (req, res) => {
     });
     return res.json({
       rdsUsers: rdsUserData.length,
-      discordUsers: discordUserData.length,
-      message: "worked with above count",
+      message: "Data Sync Complete",
     });
   } catch (err) {
     logger.error("Error in syncind users discord joined at", err);
-    return res.json({ message: "Some Internal error occured" });
+    return res.status(500).json({ message: INTERNAL_SERVER_ERROR });
   }
 };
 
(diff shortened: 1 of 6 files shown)


controllers/external-accounts.js
```diff
@@ -41,4 +44,32 @@ const getExternalAccountData = async (req, res) => {
...
+          const userData = {
+            discordJoinedAt: discordUser.joined_at,
+            roles: {
+              ...rdsUser.roles,
+              in_discord: true,
+            },
+          };
+          await addOrUpdate(userData, rdsUser.id);
+        }
+      });
```
yesyash, line 63: - Rather than have two for each loop a cleaner way could have been we filter out the users from `rdsUserdata` if they exist in `discordUserData` which would give us the required users to then run the `addOrUpdate` operation on, what do you think?

controllers/external-accounts.js
```diff
@@ -41,4 +44,32 @@ const getExternalAccountData = async (req, res) => {
   }
 };
 
-module.exports = { addExternalAccountData, getExternalAccountData };
+const syncExternalAccountData = async (req, res) => {
+  try {
+    const discordUserData = await getDiscordMembers();
```
yesyash, line 49: - can we not just name it `discordMembersData` to keep it consistent with the function name?

services/discordService.js
```diff
@@ -7,22 +7,17 @@ const getDiscordMembers = async () => {
     algorithm: "RS256",
     expiresIn: config.get("rdsServerlessBot.ttl"),
```
yesyash, line 8: - NIT PICK: Moving the places we do `config.get(<config>)` into its own variable would make this code look much cleaner. 
    - This is optional not a P0.

yesyash (changes requested):

controllers/external-accounts.js
RitikJaiswal75: The data is already filtered, so basically the outer loop has less data than the inner loop, the rdsUserData is the users who have run the `/verify` command. So, I don't feel the need to filter further.

controllers/external-accounts.js
RitikJaiswal75: Members will give us a sense of people with `members` roles soi used this as the variable name.

controllers/external-accounts.js
RitikJaiswal75: Update the code using 2 loops and a map reducing the complexity from o(n^2) to o(m+n).
In the process, I have created a map that uses extra memory. Are we ok with the tradeoff?

services/discordService.js
RitikJaiswal75: Will cover this in the coming PR as that service was written earlier, to modify it I would raise a new PR. (Focusing on the delta minimization of the current PR).

controllers/external-accounts.js
```diff
@@ -41,4 +44,50 @@ const getExternalAccountData = async (req, res) => {
...
-module.exports = { addExternalAccountData, getExternalAccountData };
+/**
+ * Gets all group-roles
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const syncExternalAccountData = async (req, res) => {
+  try {
+    const discordUserData = await getDiscordMembers();
+    const rdsUserData = await getDiscordUsers();
```
DashDeipayan: NIT: Make it 54,55 in one line if possible

test/integration/external-accounts.test.js
```diff
@@ -220,4 +227,63 @@ describe("External Accounts", function () {
         });
     });
   });
+
+  describe("PATCH /external-accounts/discord-sync", function () {
+    let jwtToken, fetchStub;
+
+    beforeEach(async function () {
+      const userId = await addUser(userData[4]);
+      jwtToken = authService.generateAuthToken({ userId });
```
DashDeipayan, line 236: Change token name to some variable name specific to SUPERUSER

DashDeipayan (changes requested):
added comments

controllers/external-accounts.js
RitikJaiswal75: done

test/integration/external-accounts.test.js
RitikJaiswal75: done

DashDeipayan (approved):
LGTM

services/discordService.js
```diff
@@ -8,21 +8,13 @@ const getDiscordMembers = async () => {
...
   });
 
-  try {
-    const response = await fetch(`${DISCORD_BASE_URL}/discord-members`, {
+  const response = await (
```
yesyash, line 11: - Is the error case being handled by the consumer of this service?

yesyash (changes requested):

controllers/external-accounts.js
```diff
@@ -41,4 +44,49 @@ const getExternalAccountData = async (req, res) => {
...
+/**
+ * Gets all group-roles
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const syncExternalAccountData = async (req, res) => {
+  try {
+    const [discordUserData, rdsUserData] = await Promise.all([getDiscordMembers(), getDiscordUsers()]);
+    const rdsUserDataMap = {};
+    const userDataUpdatorPromises = [];
```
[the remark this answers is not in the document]
RitikJaiswal75: I am not sure, both convey the same meaning.

services/discordService.js
RitikJaiswal75: yes, the consumer is the controller named syncExternalAccounts which has a try-catch block to handle errors.

RitikJaiswal75:
> Please create a ticket and link that here.

Hey, @yesyash here is the ticket.

https://github.com/Real-Dev-Squad/website-backend/issues/1155

controllers/external-accounts.js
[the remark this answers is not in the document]
yesyash: - Is the word `updator` the right one to use here? Is it a word at first?

yesyash:
> > Please create a ticket and link that here.
> 
> Hey, @yesyash here is the ticket.
> 
> #1155

Thanks, @RitikJaiswal75 added it to the PR description.

controllers/external-accounts.js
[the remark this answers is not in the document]
RitikJaiswal75: oops its a typo, i can get it fixed it should be `updater`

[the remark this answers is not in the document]
RitikJaiswal75: Are there any changes apart from that that you would like to mention? So that they can go in a single commit rather than having multiple commits. @yesyash

vvaibhavdesai (approved):

PR #1160, opened 2023-06-14
add a function to fetch all users from db and write tests

RitikJaiswal75:
This PR is a sub-part of a bigger feature
The feature overview: 
we need to add an in_discord role with the value false to each user in RDS which in turn is a sub-part of the discord syncing API. In the mentioned API we plan to add the discord joined date to every user in the server and change their in_discord role to true.

Test Coverage: 
Lines written by me: 486-491
image

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -483,6 +483,13 @@ const getDiscordUsers = async () => {
   }
 };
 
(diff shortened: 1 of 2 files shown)


vinayak-trivedi (approved):
LGTM

PR #1161, opened 2023-06-14
add a service to set in discord role as false for all users

RitikJaiswal75:
This PR is part of a big PR to create an endpoint to add an in_discord role as false for all users. This PR just creates and tests the service we are still pending on the route creation.

Test Coverage:
Lines written by me: 34-47
image

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -486,7 +486,7 @@ const getDiscordUsers = async () => {
 const fetchAllUsers = async () => {
   const users = [];
   const usersQuerySnapshot = await userModel.get();
-  usersQuerySnapshot.forEach((user) => users.push(user.data()));
+  usersQuerySnapshot.forEach((user) => users.push({ ...user.data(), id: user.id }));
   return users;
 };
 
--- a/services/discordService.js
+++ b/services/discordService.js
@@ -1,4 +1,5 @@
 const jwt = require("jsonwebtoken");
+const { fetchAllUsers, addOrUpdate } = require("../models/users");
 
 const DISCORD_BASE_URL = config.get("services.discordBot.baseUrl");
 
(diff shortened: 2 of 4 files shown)


DashDeipayan (dismissed):
LGTM

DashDeipayan (approved):
approved

vvaibhavdesai (approved):

models/users.js
```diff
@@ -486,7 +486,7 @@ const getDiscordUsers = async () => {
 const fetchAllUsers = async () => {
   const users = [];
   const usersQuerySnapshot = await userModel.get();
-  usersQuerySnapshot.forEach((user) => users.push(user.data()));
+  usersQuerySnapshot.forEach((user) => users.push({ ...user.data(), id: user.id }));
```
iamitprakash, line 489: don't you think  id property is pushed into the users array twice.

what about below code 
usersQuerySnapshot.forEach((user) => users.push({ ...user.data() }));
return users;

services/discordService.js
```diff
@@ -30,6 +33,26 @@ const getDiscordMembers = async () => {
   }
 };
 
+const setInDiscordFalseScript = async () => {
+  const users = await fetchAllUsers();
+  const updateUsersPromises = [];
+  users.forEach((user) => {
+    const id = user.id;
```
iamitprakash, line 40: delete user[id] statement deletes the id property from the user object. This means that the id property will no longer be present in the userData object that is passed to the userModel.doc(id).update(userData) method. This could allow an attacker to inject malicious code into the database by creating a user object with an id property that contains malicious code.

can't reframe like below
const setInDiscordFalseScript = async () => {
  const users = await fetchAllUsers();
  const updateUsersPromises = [];
  users.forEach((user) => {
    const userData = Object.assign({}, user, {
      roles: {
        ...user.roles,
        in_discord: false,
      },
    });
    updateUsersPromises.push(userModel.doc(user.id).update(userData));
  });
  await Promise.all(updateUsersPromises);
};

PR #1162, opened 2023-06-14
Feat/set in discord false controller

RitikJaiswal75:
Create a controller to consume previously created services and set in-discord of a user to false

After a discussion with ankush over the standup call on 15 June 2023 at 6:00 am IST we can skip 5xx tests for this PR.


routes/users.js
```diff
@@ -9,6 +9,7 @@
 const { getUserBadges } = require("../controllers/badges");
 const checkIsVerifiedDiscord = require("../middlewares/verifydiscord");
 
+router.post("/update-in-discord", authenticate, authorizeRoles([SUPERUSER]), users.setInDiscordScript);
```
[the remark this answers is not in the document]
harshith-venkatesh: Could this be looked into please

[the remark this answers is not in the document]
RitikJaiswal75: We do not have a rate limiter yet, and the API is limited to only super user so I am not sure about how to solve this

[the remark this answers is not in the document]
RitikJaiswal75: https://github.com/Real-Dev-Squad/website-backend/issues/1073

harshith-venkatesh (approved):
Approving it based on the discussion over the call where @RitikJaiswal75 confirmed 200 or 500 responses for the `setInDiscordScript` .

controllers/users.js
```diff
@@ -558,6 +559,15 @@ const nonVerifiedDiscordUsers = async (req, res) => {
   return res.json(data);
 };
 
+const setInDiscordScript = async (req, res) => {
+  try {
+    await setInDiscordFalseScript();
```
yesyash, line 564: - `setInDiscordFalseScript` is this a script?

controllers/users.js
```diff
@@ -558,6 +559,15 @@ const nonVerifiedDiscordUsers = async (req, res) => {
...
+    return res.json({ message: "Successfully added the in_discord field to false for all users" });
```
yesyash, line 565: - Should we store this in a constant, than hardcoding the string?

test/integration/users.test.js
```diff
@@ -1161,4 +1161,21 @@ describe("Users", function () {
         });
     });
   });
+
+  describe("POST /update-in-discord", function () {
+    it("it returns proper response", function (done) {
```
yesyash, line 1166: - What does "proper response" mean?

yesyash:
> Create a controller to consume previously created services and set in-discord of a user to false
> 
> After a discussion with ankush over the standup call on 15 June 2023 at 6:00 am IST we can skip 5xx tests for this PR.

What is the reason to skip 5xx for tests? 
Please be more descriptive this helps other reviewers.

controllers/users.js
RitikJaiswal75: I do not feel so, because this API is just a 1-time thing, creating a constant didn't make sense.

RitikJaiswal75:
> What is the reason to skip 5xx for tests?
Please be more descriptive this helps other reviewers.

The reason to skip 5xx tests is that 5xx only occurs when firebase fails to work, and I was not able to emulate that thus I asked Ankush and he told I cand send the PR and get it merged and send the pr for 5xx errors in future.

test/integration/users.test.js
RitikJaiswal75: It means a not failing, or the mentioned response. 
For a one-time script do we need to focus on the test title as well?

controllers/users.js
RitikJaiswal75: it is a function written in a services file which works as a script, it gets all the users in our db and sets a role of in_discord false to them.

vinayak-trivedi (approved):
LGTM

Issue #1119, opened 2023-05-26
API to change users' discord nickname


RitikJaiswal75:
# Create an API that can change the users' nicknames in the Real Dev Squad server.

**Description**
Create an API under the `/discord-actions` API which will call upon the endpoint exposed in cloudflare workers named: `/guild/members`

### API contracts for the `/guild/members`
- method: patch
- authentication: JWT signed with rds-serverless-private-key
- request body
```js
{
    discordId: <Discord id is available in the user's data once they link their discord id>, (Remember `I` is capital in Id )
    userName: <rds username of the selected user also available in users' database>, (Remember `N` is capital in username)
}
```

### Overview of the feature

**Use case 1**
- users join in Real Dev Squad Server
- Run the command `/verify`
- come to my site
- Accept terms and conditions
- click on authorize
- this will add a discord id in the users collection for the specific user
- now once the discord id is added the name change API must be called automatically and change the nickname of the user in the server

**Use case 2**
- users mark themselves OOO on my site
- the API needs to be called automatically
- set the nickname based on the standards defined (Will be defined in the near future)

### Expectations

- Follow TDD
- Keep a minimum test coverage of 80%
- Make sure API can not be called by users directly (Apart from superuser)

### Need help? Try looking here;
- Stuck with failed authentication/bad request signature response?
  - Look here how the auth token is set: https://github.com/Real-Dev-Squad/website-backend/blob/dead12da7ad289dd89dbf0b9105b94b1e19abb44/controllers/discordactions.js#L36
  
- unable to make a call to the cloudflare workers?
  - Look here how the fetch call is made: https://github.com/Real-Dev-Squad/website-backend/blob/dead12da7ad289dd89dbf0b9105b94b1e19abb44/controllers/discordactions.js#L110

khushi818:
I will be taking maximum two weeks time 
1 week - to do the task 
1 week - for test
I will try to complete it before that.

khushi818:
``` /** 
  * Middleware to check if the user has verified themself with the discord bot. 
  *  If user has not verified, then no actions based on discord bot should be allowed. 
  * Note: This requires that user is authenticated hence must be called after 
  * the user authentication middleware. We are calling it from within the 
  * `authenticate` middleware itself to avoid explicitly adding this middleware 
  * while defining routes. 
  * 
  * @param {Object} req - Express request object 
  * @param {Object} res - Express response object 
  * @param {Function} next - Express middleware function 
[...]

PR #1169, opened 2023-06-15
feat/change-nickname

khushi818:
Issue(#1119 )

### DIFF
--- a/controllers/discordactions.js
+++ b/controllers/discordactions.js
@@ -3,7 +3,7 @@ const admin = require("firebase-admin");
 const config = require("config");
 const jwt = require("jsonwebtoken");
 const discordRolesModel = require("../models/discordactions");
-
+const userQuery = require("../models/users");
 /**
  * Creates a role
  *
@@ -122,8 +122,45 @@ const addGroupRoleToMember = async (req, res) => {
   }
 };
 
+/**
+ * Patch Update user nickname
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const changeNicknameOfUsers = async (req, res) => {
+        try{
+          const userName = req.body.userName
+          const result = await userQuery.fetchUser({ username: userName });
+          
+          const dataForDiscord = {
+            userName : `${userName}-ooo`,
+            discordId : '670589026616475658'
+          }
+          
+          const authToken = await jwt.sign({}, config.get("rdsServerlessBot.rdsServerLessPrivateKey"), {
+            algorithm: "RS256",
+            expiresIn: config.get("rdsServerlessBot.ttl"),
+          });
+
+          await fetch(`${DISCORD_BASE_URL}/guild/member`, {
+            method: "PATCH",
+            body: JSON.stringify(dataForDiscord),
+            headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+          }).then((response) => response.json());
+                
+          return res.status(200).json({
+            message : "nickname has been changed"
+          })
+        } 
+        catch(err){
+          logger.error(`Error while updating nickname: ${err}`)
+          return res.boom.badImplementation(INTERNAL_SERVER_ERROR)
+        }    
+}
+
 module.exports = {
   createGroupRole,
   getAllGroupRoles,
   addGroupRoleToMember,
+  changeNicknameOfUsers
 };
--- a/routes/discordactions.js
+++ b/routes/discordactions.js
@@ -1,7 +1,7 @@
 const express = require("express");
 const authenticate = require("../middlewares/authenticate");
-const { createGroupRole, getAllGroupRoles, addGroupRoleToMember } = require("../controllers/discordactions");
-const { validateGroupRoleBody, validateMemberRoleBody } = require("../middlewares/validators/discordactions");
+const { createGroupRole, getAllGroupRoles, addGroupRoleToMember, changeNicknameOfUsers } = require("../controllers/discordactions");
+const { validateGroupRoleBody, validateMemberRoleBody , ValidateNickNamechangeBody} = require("../middlewares/validators/discordactions");
 const checkIsVerifiedDiscord = require("../middlewares/verifydiscord");
 
 const router = express.Router();
@@ -9,5 +9,6 @@ const router = express.Router();
 router.post("/groups", authenticate, checkIsVerifiedDiscord, validateGroupRoleBody, createGroupRole);
 router.get("/groups", authenticate, checkIsVerifiedDiscord, getAllGroupRoles);
 router.post("/roles", authenticate, checkIsVerifiedDiscord, validateMemberRoleBody, addGroupRoleToMember);
+router.post("/nickname" , authenticate,checkIsVerifiedDiscord, ValidateNickNamechangeBody, changeNicknameOfUsers)
 
 module.exports = router;
--- a/middlewares/validators/discordactions.js
+++ b/middlewares/validators/discordactions.js
@@ -28,7 +28,22 @@ const validateMemberRoleBody = async (req, res, next) => {
   }
 };
 
+const ValidateNickNamechangeBody = async(req,res, next) =>{
+  const schema = Joi.object({
+    userName: Joi.string().trim().required()
+  })
+
+  try {
+    await schema.validateAsync(req.body);
+    next();
+  } catch (error) {
+    logger.error(`Error validating user crendential : ${error}`);
+    res.boom.badRequest(error.details[0].message);
+  }
+}
+
 module.exports = {
   validateGroupRoleBody,
   validateMemberRoleBody,
+  ValidateNickNamechangeBody
 };


controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
...
+    const authToken = await jwt.sign({}, config.get("rdsServerlessBot.rdsServerLessPrivateKey"), {
+      algorithm: "RS256",
+      expiresIn: config.get("rdsServerlessBot.ttl"),
+    });
+
+    await fetch(`${DISCORD_BASE_URL}/guild/member`, {
+      method: "PATCH",
+      body: JSON.stringify(dataForDiscord),
+      headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+    }).then((response) => response.json());
```
heyrandhir, line 148: it is generally a good practice to wrap `fetch` calls in a `try...catch` block to handle any errors that may occur during the request.

controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
     return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
   }
 };
+/**
+ * Patch Update user nickname
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const changeNicknameOfUsers = async (req, res) => {
```
heyrandhir, line 128: shouldn't the business logic be inside controller

controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
...
+  try {
+    const userData = req.userData;
 
+    const { discordId, username } = userData;
```
heyrandhir, line 132: Not all users in the DB have `discordId`. how do we handle that 🤔

controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
...
+
+    const dataForDiscord = {
+      userName: username,
+      discordId,
+    };
```
heyrandhir, line 137: Nit: `discordData` sounds better than `dataForDiscord`.

routes/discordactions.js
```diff
@@ -16,6 +18,7 @@ const router = express.Router();
 router.post("/groups", authenticate, checkIsVerifiedDiscord, validateGroupRoleBody, createGroupRole);
 router.get("/groups", authenticate, checkIsVerifiedDiscord, getAllGroupRoles);
 router.post("/roles", authenticate, checkIsVerifiedDiscord, validateMemberRoleBody, addGroupRoleToMember);
+router.post("/nickname", authenticate, checkIsVerifiedDiscord, changeNicknameOfUsers);
```
heyrandhir, line 21: should it be post or patch as you are updating a resource partially

heyrandhir (changes requested):

controllers/discordactions.js
khushi818: in other APIs similar to this, it is done the way I have written. I followed the same pattern of calling fetch inside API call.

controllers/discordactions.js
khushi818: we have checkInVerfified middleware for the same. if you are verified then this API will be called then it will add discordId in the userData.

controllers/discordactions.js
khushi818: yup changed

routes/discordactions.js
khushi818: with  patch it shows not found and I referred from another api they are post and called put request inside it. so followed the same.

controllers/discordactions.js
khushi818: same reason @heyrandhir https://github.com/Real-Dev-Squad/website-backend/pull/1169#discussion_r1237127130

controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
...
+
+    const authToken = await jwt.sign({}, config.get("rdsServerlessBot.rdsServerLessPrivateKey"), {
+      algorithm: "RS256",
+      expiresIn: config.get("rdsServerlessBot.ttl"),
+    });
+
+    await fetch(`${DISCORD_BASE_URL}/guild/member`, {
```
DashDeipayan, line 144: we shouldn't use await and then together. Rather it is encouraged to use await over then for more predictive execution. Let me know if you need a bit more clarification.

controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
...
+      method: "PATCH",
+      body: JSON.stringify(discordData),
+      headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+    }).then((response) => response.json());
+
+    return res.status(200).json({
```
DashDeipayan, line 150: no need to use `status(200)` as it is 200 by default

controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
...
+  try {
+    const userData = req.userData;
 
+    const { discordId, username } = userData;
+
+    const discordData = {
+      userName: username,
+      discordId,
+    };
+
```
DashDeipayan, line 138:
```suggestion

    const { discordId, username:userName } = req.userData;

    const discordData = { userName,  discordId };

```

routes/discordactions.js
DashDeipayan, line 21: Just asking, you sure its not a SUPERUSER-only route?

test/integration/discord.test.js
```diff
@@ -75,4 +75,70 @@ describe("test discord actions", function () {
         });
     });
   });
+
+  describe("test discord actions for nickname for verified user", function () {
+    beforeEach(async function () {
+      fetchStub = sinon.stub(global, "fetch");
+      const user = { ...userData[4], discordId: "123456789" };
```
DashDeipayan, line 82: any reason for specifically choosing `userData[4]`?
If the reason is SUPER_USER, then we may name the variable appropriately :)

controllers/discordactions.js
```diff
@@ -121,7 +120,41 @@ const addGroupRoleToMember = async (req, res) => {
...
+    const authToken = await jwt.sign({}, config.get("rdsServerlessBot.rdsServerLessPrivateKey"), {
+      algorithm: "RS256",
+      expiresIn: config.get("rdsServerlessBot.ttl"),
+    });
+
+    await fetch(`${DISCORD_BASE_URL}/guild/member`, {
+      method: "PATCH",
+      body: JSON.stringify(discordData),
+      headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+    }).then((response) => response.json());
```
DashDeipayan, line 148: Any reason why we aren't storing any record of the name-change in our database?

DashDeipayan (changes requested):

controllers/discordactions.js
khushi818: it directly changes in discord itself. Rest I received no guideline related to it

test/integration/discord.test.js
khushi818: no I just chose it and added discord id. I can change it if you suggest.

routes/discordactions.js
khushi818: I have added superuser route

test/integration/discord.test.js
khushi818: I have updated

DashDeipayan (dismissed):
LGTM

test/integration/discord.test.js
```diff
@@ -75,4 +78,74 @@ describe("test discord actions", function () {
...
+      chai
+        .request(app)
+        .post("/discord-actions/nickname")
+        .set("Cookie", `${cookieName}=${superUserAuthToken}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body.message).to.equal("nickname has been changed");
```
heyrandhir, line 112: would it be a good idea to even send the old name and updated name in the response as It will allow clients to validate that the expected change was made  correctly.
Something of the sort 
```json
{
  "status": 200,  
  "message": "nickname has been changed",               
  "data": {
     "previous": "Khushi",
     "updated": "Amazing Khushi"
   }
}
```

heyrandhir (changes requested):
I still believe Instead of making a POST call that internally patches the data,  we should investigate why the PATCH call is not working correctly.

test/integration/discord.test.js
```diff
@@ -75,4 +78,74 @@ describe("test discord actions", function () {
         });
     });
   });
+
+  describe("test discord actions for nickname for verified user", function () {
+    beforeEach(async function () {
+      fetchStub = sinon.stub(global, "fetch");
+      superUser = { ...userData[4], discordId: "123456789" };
+      userId = await addUser();
```
heyrandhir, line 86: if `userId` is unused then it can be removed.

test/integration/discord.test.js
```diff
@@ -75,4 +78,74 @@ describe("test discord actions", function () {
...
+          expect(res.body.message).to.equal("nickname has been changed");
+          return done();
+        });
+    });
+  });
+
+  describe("test discord actions for nickname for unverified user", function () {
+    beforeEach(async function () {
+      const { discordId, ...superUser } = userData[4];
+      userId = await addUser();
```
heyrandhir, line 121: if `userId` is unused then it can be removed.

heyrandhir (changes requested):
Also please confirm if the API Contract has been updated.

test/integration/discord.test.js
khushi818: as you see in the api , There are no previous and updated username. these will be updated from mysite so there is no need to add this.

khushi818:
> I still believe Instead of making a POST call that internally patches the data, we should investigate why the PATCH call is not working correctly.

How much I understand it , we are not taking payload also we are calling patch api from discord-slash. we are also not changing the database. I don't see point in patch request. @heyrandhir

RitikJaiswal75:
please add test coverage report

heyrandhir:
> > I still believe Instead of making a POST call that internally patches the data, we should investigate why the PATCH call is not working correctly.
> 
> How much I understand it , we are not taking payload also we are calling patch api from discord-slash. we are also not changing the database. I don't see point in patch request. @heyrandhir

However we are still updating an existing member of the guild. It is updating just the nickname of the user, not replacing the entire member object. This is a partial update, which points to PATCH. 

@DashDeipayan  @RitikJaiswal75 please what is your opinion on this.

test/integration/discord.test.js
heyrandhir: The `userId ` on line 88 is the key. This `userId` is not used anywhere I think

heyrandhir (changes requested):

RitikJaiswal75:
> However we are still updating an existing member of the guild. It is updating just the nickname of the user, not replacing the entire member object. This is a partial update, which points to PATCH.

I agree on this since we are just modifying 1 user in 1 call

routes/discordactions.js
```diff
@@ -16,6 +18,7 @@ const router = express.Router();
 router.post("/groups", authenticate, checkIsVerifiedDiscord, validateGroupRoleBody, createGroupRole);
 router.get("/groups", authenticate, checkIsVerifiedDiscord, getAllGroupRoles);
 router.post("/roles", authenticate, checkIsVerifiedDiscord, validateMemberRoleBody, addGroupRoleToMember);
+router.post("/nickname", authenticate, authorizeRoles([SUPERUSER]), checkIsVerifiedDiscord, changeNicknameOfUsers);
```
RitikJaiswal75, line 21: Please change the the API name it shouldn't be /discord-actions/nickname

please use nouns and identify which resource is being updated discord-actions is not a resource

for better API naming please read:https://swagger.io/resources/articles/best-practices-in-api-design/

controllers/discordactions.js
```diff
@@ -121,7 +120,38 @@ const addGroupRoleToMember = async (req, res) => {
     return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
   }
 };
+/**
+ * Patch Update user nickname
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const changeNicknameOfUsers = async (req, res) => {
```
heyrandhir, line 128: Nit: we are changing a single user name at time. we can name the function better such as `changeUserNickname`

controllers/discordactions.js
```diff
@@ -3,7 +3,6 @@ const admin = require("firebase-admin");
 const config = require("config");
 const jwt = require("jsonwebtoken");
 const discordRolesModel = require("../models/discordactions");
-
```
khushi818, line 6: space removal is linting

controllers/users.js
```diff
@@ -602,6 +601,31 @@ const setInDiscordScript = async (req, res) => {
...
+    const type = req.body.type;
+    const { user } = await userQuery.fetchUser({ userId: req.params.userId });
+
+    const { discordId, username: userName } = user;
+
+    if (type === "discord") {
+      setDiscordNickname(userName, discordId);
+    }
+
+    return res.json({
```
shubham-y, line 620: I will suggest using `res.status(200).json()` as from express v5 `res.json` will be deprecated
Reference - https://expressjs.com/en/guide/migrating-5.html#res.json

controllers/users.js
```diff
@@ -602,6 +601,31 @@ const setInDiscordScript = async (req, res) => {
...
+const changeUserNickname = async (req, res) => {
+  try {
+    const type = req.body.type;
+    const { user } = await userQuery.fetchUser({ userId: req.params.userId });
+
+    const { discordId, username: userName } = user;
+
+    if (type === "discord") {
+      setDiscordNickname(userName, discordId);
+    }
```
shubham-y, line 618: What if this condition results in a false?
Even then we will be sending a successful response?

controllers/users.js
```diff
@@ -602,6 +601,31 @@ const setInDiscordScript = async (req, res) => {
...
+ * Patch Update user nickname
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const changeUserNickname = async (req, res) => {
+  try {
+    const type = req.body.type;
+    const { user } = await userQuery.fetchUser({ userId: req.params.userId });
+
+    const { discordId, username: userName } = user;
```
shubham-y, line 614: Just wanted to check that we are confident that the user object will have `discordId` present once we reach this controller function otherwise the API will fail

controllers/users.js
```diff
@@ -602,6 +601,31 @@ const setInDiscordScript = async (req, res) => {
   }
 };
 
+/**
+ * Patch Update user nickname
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const changeUserNickname = async (req, res) => {
```
shubham-y, line 609: **Suggestion**: I think the function name `updateDiscordNickname` will be more readable and easy to understand. Let me know what do you think about it

test/integration/discord.test.js
```diff
@@ -75,4 +78,75 @@ describe("test discord actions", function () {
...
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body.message).to.equal("nickname has been changed");
+          return done();
+        });
+    });
+  });
+
+  describe("test discord actions for nickname for unverified user", function () {
```
shubham-y, line 119: I suggest we don't need separate describe blocks for the same API. We can have a single describe block and have one it block for verified user expecting 200 response and another it block for unverified users expecting 403 response

controllers/users.js
```diff
@@ -602,6 +601,31 @@ const setInDiscordScript = async (req, res) => {
...
+  try {
+    const type = req.body.type;
+    const { user } = await userQuery.fetchUser({ userId: req.params.userId });
+
+    const { discordId, username: userName } = user;
+
+    if (type === "discord") {
+      setDiscordNickname(userName, discordId);
```
shubhamsigdar1, line 617: Nit: setDiscordNickname : setUserDiscordNickname

test/integration/discord.test.js
```diff
@@ -75,4 +78,75 @@ describe("test discord actions", function () {
...
+    beforeEach(async function () {
+      const { discordId, ...superUser } = userData[4];
+      superUserId = await addUser(superUser);
+      superUserAuthToken = authService.generateAuthToken({ userId: superUserId });
+    });
```
shubham-y, line 124: Don't we need to initialize fetchStub here as well before using it?
```js
fetchStub = sinon.stub(global, "fetch");
```

services/discordService.js
```diff
@@ -64,8 +64,24 @@ const addRoleToUser = async (userid, roleid) => {
...
+  });
+
+  await (
+    await fetch(`${DISCORD_BASE_URL}/guild/member`, {
+      method: "PATCH",
+      body: JSON.stringify({ userName, discordId }),
+      headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+    })
+  ).json();
+};
```
shubhamsigdar1, line 80: why it is not inside try and catch??

routes/users.js
```diff
@@ -40,5 +40,12 @@ router.patch("/profileURL", authenticate, userValidator.updateProfileURL, users.
 router.patch("/rejectDiff", authenticate, authorizeRoles([SUPERUSER]), users.rejectProfileDiff);
 router.patch("/:userId", authenticate, authorizeRoles([SUPERUSER]), users.updateUser);
 router.get("/suggestedUsers/:skillId", authenticate, authorizeRoles([SUPERUSER]), users.getSuggestedUsers);
+router.patch(
+  "/servername/:userId",
+  authenticate,
+  authorizeRoles([SUPERUSER]),
+  checkIsVerifiedDiscord,
```
shubham-y, line 47: Please update the JSDoc comment on this middleware. As it's mentioned that this middleware is not needed to be called explicitly

Check line number 5 below

https://github.com/Real-Dev-Squad/website-backend/blob/8e951ada9106cada0020d08460241f0652bd0e81/middlewares/verifydiscord.js#L1-L15

services/discordService.js
```diff
@@ -64,8 +64,24 @@ const addRoleToUser = async (userid, roleid) => {
...
+
```
shubhamsigdar1, line 81: Nit: Want to Know this
Is this api is inbuild `${DISCORD_BASE_URL}/guild/member`
I didn't understand what this fetch is doing??

Can you give me some idea??

controllers/users.js
```diff
@@ -602,6 +601,31 @@ const setInDiscordScript = async (req, res) => {
...
 };
 
+/**
+ * Patch Update user nickname
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+const changeUserNickname = async (req, res) => {
+  try {
+    const type = req.body.type;
```
shubhamsigdar1, line 611: Nit: Want to know
what are the data you sending in body object...can you please refer me on API Contract??

shubhamsigdar1 (commented):
Great work @khushi818

controllers/users.js
RitikJaiswal75, line 1: I can not see this file in your test coverage report

routes/users.js
```diff
@@ -40,5 +40,12 @@ router.patch("/profileURL", authenticate, userValidator.updateProfileURL, users.
 router.patch("/rejectDiff", authenticate, authorizeRoles([SUPERUSER]), users.rejectProfileDiff);
 router.patch("/:userId", authenticate, authorizeRoles([SUPERUSER]), users.updateUser);
 router.get("/suggestedUsers/:skillId", authenticate, authorizeRoles([SUPERUSER]), users.getSuggestedUsers);
+router.patch(
+  "/servername/:userId",
```
RitikJaiswal75, line 44: could you link me to the discussion which lead to this name

services/discordService.js
```diff
@@ -64,8 +64,24 @@ const addRoleToUser = async (userid, roleid) => {
   return response;
 };
 
+const setUserDiscordNickname = async (userName, discordId) => {
+  const authToken = jwt.sign({}, config.get("rdsServerlessBot.rdsServerLessPrivateKey"), {
+    algorithm: "RS256",
+    expiresIn: config.get("rdsServerlessBot.ttl"),
+  });
```
RitikJaiswal75, line 71: there is a utility named `generateAuthTokenForCloudflare` Please use that here

test/integration/discord.test.js
RitikJaiswal75, line 1: of this is a users api why are we writing its tests in discord test file?

services/discordService.js
```diff
@@ -51,6 +51,20 @@ const addRoleToUser = async (userid, roleid) => {
   return response;
 };
 
+const setUserDiscordNickname = async (userName, discordId) => {
+  const authToken = await generateAuthTokenForCloudflare();
```
DashDeipayan, line 55: It should be in a try-catch block ideally

test/integration/users.test.js
```diff
@@ -1229,4 +1230,71 @@ describe("Users", function () {
...
+      fetchStub.returns(
+        Promise.resolve({
+          status: 200,
+          json: () => Promise.resolve({}),
+        })
+      );
+      chai
+        .request(app)
+        .patch(`/users/servername/${userId}`)
+        .set("Cookie", `${cookieName}=${superUserAuthToken}`)
```
DashDeipayan, line 1255: the route isn't `SUPERUSER` restricted, I guess we do not need to provide `SuperUserAuthToken` for this

test/integration/users.test.js
```diff
@@ -1229,4 +1230,71 @@ describe("Users", function () {
...
```
DashDeipayan, line 1290: same as above

DashDeipayan (changes requested):

controllers/users.js
```diff
@@ -607,6 +606,32 @@ const setInDiscordScript = async (req, res) => {
...
+    const { discordId, username: userName } = user;
+
+    if (!discordId) {
+      throw new Error("user not verified");
+    }
+
+    const response = await setUserDiscordNickname(userName, discordId);
+
+    return res.json({
+      message: response,
```
RitikJaiswal75, line 627: add some better information in response rather than just sending back message, try making the response more user friendly.

services/discordService.js
RitikJaiswal75: +1

test/integration/users.test.js
RitikJaiswal75: in the current iteration, it is super user restricted.

PR #1170, opened 2023-06-16
fixed the /tasks/id route

Pratiyushkumar:
Issue - while sending the patch request to /task/:id route, was getting a response message as "User doesn't exist" with status 404

screenshot of the bug
image

fix - in updateTask controller just changed the if condition from req.body?.assignee -> task.taskData.assignee and also pass username in fetchUser function as task.taskData.assignee

screenshot after the fix
image
image

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -178,7 +178,7 @@ const updateTask = async (req, res) => {
     if (!task.taskData) {
       return res.boom.notFound("Task not found");
     }
-    if (task.taskData.assignee) {
+    if (task.taskData?.assignee) {
       const user = await userQuery.fetchUser({ username: task.taskData.assignee });
       if (!user.userExists) {
         return res.boom.notFound("User doesn't exist");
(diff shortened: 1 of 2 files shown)


RitikJaiswal75 (dismissed):

khushi818 (dismissed):

controllers/tasks.js
```diff
@@ -178,8 +178,8 @@ const updateTask = async (req, res) => {
     if (!task.taskData) {
       return res.boom.notFound("Task not found");
     }
-    if (req.body?.assignee) {
-      const user = await userQuery.fetchUser({ username: req.body.assignee });
+    if (task.taskData.assignee) {
```
khushi818: do the optional chaining here

khushi818 (approved):

khushi818 (approved):

test/integration/tasks.test.js
```diff
@@ -381,7 +381,7 @@ describe("Tasks", function () {
         });
     });
 
-    it("should return 404 if assignee is not in user db", function (done) {
+    it("should return 204 if assignee is not in user db", function (done) {
```
RitikJaiswal75, line 384: why do we return 204 for users not found? not found should be 404

RitikJaiswal75 (changes requested):

Pratiyushkumar:
This was a false alarm, the change can break existing flow

PR #1181, opened 2023-06-19
Script to Add Unverified Role to Users Without Discord Link

RitikJaiswal75:
### Description
This pull request introduces a script that adds an "unverified" role to all users who haven't linked their Discord accounts with our backend. The purpose of this script is to streamline the verification process and ensure consistent user management. Additionally, a superuser-only API endpoint has been created to facilitate this action securely.

### Changes Made
- Developed a script that identifies users without a linked Discord account and assigns them the "unverified" role.
- Created a new superuser-only API endpoint specifically designed for executing this action securely.
- Tested the script and API endpoint extensively to verify their functionality and performance.

### Testing
The script and API endpoint have undergone comprehensive testing to ensure accurate identification of users without a linked Discord account and proper assignment of the "unverified" role. Testing includes various scenarios, such as users with and without existing roles, to ensure the script behaves as expected.

### Impact
Implementing this script and API endpoint offers the following benefits:
- Improved user management: Automatically assigning the "unverified" role simplifies the verification process for users who haven't linked their Discord accounts.
- Streamlined workflow: Superusers can execute this action securely through the dedicated API endpoint, reducing manual intervention.

Please review the changes made in this pull request and provide any feedback or suggestions for improvement.

Thank you.

### DIFF
--- a/controllers/discordactions.js
+++ b/controllers/discordactions.js
@@ -3,6 +3,8 @@ const admin = require("firebase-admin");
 const config = require("config");
 const jwt = require("jsonwebtoken");
 const discordRolesModel = require("../models/discordactions");
+const { addRoleToUser, getDiscordMembers } = require("../services/discordService");
+const { fetchAllUsers } = require("../models/users");
 
 /**
  * Creates a role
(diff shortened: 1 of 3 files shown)


controllers/discordactions.js
```diff
@@ -141,9 +143,36 @@ const updateDiscordImageForVerification = async (req, res) => {
   }
 };
 
+const markUnverified = async (req, res) => {
+  try {
+    const [usersInRdsDiscordServer, allRdsLoggedInUsers] = await Promise.all([getDiscordMembers(), fetchAllUsers()]);
+    const rdsUserMap = {};
```
iamitprakash, line 149: i don't see this rdsUserMap is not cleared anywhere, that mean's this is grow as user(s) will login

services/discordService.js
```diff
@@ -52,7 +52,31 @@ const setInDiscordFalseScript = async () => {
   await Promise.all(updateUsersPromises);
 };
 
+const generateAuthTokenForCloudflare = async () => {
```
iamitprakash, line 55: where did you use authToken, that means 403 error

iamitprakash (changes requested):

services/discordService.js
yesyash, line 55: - Why is this inside `services` and not `utils` ?

services/discordService.js
```diff
@@ -52,7 +52,31 @@ const setInDiscordFalseScript = async () => {
...
+
+const addRoleToUser = async (userid, roleid) => {
+  const authToken = await generateAuthTokenForCloudflare();
+  const response = await (
+    await fetch(`${DISCORD_BASE_URL}/roles/add`, {
+      method: "PUT",
+      headers: { "Content-Type": "application/json", Authorization: `Bearer ${authToken}` },
+      body: JSON.stringify({ userid, roleid }),
+    })
+  ).json();
```
yesyash, line 73: - I think a better way would be to handle this in two separate variables and add a try-catch block to the function. What do you say ?

services/discordService.js
RitikJaiswal75: Made the changes

services/discordService.js
RitikJaiswal75: The consumer of this service has a try catch block that would handle it, Wont that suffice? would love to hear your inputs on this.

services/discordService.js
RitikJaiswal75: It is being consumed in the function that makes the fetch call just below this function

controllers/discordactions.js
RitikJaiswal75: As per the discussion, the map won't increase as this code works for existing users only, new users automatically get that role applied.

iamitprakash (approved):

heyrandhir (approved):

PR #1190, opened 2023-06-22
Patch call for updating authorization status of user on mobile app

fakhruddinkw:
This PR is regarding: #942 

before-patch-db-state

patch-call-response

after-patch-db-state

**Error: 401 response**
401-unauthorized

**Error: 404 response**
404-document_not_found

**Error: 500 response**
500-error


controllers/qrCodeAuth.js
```diff
@@ -0,0 +1,26 @@
+const qrAuthQuery = require("../models/qrCodeAuth");
+const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
+
+const updateAuthStatus = async (req, res) => {
+  try {
+    const userId = req.userData.id;
```
isVivek99, line 6: i don't think we are passing any data to the request

isVivek99: we will be getting it from the token itself

isVivek99:
make sure we handle all status response `4xx` given in the API contract
(here)[https://github.com/Real-Dev-Squad/website-api-contracts/pull/134/files#diff-4df11ed74a4b8d770caaa7512db70b9845093a1d1581d82452748ad27012ea34]

controllers/qrCodeAuth.js
fakhruddinkw: Yes you are right, we are not passing any data to request. But the thing is, in our authenticate middleware, we are storing userId on req.userData object, so i just referenced id on that object. 😅😅

isVivek99: ok i checked , rn looks good then

models/qrCodeAuth.js
```diff
@@ -0,0 +1,34 @@
...
+
+    if (!authData) {
+      return {
+        userExists: false,
+      };
+    }
+
+    await qrAuthModel.doc(userId).set({
+      ...authData.data(),
+      authStatus,
```
isVivek99, line 16: it is not `authStatus` we are returning `authorization_status`, according to the contract.
jut make it
```
...set({
 ...authData.data(),
    authorization_status : authStatus,
})
```

models/qrCodeAuth.js
```diff
@@ -0,0 +1,34 @@
+const firestore = require("../utils/firestore");
+const qrAuthModel = firestore.collection("Qr_auth");
```
isVivek99, line 2: collection name must be `QrCodeAuth` according to the contract

isVivek99 (changes requested):

isVivek99:
Can we also add `unit` and `integration-test`, now? everything else looks same

heyrandhir (changes requested):
Please add the tests.

middlewares/validators/qrCodeAuth.js
```diff
@@ -0,0 +1,22 @@
+const joi = require("joi");
+
+const validateAuthStatus = async (req, res, next) => {
+  const schema = joi
+    .object()
+    .strict()
+    .keys({
+      authorization_status: joi.string().valid("AUTHORIZED", "REJECTED", "NOT_INIT").optional(),
```
shubham-y, line 8: Is `optional()` needed here?
As `authorization_status` request param needs to be present for this route to be called

controllers/qrCodeAuth.js
```diff
@@ -0,0 +1,26 @@
...
+    const authStatus = req.params.authorization_status;
+    const result = await qrAuthQuery.updateStatus(userId, authStatus);
+
+    if (result.userExists) {
```
shubham-y, line 10: I would suggest following the fail fast principle and add a check for `!result.userExisits` and returning 404
Otherwise, return success response outside `if` block
It also improves the readability of code

controllers/qrCodeAuth.js
```diff
@@ -0,0 +1,26 @@
...
+      return res.json({
+        message: `Authentication document for user ${userId} updated successfully`,
+        data: { ...result.data },
+      });
+    }
+
+    return res.boom.notFound("Document not found!");
+  } catch (error) {
+    logger.error(`Error while fetching user: ${error}`);
+    return res.boom.serverUnavailable(SOMETHING_WENT_WRONG);
```
shubham-y, line 20: Please use `res.boom.badImplementaion` instead

middlewares/validators/qrCodeAuth.js
fakhruddinkw: I kept this as optional, because if no status is passed, we are setting authorization_status as "NOT_INIT" by default. @shubham-y

fakhruddinkw: the authorization_status is optional, route: "/qr-code-auth/authorization_status/:authorization_status?" @shubham-y

shreya-mishra (commented):
please resolve all the comments which are already mentioned

middlewares/validators/qrCodeAuth.js
isVivek99: is it optional in the contract, i don't think it is, can u check once?

test/integration/qrCodeAuth.test.js
```diff
@@ -0,0 +1,112 @@
...
+    jwt = authService.generateAuthToken({ userId });
+    userDeviceInfoData = { ...userDeviceInfoDataArray[0], user_id: userId };
+    userDeviceInfoWithAuthStatus = { ...userDeviceInfoData, authorization_status: "NOT_INIT" };
+  });
+
+  afterEach(async function () {
+    await cleanDb();
+  });
+
+  describe("PATCH CALL", function () {
```
isVivek99, line 35: why captial?, can describe better

middlewares/validators/qrCodeAuth.js
fakhruddinkw: I used optional, so that when we call the patch route without any status, by default it will set the status to "NOT_INIT".

isVivek99: can we confirm, this with @shreya-mishra or @heyrandhir ?

shubham-y: > the authorization_status is optional, route: "/qr-code-auth/authorization_status/:authorization_status?" @shubham-y

Ohh right, I wasn't aware that it's possible to do this way in express.

shubham-y: @fakhruddinkw So have we decided to remove optional?

models/qrCodeAuth.js
```diff
@@ -0,0 +1,53 @@
...
+        authorization_status: authStatus,
+      },
+    };
+  } catch (err) {
+    logger.error("Error in updating auth status", err);
+    throw err;
+  }
+};
+
+const storeUserDeviceInfo = async (userDeviceInfoData) => {
```
shubham-y, line 34: It seems like we are using this model function only for tests. Are we going to use it elsewhere as well for a route?
If not I would suggest moving it to the utils folder of test

models/qrCodeAuth.js
```diff
@@ -0,0 +1,53 @@
...
+  try {
+    const { user_id: userId } = userDeviceInfoData;
+    const user = await userModel.doc(userId).get();
+    if (user.data()) {
+      await qrAuthModel.doc(userId).set(userDeviceInfoData);
+      return { userDeviceInfoData };
+    } else {
+      throw new Error(USER_DOES_NOT_EXIST_ERROR);
+    }
```
shubham-y, line 43: The same suggestion of using the fail fast principle
https://github.com/Real-Dev-Squad/website-backend/pull/1190#discussion_r1242117868

models/qrCodeAuth.js
fakhruddinkw: This function is part of @isVivek99 's PR for post call. I just included this here for testing

models/qrCodeAuth.js
shubham-y: As discussed, this will be handled in @isVivek99  PR

shubham-y (approved):

middlewares/validators/qrCodeAuth.js
fakhruddinkw: @shubham-y Yes i had a discussion regarding this with @prakashchoudhary07 , and keeping the route as well as the schema optional is not needed.

heyrandhir (approved):

shreya-mishra (approved):

isVivek99 (approved):

sakshambhatt (approved):
LGTM :rocket:

PR #1191, opened 2023-06-23
Remove user personal details filters data

sahsisunny:
## Closes: #1192

API Link : https://api.realdevsquad.com/users/search?state=IDLE&state=ACTIVE&state=OOO

**Description:**
The current above API returns personal details like Phone , E-mail and GitHub Access tokens. In this PR we are removing user personal details from the response.

**Test Coverage**
image

**Screenshot**
Before
image

After
image

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -576,7 +576,7 @@ const filterUsers = async (req, res) => {
       return res.boom.badRequest("filter for item not provided");
     }
     const users = await userQuery.getUsersBasedOnFilter(req.query);
-    const cleanedUsers = users.map((user) => {
+    const sanitizedUsers = users.map((user) => {
       delete user.tokens;
       delete user.email;
       delete user.phone;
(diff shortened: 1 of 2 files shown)


heyrandhir (dismissed):

RitikJaiswal75 (dismissed):

shubhamsigdar1 (approved):
Great work @sahsisunny

heyrandhir (approved):

controllers/users.js
```diff
@@ -576,10 +576,15 @@ const filterUsers = async (req, res) => {
       return res.boom.badRequest("filter for item not provided");
     }
     const users = await userQuery.getUsersBasedOnFilter(req.query);
-
+    const sanitizedUsers = users.map((user) => {
```
heyrandhir, line 579: I liked `sanitizedUsers`. Sounds better than `cleanedUsers` 👍🏻

shubhamsigdar1: `sanitizedUsersDetail`?? :)

prakashchoudhary07:
We should create a data access layer, which will always return the user's filtered data

sahsisunny:
> We should create a data access layer, which will always return the user's filtered data/
> So that if passing it through that code, will remove sensitive data

Okay so can I create the issue ticket for this ?

PR #1210, opened 2023-06-28
FIX : Add custom word limit module

sahsisunny:


middlewares/validators/user.js
```diff
@@ -66,9 +68,18 @@ const validateJoinData = async (req, res, next) => {
       country: joi.string().min(1).required(),
       foundFrom: joi.string().min(1).required(),
       introduction: joi.string().min(1).required(),
```
shubhamsigdar1, line 70: On introduction also validate based on word length

shubhamsigdar1 (commented):
Great work @sahsisunny...can you please check my 1 comment

middlewares/validators/user.js
sahsisunny: For how many words?

shubhamsigdar1: 100 words

shubhamsigdar1 (approved):

heyrandhir (approved):

Issue #1184, opened 2023-06-21
User Status Management Updates


heyrandhir:
**Parent Ticket:** https://github.com/Real-Dev-Squad/todo-action-items/issues/155

**Summary:**
Update user status management on "My Site" to remove the ability for users to manually mark themselves as Idle or Active. Implement a task-based status change logic in the backend as per the workflow described.

**Details:**
1. Currently, users on "My Site" have the ability to manually mark themselves as Idle, Active, or OOO. However, this functionality will be removed from the site.
2. Users will only be able to mark themselves as OOO from "My Site." All other status changes (Idle and Active) will be determined based on task assignments and task completion.
3. Implement the following workflow for user status changes:
   - When a task is assigned to a user, their status should be automatically set as Active if their current status is not already Active (i.e., Idle or Onboarding).
   - If a user is marked as OOO, their future status should be set as Active, allowing them to resume their tasks upon returning.
   - When a user completes a task, the system should check if there are any existing assigned or in-progress tasks for that user.
     - If there are assigned or in-progress tasks, the user's status should remain Active.
     - If there are no assigned or in-progress tasks, the system should check if the user is OOO.
       - If the user is not OOO, their status should be marked as Idle.
       - If the user is OOO, their future status should be marked as OOO.
4. Update the backend logic to handle the task-based status changes as per the workflow mentioned above.

**UML Sequence Diagram:**

The UML sequence diagram depicting the updated user status management logic has been attached for reference. It illustrates the revised flow of actions and interactions between the user and the system components.

1. UML Sequence Diagram for Status update from My Site 
![image](https://github.com/Real-Dev-Squad/website-backend/assets/97341921/d0fa3f59-6ea9-493e-a787-42993f29163d)

2. UML Sequence Diagram for User Status update when New Task is Created or Updated (Website Backend)
![image](https://github.com/Real-Dev-Squad/website-backend/assets/97341921/011c3418-32c7-4aa2-ba8f-fd60b3f2914d)

3. UML Sequence Diagram for User Status update post Task Completion (Website Backend)
![image](https://github.com/Real-Dev-Squad/website-backend/assets/97341921/04dd0e75-39ff-4d6e-b813-541b2bafde66)

heyrandhir:
![image](https://github.com/Real-Dev-Squad/website-backend/assets/97341921/ade1534d-22b2-408b-bfe3-df1c7c64897c)

heyrandhir:
[...]

PR #1220, opened 2023-07-01
Status Change Logic for Task Completion and Assignment

heyrandhir:
**Parent Ticket** - https://github.com/Real-Dev-Squad/website-backend/issues/1184
This PR closes the 1st and 2nd task from the parent ticket https://github.com/Real-Dev-Squad/website-backend/issues/1184#issuecomment-1603016625

This PR introduces the logic for handling status changes based on task completion and assignment, along with the implementation of a feature flag. The feature flag allows controlling the application's behavior related to the status changes until the feature is stable. The ticket to remove the feature flag post 6 weeks of feature going live  https://github.com/Real-Dev-Squad/website-backend/issues/1224

The following table summarizes the status changes in different scenarios:

**Status Change on Marking Task as Completed**

| Sl No | Any other Task Assigned | Current Status | New Status                 |
|-------|---------------|----------------|----------------------------|
| 1     | yes           | DNE            | Current Status - ACTIVE       |
| 2     | Yes           | OOO            | Future Status - ACTIVE        |
| 3     | Yes           | ACTIVE         | No Change                  |
| 4     | Yes           | != ACTIVE & != OOO      | Current Status - ACTIVE       |
| 5     | No            | DNE            | Current Status - IDLE         |
| 6     | No            | OOO            | Future Status - IDLE       |
| 7     | No            | IDLE           | Current Status - No Change |Ï
| 8     | No            | != IDLE  & != OOO      | Current Status - IDLE      |

**Status Change on Assigning / Updating a new Task**

 | Sl No | Current Status | New Status               |
|-------|----------------|--------------------------|
| 1     | DNE            | Current Status \- ACTIVE |
| 2     | OOO            | Future Status \- ACTIVE  |
| 3     | ACTIVE         | No Change                |
| 4     | \!= ACTIVE  & != OOO   | Current Status \- ACTIVE |

_DNE - Does Not Exist, OOO - Out of Office_

This logic ensures that the status is appropriately updated based on task completion and assignment scenarios.

The Test Coverage Stats are as follows : 

Models 

image

Utils 

image

Controllers

image


models/userStatus.js
```diff
@@ -193,4 +205,133 @@ const updateAllUserStatus = async () => {
...
+ */
+const updateUserStatusOnNewTaskAssignment = async (userId) => {
+  try {
+    let latestStatusData;
+    try {
+      latestStatusData = await getUserStatus(userId);
+    } catch (error) {
+      logger.error("Unable to retrieve the current status" + error.message);
+      throw new Error("Unable to retrieve the current status");
+    }
```
bhtibrewal, line 226: Why do we have 2 try catch blocks here, never seen somewhere

models/userStatus.js
```diff
@@ -193,4 +205,133 @@ const updateAllUserStatus = async () => {
...
+      return createUserStatusWithState(userId, userStatusModel, userState.ACTIVE);
+    }
+    const {
+      data: {
+        currentStatus: { state },
+      },
+    } = latestStatusData;
+    if (state === userState.ACTIVE) {
+      return generateAlreadyExistingStatusResponse(userState.ACTIVE);
+    } else if (state === userState.IDLE || state === userState.ONBOARDING) {
```
bhtibrewal, line 238:
```suggestion
    if (state === userState.ACTIVE) {
      return generateAlreadyExistingStatusResponse(userState.ACTIVE);
    } if (state === userState.IDLE || state === userState.ONBOARDING) {
```

models/userStatus.js
```diff
@@ -193,4 +205,133 @@ const updateAllUserStatus = async () => {
...
+ * }>} - The response object indicating the status of the user status update or an error.
+ */
+
+const updateUserStatusOnTaskUpdate = async (userName) => {
+  let userId;
+  let userStatusUpdate;
+  try {
+    userId = await getUserIdFromUserName(userName, usersCollection);
+    userStatusUpdate = await updateUserStatusOnNewTaskAssignment(userId);
+    return userStatusUpdate;
```
bhtibrewal, line 270:
```suggestion

  try {
   const userId = await getUserIdFromUserName(userName, usersCollection);
   const userStatusUpdate = await updateUserStatusOnNewTaskAssignment(userId);
    return userStatusUpdate;
```

models/userStatus.js
```diff
@@ -193,4 +205,133 @@ const updateAllUserStatus = async () => {
...
+        }
+        return updateCurrentStatusToState(userStatusModel, latestStatusData, userState.ACTIVE);
+      }
+    } else {
+      if (state === userState.OOO) {
+        return updateFutureStatusToState(userStatusModel, latestStatusData, userState.IDLE);
+      } else {
+        if (state === userState.IDLE) {
+          return generateAlreadyExistingStatusResponse(userState.IDLE);
+        }
```
bhtibrewal, line 317:
```suggestion
     switch (state){
  case userState.OOO:
  return updateFutureStatusToState(userStatusModel, latestStatusData, userState.ACTIVE);
}
```

models/userStatus.js
bhtibrewal: yes they are mutually exclusive and that will be maintained as you have an early return if one of them fulfills we will automatically not go to the other

heyrandhir: Done

bhtibrewal (approved):
LGTM, but this PR is very long please try to keep the PRs smaller

models/userStatus.js
```diff
@@ -193,4 +205,133 @@ const updateAllUserStatus = async () => {
...
+ *   message: string,
+ *   error: string,
+ * } | {
+ *   status: string,
+ *   message: string,
+ *   data: object
+ * }>} - The response object indicating the status of the user status update or an error.
+ */
+
+const updateUserStatusOnTaskUpdate = async (userName) => {
```
iamitprakash, line 266: this function is trying to update the user's status when a task is updated, but it is not checking if the user exists. If the user does not exist, the getUserIdFromUserName function will throw a NotFound error. However, the updateUserStatusOnTaskUpdate function will not handle this error correctly. It will simply return a 500 error, even though the user does not exist.

models/userStatus.js
```diff
@@ -193,4 +205,133 @@ const updateAllUserStatus = async () => {
...
+    }
+    return {
+      status: 500,
+      error: "Internal Server Error",
+      message: error.message,
+    };
+  }
+};
+
+const updateStatusOnTaskCompletion = async (userId) => {
```
iamitprakash, line 287: here we are trying to update the user's status when a task is completed, but it is not handling the case where the user does not have any active tasks. If the user does not have any active tasks, the function will try to update the user's status to IDLE, but the updateCurrentStatusToState function will throw an error because the user's current status is not OOO

iamitprakash (changes requested):

models/userStatus.js
heyrandhir: So if you see here  we have a single task assigned to user SooJK37gzjIZfFNH0tlL
image

On marking this task as completed. 
image

Since there were no other task assigned the user SooJK37gzjIZfFNH0tlL their status got updated to IDLE
image

controllers/tasks.js
```diff
@@ -17,8 +18,12 @@ const userQuery = require("../models/users");
  */
 const addNewTask = async (req, res) => {
   try {
+    // userStatusFlag is the Feature flag for status update based on task status. This flag is temporary and will be removed once the feature becomes stable.
+    const { userStatusFlag } = req.query;
```
bharati-21, line 22: Are we using dev query here or leveraging the feature flag service? Is the `true` denoting the `dev=true` feature in the UI?

bharati-21 (approved):
LGTM. 

This PR was long. Kindly break the PRs into smaller ones in the future.

controllers/tasks.js
heyrandhir: we are using the feature flag service. As the dev = true was already used. So had to create a new one. In the UI Changes are yet to be implemented.

iamitprakash (approved):

PR #1221, opened 2023-07-02
feat: includes archived users

krishanu2911:
## Before Changes
- before in API `/users`  only return archived users

### feature:
- create query param to include archived users
- query template `/users?q=includes%3Aarchived`
- with query param `/users?size=10&page=0&q=includes%3Aarchived` includes both archived true and false.

### DIFF
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -102,6 +102,9 @@ async function getUsers(req, res, next) {
           "string.empty": "size must contain value in range 1-100",
           "string.pattern.base": "size must be in range 1-100",
         }),
+      q: joi.string().optional().messages({
+        "string.empty": "q must not be empty and should have both a key and a value",
+      }),
       page: joi
         .string()
         .optional()
--- a/models/users.js
+++ b/models/users.js
@@ -136,7 +136,7 @@ const getSuggestedUsers = async (skill) => {
 
 /**
  * Fetches the data about our users
- * @param query { search, next, prev, size, page }: Filter for users
+ * @param query { search, next, prev, size, page, include }: Filter for users
  * @return {Promise<userModel|Array>}
  */
 const fetchPaginatedUsers = async (query) => {
@@ -145,9 +145,15 @@ const fetchPaginatedUsers = async (query) => {
     // INFO: https://github.com/Real-Dev-Squad/website-backend/pull/873#discussion_r1064229932
     const size = parseInt(query.size) || 100;
     const doc = (query.next || query.prev) && (await userModel.doc(query.next || query.prev).get());
-
-    let dbQuery = userModel.where("roles.archived", "==", false).orderBy("username");
-
+    let dbQuery;
+    if (query.q) {
+      const [queryName, queryValue] = query.q.split(":");
+      if (queryName === "includes" && queryValue === "archived") {
+        dbQuery = userModel;
+      }
+    } else {
+      dbQuery = userModel.where("roles.archived", "==", false).orderBy("username");
+    }
     if (query.prev) {
       dbQuery = dbQuery.limitToLast(size);
     } else {
--- a/test/integration/users.test.js
+++ b/test/integration/users.test.js
@@ -176,6 +176,7 @@ describe("Users", function () {
         .query({
           size: 1,
           page: 0,
+          q: "includes:archived",
         })
         .end((err, res) => {
           if (err) {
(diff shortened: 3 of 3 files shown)


middlewares/validators/user.js
```diff
@@ -102,6 +102,9 @@ async function getUsers(req, res, next) {
           "string.empty": "size must contain value in range 1-100",
           "string.pattern.base": "size must be in range 1-100",
         }),
+      q: joi.string().optional().messages({
+        "string.empty": "q must not be empty and should have both a key and a value",
```
harshith-venkatesh, line 106: Does q here refer to query?
Since this would be appearing in either logs or in user message, could we keep it more descriptive please

models/users.js
```diff
@@ -145,9 +145,15 @@ const fetchPaginatedUsers = async (query) => {
...
-
-    let dbQuery = userModel.where("roles.archived", "==", false).orderBy("username");
-
+    let dbQuery;
+    if (query.q) {
+      const [queryName, queryValue] = query.q.split(":");
+      if (queryName === "includes" && queryValue === "archived") {
+        dbQuery = userModel;
+      }
+    } else {
```
harshith-venkatesh, line 154: Is this a specific use case we are solving here, could we consider generalizing it.
We would receive different queries right in the future, could we think of a better way here?
I am also thinking about it, will update once I find relevant document

harshith-venkatesh (changes requested):

models/users.js
harshith-venkatesh: Have you looked in the repo, for the cases similar to this?

krishanu2911: So Ankush suggested this API structure go ahead with
image

Regarding the generalizing will make the changes

middlewares/validators/user.js
krishanu2911: Done Added the changes with `query` in the description.

models/users.js
krishanu2911: Done added the changes

harshith-venkatesh: please add test cases to validate with this params, when this params are absent as well, rest looks good

krishanu2911: have added the test case for this new scenario

harshith-venkatesh (dismissed):
Good work @krishanu2911 , I believe you had a great learning working on this PR. Good to see the constant communication and updates. GG

test/integration/users.test.js
```diff
@@ -169,13 +169,38 @@ describe("Users", function () {
...
+          }
+
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Users returned successfully!");
+          expect(res.body.users).to.be.a("array");
+          const userData = res.body.users;
+          userData.forEach((user) => {
+            expect(user.roles.archived).to.be.oneOf([false, true]);
+          });
```
heyrandhir, line 191: this doesn't guarantee that we are getting archived true users.  even if all users are not archived this will still pass.

heyrandhir (changes requested):
There are other validators for /users as well it wouldn't look good if we pass some query as ?query1=abc&q:includes:archived.

test/integration/users.test.js
krishanu2911: means if all users are archived false then it should also return those users too.
It will return a mix of both archived true and false.
**Expected behaviour**: will be when we pass this query we should get a mix of both archive true and false.
let me know if I missed something

krishanu2911:
> There are other validators for /users as well it wouldn't look good if we pass some query as ?query1=abc&q:includes:archived.

image
Let me know if this API structure needs to be changed.

test/integration/users.test.js
heyrandhir: but we should also ascertain that at least we are getting 1 archived user and 1 non archived users in the list

heyrandhir:
> > There are other validators for /users as well it wouldn't look good if we pass some query as ?query1=abc&q:includes:archived.
> 
> image Let me know if this API structure needs to be changed.

Fine but please confirm with @ankushdharkar once that we will be having query of both type in a single URL. Is that ok ?

heyrandhir (changes requested):
Also please update the tests for models and validator functions as well

test/integration/users.test.js
krishanu2911: Okay got it 👍🏻

krishanu2911: Added the test case for this case.

Issue #1196, opened 2023-06-23
We should create a data access layer, which will always return the user's filtered data


prakashchoudhary07:
We should create a data access layer, which will always return the user's filtered data/
So that if passing it through that code, will remove sensitive data

prakashchoudhary07:
Yes sure, will assign it to you
Thanks for picking this up

PR #1222, opened 2023-07-02
Create a data access layer, which always return the user's filtered data for `/users` route

mailmesriza98:
coveragereport
related to https://github.com/Real-Dev-Squad/website-backend/issues/1196

modifications for /users route


services/dataAccessLayer.js
```diff
@@ -0,0 +1,79 @@
...
+      if (!result.userExists) {
+        return res.boom.notFound("User doesn't exist");
+      }
+
+      const user = result.user;
+      delete user.phone;
+      delete user.email;
+      delete user.chaincode;
+      delete user.tokens;
+
```
Pratiyushkumar, line 34: i am not sure but if this can be done in one line then please do it

controllers/users.js
```diff
@@ -5,13 +5,12 @@ const logsQuery = require("../models/logs");
 const imageService = require("../services/imageService");
 const { profileDiffStatus } = require("../constants/profileDiff");
 const { logType } = require("../constants/logs");
-
+const dataAccess = require("../services/dataAccessLayer");
 const logger = require("../utils/logger");
 const obfuscate = require("../utils/obfuscate");
-const { getPaginationLink, getUsernamesFromPRs } = require("../utils/users");
-const { getQualifiers } = require("../utils/helper");
+// const { getUsernamesFromPRs } = require("../utils/users");
```
Pratiyushkumar, line 11: remove commented code please

controllers/users.js
```diff
@@ -5,13 +5,12 @@ const logsQuery = require("../models/logs");
...
 const { SOMETHING_WENT_WRONG, INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
-const { getFilteredPRsOrIssues } = require("../utils/pullRequests");
+// const { getFilteredPRsOrIssues } = require("../utils/pullRequests");
```
Pratiyushkumar, line 13: remove commented code please

services/dataAccessLayer.js
```diff
@@ -0,0 +1,79 @@
...
+        users,
+      });
+    }
+
+    const { allUsers, nextId, prevId } = await userQuery.fetchPaginatedUsers(req.query);
+    allUsers.forEach((element) => {
+      delete element.phone;
+      delete element.email;
+      delete element.chaincode;
+      delete element.tokens;
```
Pratiyushkumar, line 61: if possible here also please

services/dataAccessLayer.js
```diff
@@ -0,0 +1,78 @@
...
+        prev: prevId ? getPaginationLink(req.query, "prev", prevId) : "",
+      },
+    });
+  } catch (error) {
+    logger.error(`Error while fetching all users: ${error}`);
+    return res.boom.serverUnavailable(SOMETHING_WENT_WRONG);
+  }
+};
+
+const removeSensitiveInfo = function (obj, props) {
```
bksh05, line 68: A suggestion here

Can we have the function definition as given below?
This way, you don't have to pass an array every time you want to call the removeSensitiveInfo function, and it also gives you the flexibility to change parameters when required by passing an array as the second parameter.

```
function removeSensitiveInfo(data, properties = ["phone", "email", "chaincode", "tokens"]) {
  // Logic to remove sensitive information based on the properties array
  // ...
}
```

Example: 
```
 removeSensitiveInfo(element) // removes "phone", "email", "chaincode", and  "tokens"  as the default behavior
 removeSensitiveInfo(element, [token, chaincode])  // For cases where tokens and chaincode need to be removed.
```

bksh05 (commented):
Please add tests for your code as well.

iamitprakash (changes requested):
please add test coverage stats

services/dataAccessLayer.js
```diff
@@ -0,0 +1,78 @@
...
+        message: "User returned successfully!",
+        user,
+      });
+    }
+
+    if (qualifiers?.filterBy) {
+      const allPRs = await getFilteredPRsOrIssues(qualifiers);
+      const usernames = getUsernamesFromPRs(allPRs);
+      const { users } = await userQuery.fetchUsers(usernames);
+      removeSensitiveInfo(users, ["phone", "email", "chaincode", "tokens"]);
```
bksh05, line 42: Since there is the same array in the function's default parameter, you don't have to pass this array. 
You can remove it here and in other instances too.

bksh05 (changes requested):

services/dataAccessLayer.js
```diff
@@ -0,0 +1,78 @@
...
+        logger.error(`Error while fetching user: ${error}`);
+        return res.boom.serverUnavailable(SOMETHING_WENT_WRONG);
+      }
+
+      if (!result.userExists) {
+        return res.boom.notFound("User doesn't exist");
+      }
+
+      const user = result.user;
+      removeSensitiveInfo(user, ["phone", "email", "chaincode", "tokens"]);
```
bharati-21, line 30: Kindly add the array to a constant and re-use it.

services/dataAccessLayer.js
```diff
@@ -0,0 +1,78 @@
...
+  for (let i = 0; i < properties.length; i++) {
+    if (Object.prototype.hasOwnProperty.call(obj, properties[i])) {
+      delete obj[properties[i]];
+    }
+  }
+};
+
+module.exports = {
+  retrieveUsers,
+};
```
bharati-21, line 78: Was the file formatted? If not, kindly run prettier command to format. The formatting seems off.

bharati-21 (changes requested):
Kindly add test cases for the changes

services/dataAccessLayer.js
```diff
@@ -0,0 +1,34 @@
+const userQuery = require("../models/users");
+
+const retrieveUsers = async ({ id = null, usernames = null, req = null }) => {
```
iamitprakash, line 3: this function takes three arguments, but only two of them are actually used. The req argument is never used, so it could be removed from the function signature.

services/dataAccessLayer.js
```diff
@@ -0,0 +1,34 @@
...
+  } else {
+    const { allUsers, nextId, prevId } = await userQuery.fetchPaginatedUsers(req.query);
+    allUsers.forEach((element) => {
+      removeSensitiveInfo(element);
+    });
+    return { allUsers, nextId, prevId };
+  }
+};
+
+const removeSensitiveInfo = function (obj, properties = ["phone", "email", "chaincode", "tokens"]) {
```
iamitprakash, line 23: properties[i] property exists in the obj object. However, the hasOwnProperty method only checks if the property is defined in the object's prototype chain. This means that the hasOwnProperty method will not return true if the property is defined directly in the obj object.

To fix this bug, you can use the in operator instead of the hasOwnProperty method. The in operator checks if the property exists in the object, regardless of whether it is defined in the object's prototype chain

iamitprakash (changes requested):

services/dataAccessLayer.js
```diff
@@ -0,0 +1,34 @@
...
+    removeSensitiveInfo(result.user);
+    return result;
+  } else if (usernames) {
+    const { users } = await userQuery.fetchUsers(usernames);
+    users.forEach((element) => {
+      removeSensitiveInfo(element);
+    });
+    return users;
+  } else {
+    const { allUsers, nextId, prevId } = await userQuery.fetchPaginatedUsers(req.query);
```
bksh05, line 15: You should check if req is null or not. 
Let's say someone calls your function ` retrieveUsers({})`, then `req.query` will throw an error here. 

Also since you are just using the query, then why not pass only the query instead of the entire request object?

bksh05 (changes requested):

iamitprakash (dismissed):

bksh05 (approved):
LGTM

bksh05 (approved):

bharati-21 (dismissed):
LGTM

services/dataAccessLayer.js
```diff
@@ -0,0 +1,34 @@
...
+  } else {
+    const { allUsers, nextId, prevId } = await userQuery.fetchPaginatedUsers(query);
+    allUsers.forEach((element) => {
+      removeSensitiveInfo(element);
+    });
+    return { allUsers, nextId, prevId };
+  }
+};
+
+const removeSensitiveInfo = function (obj, properties = ["phone", "email", "chaincode", "tokens"]) {
```
prakashchoudhary07, line 23: Please move the properties to be removed to a constant

test/unit/services/dataAccessLayer.test.js
```diff
@@ -0,0 +1,62 @@
+const chai = require("chai");
+const chaiHttp = require("chai-http");
+
+const userQuery = require("../../../models/users");
+const sinon = require("sinon");
+const { retrieveUsers, removeSensitiveInfo } = require("../../../services/dataAccessLayer");
+const userData = require("../../fixtures/user/user")();
+const remove = require("../../fixtures/user/removalData")();
```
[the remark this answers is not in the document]
prakashchoudhary07: `remove`

don't know why other lines are showing

constants/users.js
```diff
@@ -16,4 +16,9 @@ const ALLOWED_FILTER_PARAMS = {
   ROLE: ["role"],
 };
 
-module.exports = { profileStatus, USER_STATUS, ALLOWED_FILTER_PARAMS };
+module.exports = {
+  profileStatus,
+  USER_STATUS,
+  ALLOWED_FILTER_PARAMS,
+  sensitiveData: ["phone", "email", "chaincode", "tokens"],
+};
```
Pratiyushkumar, line 24: can we not make this like
`const sensitiveData: ["phone", "email", "chaincode", "tokens"],`
and then export `sensitiveData`

test/unit/services/dataAccessLayer.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+  describe("removeSensitiveInfo", function () {
+    it("should remove sensitive information from the users object", function () {
+      removeSensitiveInfo(userData);
+      sensitiveData.forEach((key) => {
+        expect(userData[12]).to.not.have.property(key);
+      });
+    });
+  });
+});
```
Pratiyushkumar, line 62: can we make this in separate test file, let's not crowd one test file with multiple test suites

test/unit/services/dataAccessLayer.test.js
```diff
@@ -0,0 +1,62 @@
...
+const userData = require("../../fixtures/user/user")();
+const { sensitiveData } = require("../../../constants/users");
+chai.use(chaiHttp);
+const expect = chai.expect;
+
+describe("Data Access Layer", function () {
+  describe("retrieveUsers", function () {
+    it("should fetch a single user by ID and remove sensitive info", async function () {
+      const fetchUserStub = sinon.stub(userQuery, "fetchUser");
+      fetchUserStub.returns(Promise.resolve({ user: userData[12] }));
```
Pratiyushkumar, line 16: what will happen if multiple id's are passed

test/unit/services/dataAccessLayer.test.js
```diff
@@ -0,0 +1,62 @@
...
+      removeSensitiveInfo(userData[12]);
+      expect(result.user).to.deep.equal(userData[12]);
+      sensitiveData.forEach((key) => {
+        expect(userData[12]).to.not.have.property(key);
+      });
+    });
+
+    it("should fetch users by username and remove sensitive info", async function () {
+      const fetchUserStub = sinon.stub(userQuery, "fetchUsers");
+      fetchUserStub.returns(Promise.resolve({ users: [userData[12]] }));
```
Pratiyushkumar, line 28: if we try to fetch by user id, won't it be good enough because id's are always unique or username should be made unique and use it

test/unit/services/dataAccessLayer.test.js
Pratiyushkumar, line 1: can we also test if pass retrieveUsers(id=null, userName=null, query=null) then what will happen

test/unit/services/dataAccessLayer.test.js
mailmesriza98: multiple ids would not be passed in this case as per the function signature of retrieveUsers

Issue #1197, opened 2023-06-24
Remove github token from users collection


prakashchoudhary07:
GitHub token is not being used at any point in the backend.
There is use for it so we will not be storing it in the future

TODO:
- [ ] We should not store GitHub token when user's login's through github
- [ ] Remove GitHub token from user's collection [ Migration ]

PR #1227, opened 2023-07-04
Stop collecting github access token and remove it from existing user's data

fakhruddinkw:
The issue for this PR is: #1197 

**TASKS**

- [x] Stop collecting access token
- [x] Remove token from existing user's data
- [x] Tests written

**SCREENSHOTS**
result

**TEST STATS**
test-1
test-2
coverage-1
controller-coverage


controllers/users.js
```diff
@@ -607,6 +607,31 @@ const setInDiscordScript = async (req, res) => {
...
+    await Promise.all(allPromises);
+    return res.status(200).json({
+      message: "Github Token removed from all users!",
+      usersFoundWithToken: length,
+    });
+  } catch (err) {
+    return res.status(500).json({ message: INTERNAL_SERVER_ERROR });
+  }
+};
+
```
prakashchoudhary07, line 634: Can we use this here?
https://firebase.google.com/docs/firestore/manage-data/transactions

fakhruddinkw: I did read it. And i think Batch is best suitable for our use case here. Let me check  @prakashchoudhary07

heyrandhir (changes requested):

models/users.js
```diff
@@ -576,6 +576,33 @@ const fetchAllUsers = async () => {
...
+    throw err;
+  }
+};
+
+const removeGitHubToken = async (users) => {
+  try {
+    const batch = firestore.batch();
+    users.forEach((user) => {
+      batch.update(user, { tokens: admin.firestore.FieldValue.delete() });
+    });
```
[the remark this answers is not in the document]
fakhruddinkw: True. Either i need to run a separate batch if 500 is crossed, or i can use the good ol Promise.all @heyrandhir

controllers/users.js
```diff
@@ -607,6 +607,26 @@ const setInDiscordScript = async (req, res) => {
...
 };
 
+const removeGitHubTokenFromAllUsers = async (req, res) => {
+  try {
+    const users = await userQuery.usersWithGitHubToken();
+    const length = users.length;
+
+    if (length === 0) {
+      return res.status(500).json({ message: "No users found with github Token!" });
+    }
```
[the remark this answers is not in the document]
fakhruddinkw: What i thought was, after we have already ran this script once, tokens will be removed from all the users. Now, if again i try to run this script, it should give internal server error, because this action is not permitted now.
Because after you run the script once, there will be 0 users with tokens field.

models/users.js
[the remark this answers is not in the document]
prakashchoudhary07: +1 to Randhir's comment, please fix this

test/unit/models/users.test.js
```diff
@@ -220,4 +220,35 @@ describe("users", function () {
...
+    });
+
+    it("return array of users", async function () {
+      const data = await users.usersWithGitHubToken();
+      expect(data).to.have.length(7);
+    });
+    it('removes token field from user"s data', async function () {
+      const data = await users.usersWithGitHubToken();
+      await users.removeGitHubToken(data);
+    });
```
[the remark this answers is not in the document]
heyrandhir: yeah so if the field is undefined its deleted. Right ? thats what you should check that field is undefined after you call the model function. before calling assert the field is present.

heyrandhir (changes requested):

controllers/users.js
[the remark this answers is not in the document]
heyrandhir: > What i thought was, after we have already ran this script once, tokens will be removed from all the users. Now, if again i try to run this script, it should give internal server error, because this action is not permitted now. Because after you run the script once, there will be 0 users with tokens field.

then in that case return 404 with a appropriate message in the response . returning 500 will cause more confusion

controllers/users.js
```diff
@@ -607,6 +607,26 @@ const setInDiscordScript = async (req, res) => {
   }
 };
 
+const removeTokens = async (req, res) => {
+  try {
+    const users = await userQuery.fetchUsersWithToken();
+    const length = users.length;
+
+    if (length === 0) {
```
heyrandhir, line 615: Nit:
```js
    if (!users.length) {
```

controllers/users.js
```diff
@@ -607,6 +607,26 @@ const setInDiscordScript = async (req, res) => {
...
+      return res.status(404).json({ message: "No users found with github Token!" });
+    }
+
+    await userQuery.removeGitHubToken(users);
+
+    return res.status(200).json({
+      message: "Github Token removed from all users!",
+      usersFoundWithToken: length,
```
heyrandhir, line 623: I got a feedback today that we shouldn't use too long names hence if you want you can use `usersAffected` or `usersModified` or `usersUpdated` in place of `usersFoundWithToken`

heyrandhir: can also directly use users.length instead of new variable length

models/users.js
```diff
@@ -576,6 +576,52 @@ const fetchAllUsers = async () => {
...
+    throw err;
+  }
+};
+
+const removeGitHubToken = async (users) => {
+  try {
+    const length = users.length;
+
+    let numberOfBatches = length / 500;
+    numberOfBatches = numberOfBatches + 1;
```
heyrandhir, line 598: what if there are exactly multiple of 500 users found

models/users.js
```diff
@@ -576,6 +576,52 @@ const fetchAllUsers = async () => {
...
+
+    const batchArray = [];
+    for (let i = 0; i < numberOfBatches; i++) {
+      const batch = firestore.batch();
+      batchArray.push(batch);
+    }
+
+    let batchIndex = 0;
+    let operations = 0;
```
heyrandhir, line 607: Tejas has strictly mentioned that No PRs with `Injection Sink` warning will be merged  👀

test/integration/users.test.js
```diff
@@ -1229,4 +1229,31 @@ describe("Users", function () {
...
+        .request(app)
+        .patch("/users/remove-tokens")
+        .set("Cookie", `${cookieName}=${superUserAuthToken}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body.message).to.be.equal("Github Token removed from all users!");
+          return done();
```
heyrandhir, line 1255: can we also check the no of users count that you are returning

test/integration/users.test.js
heyrandhir, line 1: please confirm that you have tested this batch operation on local as well ?

heyrandhir (changes requested):

models/users.js
fakhruddinkw: Hmmm true. I ll need to check for remainder.
If there is remainder only then create one extra otherwise dont.

test/integration/users.test.js
fakhruddinkw: Yes i have ran this script on my local. It is working as intended

models/users.js
fakhruddinkw: @heyrandhir I was able to work around with the help of @TanishqSingla

fakhruddinkw: `somehow when we're accessing an object using a key, it takes it any type instead of number, and maybe that is causing it to throw this warning as the eslint doens't know if the key is a number or not` according to @TanishqSingla this is the problem

models/users.js
```diff
@@ -576,6 +576,56 @@ const fetchAllUsers = async () => {
...
+
+    for (let i = 0; i < length; i++) {
+      batchArray.at(parseInt(batchIndex)).update(users[+i], { tokens: admin.firestore.FieldValue.delete() });
+      operations++;
```
prakashchoudhary07, line 615: @fakhruddinkw , I Thought after discussion we agreed upon removing the parsing?
Am I missing something?

controllers/users.js
```diff
@@ -607,6 +607,25 @@ const setInDiscordScript = async (req, res) => {
...
+    }
+
+    await userQuery.removeGitHubToken(users);
+
+    return res.status(200).json({
+      message: "Github Token removed from all users!",
+      usersFound: users.length,
+    });
+  } catch (err) {
+    return res.status(500).json({ message: INTERNAL_SERVER_ERROR });
```
shubham-y, line 625: Please use `res.boom.badImplementation` here to maintain consistency

controllers/users.js
```diff
@@ -607,6 +607,25 @@ const setInDiscordScript = async (req, res) => {
...
+    if (!users.length) {
+      return res.status(404).json({ message: "No users found with github Token!" });
+    }
+
+    await userQuery.removeGitHubToken(users);
+
+    return res.status(200).json({
+      message: "Github Token removed from all users!",
+      usersFound: users.length,
+    });
```
shubham-y, line 623: We can return a summary of the migration in the response.
Like adding the total users data were migrated and failed number of users for which the data could not be updated if such a case happens.

test/integration/users.test.js
```diff
@@ -1229,4 +1229,31 @@ describe("Users", function () {
...
+      chai
+        .request(app)
+        .patch("/users/remove-tokens")
+        .set("Cookie", `${cookieName}=${superUserAuthToken}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body.message).to.be.equal("Github Token removed from all users!");
```
shubham-y, line 1254: I think we should run the `fetchUsersWithToken` model function after this and assert if the users array returned is empty or not to check if the migration was successful

fakhruddinkw: @shubham-y i dont think we can do await inside chai assertion

controllers/users.js
fakhruddinkw: @shubham-y as we are using batch for updates, it will be all or nothing , so i dont know if it will make sense to give metrics for failed updates

test/integration/users.test.js
shubham-y: > @shubham-y i dont think we can do await inside chai assertion

We can dot it. Check these for reference

https://github.com/Real-Dev-Squad/website-backend/blob/0c370f9aa903e8e4d6131e9a9a9a66eb45c690d4/test/integration/auth.test.js#L39-L41
https://github.com/Real-Dev-Squad/website-backend/blob/0c370f9aa903e8e4d6131e9a9a9a66eb45c690d4/test/unit/services/discordService.test.js#L24-L30

fakhruddinkw: @shubham-y I have made changes to the tests, please check

shubham-y (approved):

models/users.js
```diff
@@ -572,6 +572,56 @@ const fetchAllUsers = async () => {
...
+const removeGitHubToken = async (users) => {
+  try {
+    const length = users.length;
+
+    let numberOfBatches = length / 500;
+    const remainder = length % 500;
+
+    if (remainder) {
+      numberOfBatches = numberOfBatches + 1;
+    }
```
heyrandhir, line 598: Nit: 
```js
const numberOfBatches = Math.ceil(length / 500);
```

heyrandhir (changes requested):

routes/users.js
```diff
@@ -19,6 +19,8 @@ router.get("/self", authenticate, users.getSelfDetails);
 router.get("/isUsernameAvailable/:username", authenticate, users.getUsernameAvailabilty);
 router.get("/chaincode", authenticate, users.generateChaincode);
 router.get("/search", userValidator.validateUserQueryParams, users.filterUsers);
+router.patch("/remove-tokens", authenticate, authorizeRoles([SUPERUSER]), users.removeTokens);
```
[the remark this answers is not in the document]
fakhruddinkw: ok!

heyrandhir (approved):

routes/users.js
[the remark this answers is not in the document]
prakashchoudhary07: https://github.com/Real-Dev-Squad/website-backend/issues/1318

PR #1256, opened 2023-07-10
Remove orderBy from self tasks api

### LINKED ISSUE #1254: Fail tasks self api
Currently, self tasks API call uses orderBy to sort the query based on startedOn , but in some tasks data there is no field startedOn so orderBy is failing to get the tasks

image

vinit717:
Closes #1254

## Overview

Currently, self API of tasks is failing because there is no field of startedOn in some tasks so for the temporary solution it is been removed for a permanent solution the issue is mentioned below.

image

## Later work
- https://github.com/Real-Dev-Squad/website-backend/issues/1257

### DIFF
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -356,7 +356,7 @@ const fetchSkillLevelTask = async (userId) => {
  */
 const fetchSelfTasks = async (username) => {
   return await fetchUserTasks(username, []);
-  // removed startedOn as now task is not have field startedOn
+Removed `startedOn` field since we are getting issues with some of the documents in the tasks collection as some of the documents dont have `startedOn` present. 
 };
 
 /**
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -253,8 +253,7 @@ const fetchUserTasks = async (username, statuses = [], field, order) => {
           .where("participants", "array-contains", userId)
           .orderBy(field, order)
           .get();
-        featureTasksSnapshot = await tasksModel.where("assignee", "==", userId).get();
-        // removed orderBy(field, order) as now task is not have field startedOn
+        featureTasksSnapshot = await tasksModel.where("assignee", "==", userId).orderBy(field, order).get();
       } else {
         groupTasksSnapshot = await tasksModel.where("participants", "array-contains", userId).get();
         featureTasksSnapshot = await tasksModel.where("assignee", "==", userId).get();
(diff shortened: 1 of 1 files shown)


sahsisunny:
@vinit717 we have to also remove the indexes?

sahsisunny (dismissed):

isVivek99 (dismissed):

models/tasks.js
```diff
@@ -253,7 +253,8 @@ const fetchUserTasks = async (username, statuses = [], field, order) => {
           .where("participants", "array-contains", userId)
           .orderBy(field, order)
           .get();
-        featureTasksSnapshot = await tasksModel.where("assignee", "==", userId).orderBy(field, order).get();
+        featureTasksSnapshot = await tasksModel.where("assignee", "==", userId).get();
```
heyrandhir, line 256: rather than changing here I suggest we do it here https://github.com/Real-Dev-Squad/website-backend/blob/7e1fe072f031556f64b65320dbb53c4a769ba3da/models/tasks.js#L359. if we don't pass `field` it will go to the else loop

vinit717: done

models/tasks.js
```diff
@@ -355,7 +355,8 @@ const fetchSkillLevelTask = async (userId) => {
  * @returns {Promise<tasks>|Array}
  */
 const fetchSelfTasks = async (username) => {
-  return await fetchUserTasks(username, [], "startedOn", "desc");
+  return await fetchUserTasks(username, []);
+  // removed startedOn as now task is not have field startedOn
```
heyrandhir, line 359: Can you mention the Ticket No here in the comment as well. will be easy to track

heyrandhir (changes requested):

models/tasks.js
heyrandhir: https://github.com/Real-Dev-Squad/website-backend/issues/1257 & https://github.com/Real-Dev-Squad/website-backend/issues/1256

heyrandhir, line 359:
```suggestion
Removed `startedOn` field since we are getting issues with some of the documents in the tasks collection as some of the documents dont have `startedOn` present. 
```

heyrandhir (changes requested):

sahsisunny (dismissed):

models/tasks.js
vinit717: I had mentioned

heyrandhir (approved):

kotesh-arya (approved):

PR #1265, opened 2023-07-12
Changed status mismatched

vinit717:
**Overview**

- When a task is created by default its status is available and the assignee is false, So when the superuser assigns a task to the user its status is still available,

**Proposed Solution**

- When we passed the assignee as payload we can check if its status is available or not if available we can change it to assigned

image


TanishqSingla (dismissed):

models/tasks.js
```diff
@@ -19,6 +19,9 @@ const updateTask = async (taskData, taskId = null) => {
     taskData = await toFirestoreData(taskData);
     if (taskId) {
       const task = await tasksModel.doc(taskId).get();
+      if (taskData?.assignee && task.data().status === "AVAILABLE") {
+        taskData = { ...taskData, status: "ASSIGNED" };
```
sahsisunny, line 23: Can we use constants for this?

vinit717: done

sahsisunny (dismissed):

fakhruddinkw (approved):

sahsisunny (approved):

PR #1267, opened 2023-07-12
Check for valid dependency

vinit717:
**Overview**

- Currently, we can pass any string as dependency which may cause bad behavior So for valid dependency(taskid) we have to check that it exists or not  

**Solution**

- We can check as dependency by  checking doc id of task model exist or not if exist we can take as dependency otherwise throw an error

image


test/integration/tasks.test.js
```diff
@@ -328,6 +329,18 @@ describe("Tasks", function () {
 
       return taskId;
     });
+
+    it("Should return 400 if taskid is not exist", async function () {
+      taskId = (await tasks.updateTask(tasksData[5])).taskId;
```
heyrandhir, line 334: what if 1 task is valid and 1 in valid ?

models/tasks.js
```diff
@@ -33,15 +33,19 @@ const updateTask = async (taskData, taskId = null) => {
...
                 dependsOn: dependency,
               };
-              const docRef = dependencyModel.doc();
-              transaction.set(docRef, taskDependsOn);
+              if (dependencyDoc.exists) {
+                const docRef = dependencyModel.doc();
+                transaction.set(docRef, taskDependsOn);
+              } else {
+                throw new Error("Invalid dependency taskId");
+              }
```
heyrandhir, line 48: I don't see model tests getting updated. Please check

heyrandhir (changes requested):

test/unit/models/tasks.test.js
```diff
@@ -6,15 +6,15 @@
 
 const chai = require("chai");
 const { expect } = chai;
-
+// const sinon = require("sinon");
```
heyrandhir, line 9: please remove this if not needed

test/integration/tasks.test.js
```diff
@@ -311,15 +310,17 @@ describe("Tasks", function () {
     });
     it("Should update dependency", async function () {
       taskId = (await tasks.updateTask(tasksData[5])).taskId;
-      const dependsOn = ["taskId5", "taskId4"];
+      const taskId5 = (await tasks.updateTask(tasksData[5])).taskId;
+      const taskId6 = (await tasks.updateTask(tasksData[5])).taskId;
```
heyrandhir, line 314: why 5 and 6  in place of 1 and 2 🤔

test/integration/tasks.test.js
```diff
@@ -328,6 +329,18 @@ describe("Tasks", function () {
...
+    it("Should return 400 if taskid is not exist", async function () {
+      taskId = (await tasks.updateTask(tasksData[5])).taskId;
+
+      const dependsOn = ["taskId5", "taskId6"];
+      const res = await chai
+        .request(app)
+        .patch(`/tasks/${taskId}`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send({ dependsOn });
+      expect(res).to.have.status(400);
```
heyrandhir, line 342: can we also check the error message

models/tasks.js
vinit717: Added

test/unit/models/tasks.test.js
```diff
@@ -140,4 +145,31 @@ describe("tasks", function () {
...
+      await dependencyModel.doc("taskDependencies").set(data);
+      const dependencyData = (await dependencyModel.doc("taskDependencies").get()).data();
+
+      expect(dependencyData.dependsOn).to.be.a("array");
+      expect(dependencyData.taskId).to.be.equal("taskId1");
+
+      setStub.restore();
+      getStub.restore();
+    });
+  });
```
heyrandhir, line 174: where is the model function being called here

heyrandhir (changes requested):

test/unit/models/tasks.test.js
vinit717: Added

test/integration/tasks.test.js
vinit717: Added

test/integration/tasks.test.js
vinit717: Resolved

test/unit/models/tasks.test.js
vinit717: Removed

test/integration/tasks.test.js
vinit717: Done

heyrandhir (changes requested):

test/unit/models/tasks.test.js
```diff
@@ -140,4 +144,31 @@ describe("tasks", function () {
...
+      const data = {
+        taskId: "taskId1",
+        dependsOn: ["taskId2", "taskId3"],
+      };
+
+      const setStub = sinon.stub(dependencyModel.doc("taskDependencies"), "set").resolves();
+      const getStub = sinon.stub(dependencyModel.doc("taskDependencies"), "get").resolves({
+        data: () => data,
+      });
+      const result = await updateTask(data);
```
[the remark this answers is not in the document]
vinit717: Yes when we pass the update task it returns taskid this model function also used to create a task which return task object

controllers/tasks.js
```diff
@@ -205,8 +205,14 @@ const updateTask = async (req, res) => {
     if (isUserStatusEnabled && req.body.assignee) {
       await updateUserStatusOnTaskUpdate(req.body.assignee);
     }
+
     return res.status(204).send();
   } catch (err) {
+    if (err.message.includes("Invalid dependency taskId")) {
```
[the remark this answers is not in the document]
vinit717: done

test/unit/models/tasks.test.js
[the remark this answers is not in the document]
heyrandhir: `taskId` is string. Right ?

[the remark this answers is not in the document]
vinit717: yes also can you re-review the test I had changed the tests

heyrandhir (approved):

kotesh-arya (approved):

Issue #1216, opened 2023-07-01
Task percentage must be between 0% and 100%.


heyrandhir:
**Actual behavior:** The Task Percentage field accepts any numerical value and does not limit values to a maximum of 100%. 

**Expected behavior:** The Task Percentage field should only accept values from 0% to 100% and not allow percentages greater than 100% and only allow integer percentages.
  
**Reproduction steps:**
1. Open POSTMAN
2. Hit the PATCH call to `/tasks/:taskId` with the Request body having key percentCompleted with a value > 100 or < 0.

khalatevarun:
@heyrandhir i would like to work on this issue

PR #1269, opened 2023-07-12
feat: validate percentComplete value to be from 0 to 100

khalatevarun:
Fixes: 
#1216  

* Adds validation to percentCompleted param to accept value between 0 to 100
* Adds integration tests for the above case

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -233,10 +233,8 @@ const updateTaskStatus = async (req, res, next) => {
     if (task.taskData.status === TASK_STATUS.VERIFIED || req.body.status === TASK_STATUS.MERGED)
       return res.boom.forbidden("Status cannot be updated. Please contact admin.");
 
-    if (task.taskData.status === TASK_STATUS.COMPLETED && req.body.percentCompleted < 100) {
-      if (req.body.status === TASK_STATUS.COMPLETED || !req.body.status) {
-        return res.boom.badRequest("Task percent can't be updated as status is COMPLETED");
-      }
+    if (task.taskData.status === TASK_STATUS.COMPLETED) {
+      return res.boom.badRequest("Task details can't be updated as it is COMPLETED");
     }
 
     if (req.body.status === TASK_STATUS.COMPLETED && task.taskData.percentCompleted !== 100) {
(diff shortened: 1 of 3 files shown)


test/integration/tasks.test.js
```diff
@@ -632,7 +668,32 @@ describe("Tasks", function () {
...
+        endsOn: 1234,
+        startedOn: 4567,
+        status: "COMPLETED",
+        percentCompleted: 100,
+        participants: [],
+        assignee: appOwner.username,
+        completionAward: { [DINERO]: 3, [NEELAM]: 300 },
+        lossRate: { [DINERO]: 1 },
+        isNoteworthy: true,
+      };
```
heyrandhir, line 687: can we move this to fixture.

controllers/tasks.js
```diff
@@ -233,10 +233,8 @@ const updateTaskStatus = async (req, res, next) => {
     if (task.taskData.status === TASK_STATUS.VERIFIED || req.body.status === TASK_STATUS.MERGED)
       return res.boom.forbidden("Status cannot be updated. Please contact admin.");
 
-    if (task.taskData.status === TASK_STATUS.COMPLETED && req.body.percentCompleted < 100) {
-      if (req.body.status === TASK_STATUS.COMPLETED || !req.body.status) {
-        return res.boom.badRequest("Task percentCompleted can't updated as status is COMPLETED");
-      }
```
heyrandhir, line 239: why are we removing this condition !

controllers/tasks.js
```diff
@@ -233,10 +233,8 @@ const updateTaskStatus = async (req, res, next) => {
...
+    if (task.taskData.status === TASK_STATUS.COMPLETED) {
+      return res.boom.badRequest("Task details can't be updated as it is marked as COMPLETED");
```
heyrandhir, line 237: we should only check this if the request body has status. Maybe we would like to change the `completionAward` or `startDate` or some other parameter. if in the request body we receive new status and the current status is completed then we shouldn't allow. Also please check with @ankushdharkar once that whether is there any state after completed. Maybe `released` might come after completed. In that case we must allow `released`. just check with Ankush once.

heyrandhir (changes requested):

controllers/tasks.js
khalatevarun: changes reverted as discussed

controllers/tasks.js
khalatevarun: changes reverted as discussed

test/integration/tasks.test.js
khalatevarun: this test case has been reverted

heyrandhir (approved):

prakashchoudhary07 (approved):

Issue #1278, opened 2023-07-14
Clear cache related to tasks when there's an update


Aryex82:
- Clear cache for tasks whenever there's an update to tasks model.
- Increase cache time limit for GET calls of tasks to 10min

PR #1283, opened 2023-07-15
Add middleware to invalidate cache

Aryex82:
closes Issue: 

-  #1278 

Description:

1. Add a new data store to save all the keys generated for a particular cached API endpoint under a common key

>     eg: /tasks?page=2&size=9 / AND /tasks?page=10&size=10 will be saved as a set under a common key '__cache__ALL_TASKS'.

2. Add a middleware to evict cache data based on given common keys. 

>     eg: /tasks?page=2&size=9 / AND /tasks?page=10&size=10 cached 

3. Use the said middleware for when there's a write on tasks model
4. Increase expiry time of fetch all tasks api to 10mins.

Test coverage:
Screenshot 2023-07-16 at 6 48 12 PM

### DIFF
--- a/routes/tasks.js
+++ b/routes/tasks.js
@@ -6,16 +6,37 @@ const { createTask, updateTask, updateSelfTask, getTasksValidator } = require(".
 const authorizeRoles = require("../middlewares/authorizeRoles");
 const { APPOWNER, SUPERUSER } = require("../constants/roles");
 const assignTask = require("../middlewares/assignTask");
-const cache = require("../utils/cache");
+const { cache, invalidateCache } = require("../utils/cache");
+const { TASKS_ALL } = require("../constants/cacheKeys");
 
-router.get("/", getTasksValidator, cache(), tasks.fetchTasks);
+router.get("/", getTasksValidator, cache(TASKS_ALL, 10), tasks.fetchTasks);
 router.get("/self", authenticate, tasks.getSelfTasks);
 router.get("/overdue", authenticate, authorizeRoles([SUPERUSER]), tasks.overdueTasks);
-router.post("/", authenticate, authorizeRoles([APPOWNER, SUPERUSER]), createTask, tasks.addNewTask);
-router.patch("/:id", authenticate, authorizeRoles([APPOWNER, SUPERUSER]), updateTask, tasks.updateTask);
+router.post(
+  "/",
+  authenticate,
+  authorizeRoles([APPOWNER, SUPERUSER]),
+  invalidateCache([TASKS_ALL]),
+  createTask,
+  tasks.addNewTask
+);
+router.patch(
+  "/:id",
+  authenticate,
+  authorizeRoles([APPOWNER, SUPERUSER]),
+  invalidateCache([TASKS_ALL]),
+  updateTask,
+  tasks.updateTask
+);
 router.get("/:id/details", tasks.getTask);
 router.get("/:username", tasks.getUserTasks);
-router.patch("/self/:id", authenticate, updateSelfTask, tasks.updateTaskStatus, assignTask);
-router.patch("/assign/self", authenticate, tasks.assignTask);
-
+router.patch(
+  "/self/:id",
+  authenticate,
+  invalidateCache([TASKS_ALL]),
+  updateSelfTask,
+  tasks.updateTaskStatus,
+  assignTask
+);
+router.patch("/assign/self", authenticate, invalidateCache([TASKS_ALL]), tasks.assignTask);
 module.exports = router;
--- a/routes/tasks.js
+++ b/routes/tasks.js
@@ -6,17 +6,17 @@ const { createTask, updateTask, updateSelfTask, getTasksValidator } = require(".
 const authorizeRoles = require("../middlewares/authorizeRoles");
 const { APPOWNER, SUPERUSER } = require("../constants/roles");
 const assignTask = require("../middlewares/assignTask");
-const { cache, invalidateCache } = require("../utils/cache");
-const { TASKS_ALL } = require("../constants/cacheKeys");
+const { cacheResponse, invalidateCache } = require("../utils/cache");
+const { ALL_TASKS } = require("../constants/cacheKeys");
 
-router.get("/", getTasksValidator, cache({ invalidationKey: TASKS_ALL, expiry: 10 }), tasks.fetchTasks);
+router.get("/", getTasksValidator, cacheResponse({ invalidationKey: ALL_TASKS, expiry: 10 }), tasks.fetchTasks);
 router.get("/self", authenticate, tasks.getSelfTasks);
 router.get("/overdue", authenticate, authorizeRoles([SUPERUSER]), tasks.overdueTasks);
 router.post(
   "/",
   authenticate,
   authorizeRoles([APPOWNER, SUPERUSER]),
-  invalidateCache({ invalidationKeys: [TASKS_ALL] }),
+  invalidateCache({ invalidationKeys: [ALL_TASKS] }),
   createTask,
   tasks.addNewTask
 );
@@ -24,7 +24,7 @@ router.patch(
   "/:id",
   authenticate,
   authorizeRoles([APPOWNER, SUPERUSER]),
-  invalidateCache({ invalidationKeys: [TASKS_ALL] }),
+  invalidateCache({ invalidationKeys: [ALL_TASKS] }),
   updateTask,
   tasks.updateTask
 );
@@ -33,11 +33,11 @@ router.get("/:username", tasks.getUserTasks);
 router.patch(
   "/self/:id",
   authenticate,
-  invalidateCache({ invalidationKeys: [TASKS_ALL] }),
+  invalidateCache({ invalidationKeys: [ALL_TASKS] }),
   updateSelfTask,
   tasks.updateTaskStatus,
   assignTask
 );
-router.patch("/assign/self", authenticate, invalidateCache({ invalidationKeys: [TASKS_ALL] }), tasks.assignTask);
+router.patch("/assign/self", authenticate, invalidateCache({ invalidationKeys: [ALL_TASKS] }), tasks.assignTask);
 
 module.exports = router;
--- a/routes/tasks.js
+++ b/routes/tasks.js
@@ -9,14 +9,14 @@ const assignTask = require("../middlewares/assignTask");
 const { cache, invalidateCache } = require("../utils/cache");
 const { TASKS_ALL } = require("../constants/cacheKeys");
 
-router.get("/", getTasksValidator, cache(TASKS_ALL, 10), tasks.fetchTasks);
+router.get("/", getTasksValidator, cache({ invalidationKey: TASKS_ALL, expiry: 10 }), tasks.fetchTasks);
 router.get("/self", authenticate, tasks.getSelfTasks);
 router.get("/overdue", authenticate, authorizeRoles([SUPERUSER]), tasks.overdueTasks);
 router.post(
   "/",
   authenticate,
   authorizeRoles([APPOWNER, SUPERUSER]),
-  invalidateCache([TASKS_ALL]),
+  invalidateCache({ invalidationKeys: [TASKS_ALL] }),
   createTask,
   tasks.addNewTask
 );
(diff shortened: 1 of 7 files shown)


utils/cache.js
```diff
@@ -89,23 +138,63 @@ const cache = (data = { priority: 2, expiry: CACHE_EXPIRY_TIME_MIN }) => {
...
+  return "__cache__" + request.method + request.originalUrl;
+};
+
+/**
+ * Invalidate the cache for given keys
+ * @param {Object} [options]
+ * @param {Object} [options.invalidationKeys] Array of invalidation keys used in cache middleware.
+ * @returns {function} middleware function to help cache api response.
+ */
+const invalidateCache = (options = {}) => {
```
iamitprakash, line 178: here if the invalidationKeys property is an array. This means that if the property is not an array, the function will try to iterate over it as if it were an array. This will cause an error.

To fix the bug, the function should check if the invalidationKeys property is an array before trying to iterate over it.

Here is an example of how to fix the bug:

const invalidateCache = (options = {}) => {
  const keys = options.invalidationKeys;

  if (!Array.isArray(keys)) {
    throw new Error("Invalidation keys must be an array");
  }

  return async (req, res, next) => {
    try {
      for (const key of keys) {
        const cachedKeysList = cachedKeys.getCachedKeys(key);
        for (const ck of cachedKeysList) {
          pool.evict(ck);
        }
        cachedKeys.removeModelKey(key);
      }
    } catch (err) {
      logger.error(`Error while removing cached response ${err}`);
    } finally {
      next();
    }
  };
};

utils/cache.js
```diff
@@ -59,25 +69,64 @@ const cachePool = (opt = { maximumSize: CACHE_SIZE_MB }) => {
     }
   };
 
-  return { get, set, hits, cacheStore };
+  return { get, set, evict, hits, cacheStore };
+};
+
+const cachedKeysStore = () => {
```
iamitprakash, line 75: here we are not checking if the modelKey property already exists in the keyStore object. This means that if the modelKey property already exists in the keyStore object, the function will add the cachedKey property to the keyStore object twice.

iamitprakash (dismissed):

utils/cache.js
Aryex82: This is the function where I'm adding a cachedKey. Here, I am checking if there is a model key present and adding a cachedKey based on that. Please let me know if I got it wrong.


```javascript
 const addCachedKey = (modelKey, cachedKey) => {
     if (keyStore.has(modelKey)) {
       keyStore.get(modelKey).add(cachedKey);
     } else {
       const set = new Set();
       set.add(cachedKey);
       keyStore.set(modelKey, set);
     }
   };
``` 
PR link link: https://github.com/Real-Dev-Squad/website-backend/pull/1283/files#diff-5ceb95acc7e3f0b7328eba1584f1eb987b65f01cb9c396d74a733d3a6b0565b9R86

utils/cache.js
Aryex82: I have added the requested changes

iamitprakash (dismissed):

utils/cache.js
```diff
@@ -59,25 +69,64 @@ const cachePool = (opt = { maximumSize: CACHE_SIZE_MB }) => {
...
+  const keyStore = new Map();
+
+  const getCachedKeys = (modelKey) => {
+    if (keyStore.has(modelKey)) {
+      return [...keyStore.get(modelKey).keys()];
```
yesyash, line 80: - Are we storing a complex data type in keyStore or is it just going to be array of key strings? 
Can you please tell the structure of the value we're storing in keyStore.

routes/tasks.js
```diff
@@ -6,16 +6,38 @@ const { createTask, updateTask, updateSelfTask, getTasksValidator } = require(".
 const authorizeRoles = require("../middlewares/authorizeRoles");
 const { APPOWNER, SUPERUSER } = require("../constants/roles");
 const assignTask = require("../middlewares/assignTask");
-const cache = require("../utils/cache");
+const { cache, invalidateCache } = require("../utils/cache");
+const { TASKS_ALL } = require("../constants/cacheKeys");
 
-router.get("/", getTasksValidator, cache(), tasks.fetchTasks);
+router.get("/", getTasksValidator, cache({ invalidationKey: TASKS_ALL, expiry: 10 }), tasks.fetchTasks);
```
heyrandhir, line 12: can we rename cache to a more descriptive middleware such as `getCachedResponse ` or something on similar lines

routes/tasks.js
```diff
@@ -6,16 +6,38 @@ const { createTask, updateTask, updateSelfTask, getTasksValidator } = require(".
...
+  authorizeRoles([APPOWNER, SUPERUSER]),
+  invalidateCache({ invalidationKeys: [TASKS_ALL] }),
+  createTask,
+  tasks.addNewTask
+);
+router.patch(
+  "/:id",
+  authenticate,
+  authorizeRoles([APPOWNER, SUPERUSER]),
+  invalidateCache({ invalidationKeys: [TASKS_ALL] }),
```
heyrandhir, line 27: this object `{ invalidationKeys: [TASKS_ALL] }` is used at multiple places . can we make it a constant and use it in all routes

utils/cache.js
```diff
@@ -40,6 +41,15 @@ const cachePool = (opt = { maximumSize: CACHE_SIZE_MB }) => {
...
   };
 
+  /**
+   * Remove an API Response from cacheStore.
+   * @param {string} key
+   * @returns {boolean}
+   */
+  const evict = (key) => {
+    return cacheStore.delete(key);
+  };
```
heyrandhir, line 51: if this is just a single line function why are we extracting it ? it can be done in place

heyrandhir (changes requested):

utils/cache.js
Aryex82: This is just abstraction. cachePool had no exposed api to delete a key.

utils/cache.js
Aryex82: Although there won't be any significant performance improvement, changed the return type to Set.

routes/tasks.js
Aryex82: I agree it's too verbose, but if we have a situation in the future where only one API needs to invalidate some other key. It will lead to inconsistent code style.

utils/cache.js
```diff
@@ -59,25 +69,87 @@ const cachePool = (opt = { maximumSize: CACHE_SIZE_MB }) => {
...
+  /**
+   * Returns set of values mapped to given key
+   * @param {string} modelKey key for the map
+   * @returns {Set} set of values
+   */
+  const getCachedKeys = (modelKey) => {
+    if (keyStore.has(modelKey)) {
+      return keyStore.get(modelKey);
+    }
+    return new Set();
```
yesyash, line 90: - as we are returning an array inside the if condition should we also not return an empty array as the base case?

 - when doing a default return it's generally followed that we handle the false case first. 

Example:-
```js
if (!keys){
    ...do something 
}

<We can now handle our happy case>
```

What do you think?

utils/cache.js
```diff
@@ -59,25 +69,87 @@ const cachePool = (opt = { maximumSize: CACHE_SIZE_MB }) => {
...
 
-  return { get, set, hits, cacheStore };
+  return { get, set, evict, hits, cacheStore };
+};
+/**
+ * A MultiMap implementation where each key maps to set of unique values
+ *
+ */
+const cachedKeysStore = () => {
+  const keyStore = new Map();
```
yesyash, line 79: - let's add the type of `keyStore` as a jsdoc comment so that it helps others better maintain the code, what do you think?

yesyash (changes requested):

utils/cache.js
Aryex82: I agree with this for most cases, I think it's a good way to avoid unnecessary nesting in code. 
But in this context I don't feel that's the right approach.  As we are only interested in one case here and that's if the given key is present in the data store. 

for eg. this is how it would look if we handle the negative case first
```javascript
  const getCachedKeys = (modelKey) => {
     if (!keyStore.has(modelKey)) {
            return new Set();
     }
     // handle unnecessary edge cases where we should also return an empty set.
     
     if (keyStore.has(modelKey)) {      // still we cannot remove this check
       return keyStore.get(modelKey);
     }
   }
     
```

utils/cache.js
Aryex82: added.

utils/cache.js
```diff
@@ -59,25 +69,87 @@ const cachePool = (opt = { maximumSize: CACHE_SIZE_MB }) => {
     }
   };
 
-  return { get, set, hits, cacheStore };
+  return { get, set, evict, hits, cacheStore };
+};
+/**
+ * A MultiMap implementation where each key maps to set of unique values.
+ * It internally uses Map to store keys and values and to save multiple values it uses Set. Map<string,Set<string>>
```
yesyash, line 76: - why not just store it as an array of strings instead of set?

utils/cache.js
yesyash: Once you do this check and ensure the key is present

```js
if (!keyStore.has(modelKey)) {
        return new Set();
}
```

Why do we have to do this check again ? 

```js
if (keyStore.has(modelKey)) {
        ....
}
```

utils/cache.js
```diff
@@ -89,23 +161,65 @@ const cache = (data = { priority: 2, expiry: CACHE_EXPIRY_TIME_MIN }) => {
...
+  };
+};
+
+/**
+ * Generate key to save cache.
+ * @param {Object} [request]  HTTP request argument to the middleware function passed by express.
+ * @returns {string} cache key.
+ */
+const generateCacheKey = (request) => {
+  return "__cache__" + request.method + request.originalUrl;
```
shubham-y, line 192: Suggestion: As per key naming convention in Redis we use `:` as a separator. I think we can use the same as it will help make the code more readable and make the transition easier if in the future we decide to use Redis 

Reference - https://redis.io/docs/manual/keyspace/

utils/cache.js
Aryex82: Sets help improve the complexity of `addCachedKey` and `removeCachedKey` methods. By using array both would have O(n) complexities. 

Let me know if I got this wrong.

utils/cache.js
Aryex82: made changes to try follow the conventions mentioned there as close as possible

utils/cache.js
yesyash: - going through the code I was only able to see `.get`, `.delete` operations used for `keyStore` which is a map, but we are not using the set which is stored inside this map to push / pop any values to/from it.

```js
const key = keyStore.get(key)
key.add(value)    // Not doing this anywhere
key.delete(value)    // Not doing this anywhere
```

- If we're not using the set stored inside the `keyStore` map what benefit does it provide storing it as a set and not an array?

Can you please point to the line of code where we're using this set?

utils/cache.js
yesyash: - It is a coding style that I generally see in many places  where we handle the case of data not being present / incorrect / invalid this then makes sure that any operation being performed after the check has the correct data we want to use. 

A close example to this you would see in typescript repo's where we handle the `undefined`/`null` case at the top level of the function and have our happy path after that. 

This also make us think of the error cases before we start performing the actual operation.

utils/cache.js
Aryex82: I want the cached keys to be unique, so if we were using an array we would have to iterate through the array to check if an element is already present. Sets can do the same with better complexity right?

```javascript

keyStore.get(modelKey).add(cachedKey);

```
https://github.com/Real-Dev-Squad/website-backend/pull/1283/files#diff-5ceb95acc7e3f0b7328eba1584f1eb987b65f01cb9c396d74a733d3a6b0565b9R101


Also I plan to evict only a cached key in a future PR 
 so will be using 'removeCachedKey' there which does `key.delete(value)`

yesyash: Thanks for the info closing this.

utils/cache.js
Aryex82: added

shubham-y (approved):

heyrandhir (approved):

yesyash (approved):

Issue #1274, opened 2023-07-13
Refactor all occurrences of userQuery functions that returns user data to start consuming services/dataAccessLayer.js


mailmesriza98:
This will refactor the occurrences of userQuery model functions namely **_fetchUser, fetchUsers, getDiscordUsers, getUsersBasedOnFilter_** and some functions specific to member roles in all the controllers apart from the routes using these functions directly as the necessary routes are being handled in separate PRs to https://github.com/Real-Dev-Squad/website-backend/issues/1196, to use the service layer for querying user data

Parent issues: https://github.com/Real-Dev-Squad/website-backend/issues/1196

PR #1293, opened 2023-07-17
refactor all occurrences of fetchUser,fetchUsers

mailmesriza98:
closes https://github.com/Real-Dev-Squad/website-backend/issues/1274

251230948-c188936a-b556-43cd-967d-2e0846da0984


controllers/userStatus.js
```diff
@@ -1,8 +1,8 @@
-const { Forbidden, NotFound } = require("http-errors");
-const { fetchUser } = require("../models/users");
-const userStatusModel = require("../models/userStatus");
 const { getUserIdBasedOnRoute } = require("../utils/userStatus");
 const { INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
+const dataAccess = require("../services/dataAccessLayer");
+const { Forbidden, NotFound } = require("http-errors");
```
heyrandhir, line 4: NPM imports (external imports) in general ideally should be at top before the local imports.

controllers/users.js
```diff
@@ -212,14 +203,14 @@ const getUsernameAvailabilty = async (req, res) => {
  * @param res {Object} - Express response object
  */
 
-const getSelfDetails = (req, res) => {
+const getSelfDetails = async (req, res) => {
```
heyrandhir, line 206: since you are making this function `async` can you please confirm if the function callee is using await `keyword`.

middlewares/authenticate.js
```diff
@@ -79,13 +78,13 @@ module.exports = async (req, res, next) => {
...
         // add user data to `req.userData` for further use
-        req.userData = await users.fetchUser({ userId });
-
+        req.userData = await dataAccess.retrieveUsers({ id: userId });
         return checkRestricted(req, res, next);
       } else {
         return res.boom.unauthorized("Unauthenticated User");
       }
     } else {
+
```
heyrandhir, line 87: can you remove the trailing spaces here. Github is throwing warning 

image

heyrandhir (changes requested):

heyrandhir (changes requested):
Tests are failing please check.

controllers/contributions.js
```diff
@@ -11,7 +10,7 @@ const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
 const getUserContributions = async (req, res) => {
   try {
     const username = req.params.username;
```
DashDeipayan, line 12: `const {username} = req.params;`

DashDeipayan (changes requested):

DashDeipayan (dismissed):

iamitprakash (approved):

heyrandhir (approved):

prakashchoudhary07:
Please resolve  the merge conflicts

controllers/users.js
```diff
@@ -51,14 +50,7 @@ const getUserById = async (req, res) => {
...
 
-  const { phone = "", email = "", ...user } = result.user;
-  try {
-    user.phone = obfuscate.obfuscatePhone(phone);
-    user.email = obfuscate.obfuscateMail(email);
-  } catch (error) {
-    logger.error(`Error while formatting phone and email: ${error}`);
-    return res.boom.badImplementation("Error while formatting phone and email");
-  }
+  const user = result.user;
```
iamitprakash, line 53: are you sure result can't be null here any any edge case..?

services/dataAccessLayer.js
```diff
@@ -34,6 +46,22 @@ const retreiveFilteredUsers = async (query) => {
...
+
+const retrieveUsersWithRole = async (role) => {
+  const users = await members.fetchUsersWithRole(role);
+  users.forEach((element) => {
+    removeSensitiveInfo(element);
+  });
+  return users;
+};
+
 const removeSensitiveInfo = function (obj) {
```
iamitprakash, line 65: removeSensitiveInfo() function uses the in operator to check if a property exists on an object. However, the in operator is not safe against object injection attacks. This means that an attacker could inject a malicious object into the obj parameter, and the removeSensitiveInfo() function would not be able to detect it.

To fix this bug, you can use the hasOwnProperty() method instead of the in operator. The hasOwnProperty() method only returns true if the property exists on the object and was not inherited from a parent object. This makes it safe against object injection attacks.

iamitprakash (changes requested):

services/dataAccessLayer.js
heyrandhir: But even after this update the warning is still visible @mailmesriza98 @iamitprakash 
image

iamitprakash (approved):

bksh05 (approved):
LGTM

services/dataAccessLayer.js
```diff
@@ -37,9 +46,25 @@ const retreiveFilteredUsers = async (query) => {
...
   for (let i = 0; i < USER_SENSITIVE_DATA.length; i++) {
-    if (USER_SENSITIVE_DATA[i] in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, USER_SENSITIVE_DATA[i])) {
```
prakashchoudhary07, line 67: I think we don't need to access hasOwnProperty from prototype, it should be directly accessible

heyrandhir (approved):

PR #1299, opened 2023-07-18
Add prettier config

prakashchoudhary07:
Setup prettier config and formate command

Reason:
Windows users have issues when cloning the project, the end_of_line is set to clrf (windows), setting this up as configs can be read from prettierconfig and vs code will use those rules


prakashchoudhary07:
Closing this, need to revisit this again with new configs

Issue #1295, opened 2023-07-17
Updating idle users identifying script to consider active users 


heyrandhir:
In the batch script to update idle users consider updating the active users as well

PR #1307, opened 2023-07-19
Update Batch Script to include active users

heyrandhir:
Issue ticket 
- https://github.com/Real-Dev-Squad/website-backend/issues/1295
- https://github.com/Real-Dev-Squad/website-backend/issues/1315

The changes in this pull request (PR) update the code to handle batch updates for both idle and active users. Currently, there's a batch update script that identifies idle users and marks them as such, adding them to the pool of idle users. Now, we also want to remove users from the idle list if they become active. Therefore, besides identifying new idle users, this PR ensures that active users are correctly removed from the idle state.

Testing Stats:

Validators 
image

Models:
image

Controller:
image

### DIFF
--- a/middlewares/validators/userStatus.js
+++ b/middlewares/validators/userStatus.js
@@ -100,7 +100,7 @@ const validateMassUpdate = async (req, res, next) => {
         .items(
           Joi.object({
             userId: Joi.string().trim().required(),
-            expectedState: Joi.string().valid("IDLE", "ACTIVE").required(),
+            expectedState: Joi.string().valid(userState.IDLE, userState.ACTIVE).required(),
           })
         )
         .min(1)
--- a/middlewares/validators/userStatus.js
+++ b/middlewares/validators/userStatus.js
@@ -97,10 +97,15 @@ const validateMassUpdate = async (req, res, next) => {
   const schema = Joi.object()
     .keys({
       users: Joi.array()
-        .items(Joi.string().trim())
+        .items(
+          Joi.object({
+            userId: Joi.string().trim().required(),
+            expectedState: Joi.string().valid("IDLE", "ACTIVE").required(),
+          })
+        )
         .min(1)
         .required()
-        .error(new Error(`Invalid state value passed for users.`)),
+        .error(new Error(`Invalid object passed in users.`)),
     })
     .messages({
       "object.unknown": "Invalid key in Request payload.",
@@ -118,10 +123,7 @@ const validateMassUpdate = async (req, res, next) => {
 const validateGetQueryParams = async (req, res, next) => {
   const schema = Joi.object()
     .keys({
-      taskStatus: Joi.string()
-        .trim()
-        .valid(userState.IDLE)
-        .error(new Error(`Invalid state value passed for taskStatus.`)),
+      batch: Joi.boolean().valid(true).error(new Error(`Invalid boolean value passed for batch.`)),
       state: Joi.string()
         .trim()
         .valid(userState.IDLE, userState.ACTIVE, userState.OOO, userState.ONBOARDING)
--- a/middlewares/validators/userStatus.js
+++ b/middlewares/validators/userStatus.js
@@ -100,7 +100,7 @@ const validateMassUpdate = async (req, res, next) => {
         .items(
           Joi.object({
             userId: Joi.string().trim().required(),
-            expectedState: Joi.string().valid(userState.IDLE, userState.ACTIVE).required(),
+            state: Joi.string().valid(userState.IDLE, userState.ACTIVE).required(),
           })
         )
         .min(1)
@@ -123,7 +123,7 @@ const validateMassUpdate = async (req, res, next) => {
 const validateGetQueryParams = async (req, res, next) => {
   const schema = Joi.object()
     .keys({
-      batch: Joi.boolean().valid(true).error(new Error(`Invalid boolean value passed for batch.`)),
+      aggregate: Joi.boolean().valid(true).error(new Error(`Invalid boolean value passed for aggregate.`)),
       state: Joi.string()
         .trim()
         .valid(userState.IDLE, userState.ACTIVE, userState.OOO, userState.ONBOARDING)
--- a/models/userStatus.js
+++ b/models/userStatus.js
@@ -347,7 +347,7 @@ const batchUpdateUsersStatus = async (users) => {
       latestStatusData = await getUserStatus(userId);
     } catch (error) {
       summary.totalUnprocessedUsers++;
-      return batch;
+      continue;
     }
     const { id, userStatusExists, data } = latestStatusData;
     const statusToUpdate = {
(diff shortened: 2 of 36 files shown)


models/userStatus.js
```diff
@@ -330,98 +330,108 @@ const updateStatusOnTaskCompletion = async (userId) => {
 const massUpdateIdleUsers = async (users) => {
```
vinit717, line 330: why massUpdateIdleUsers can we do only updateIdleUsers

models/userStatus.js
```diff
@@ -330,98 +330,108 @@ const updateStatusOnTaskCompletion = async (userId) => {
...
-                    from: currentTimeStamp,
-                    until: "",
-                    updatedAt: currentTimeStamp,
-                  },
+                  currentStatus: statusToUpdate,
                 };
           batch.update(docRef, updatedStatusData);
-        } else {
-          usersWithStatusUnchanged++;
         }
```
vinit717, line 402: too much nesting is there can you clean it little bit

middlewares/validators/userStatus.js
```diff
@@ -97,10 +97,15 @@ const validateMassUpdate = async (req, res, next) => {
   const schema = Joi.object()
     .keys({
       users: Joi.array()
-        .items(Joi.string().trim())
+        .items(
+          Joi.object({
+            userId: Joi.string().trim().required(),
+            expectedState: Joi.string().valid("IDLE", "ACTIVE").required(),
```
apurvkhare, line 103: Instead of hardcoding `IDLE` and`ACTIVE`, use the constants present in `constants/userStatus.js`

models/userStatus.js
heyrandhir: Since now we are updating both idle and active users. I updated the name to batchUpdateUsersStatus as updateUserStatus already existed.

apurvkhare (dismissed):

test/integration/taskBasedStatusUpdate.test.js
```diff
@@ -287,31 +287,53 @@ describe("Task Based Status Updates", function () {
...
-      superUserId = await addUser(userData[4]);
-      superUserJwt = authService.generateAuthToken({ userId: superUserId });
-
-      userId1 = await addUser(userData[6]);
-      userId2 = await addUser(userData[8]);
-      userId3 = await addUser(userData[9]);
-      userId4 = await addUser(userData[0]);
-      userId5 = await addUser(userData[1]);
-      listUsers = [userId1, userId2, userId3, userId4, userId5];
+      userId0 = await addUser(userData[0]);
```
shreya-mishra, line 305: what's this id0,1.....
you could have store it in an array ?

test/unit/models/taskBasedStatusUpdate.test.js
```diff
@@ -226,25 +226,50 @@ describe("Update Status based on task update", function () {
   });
 
   describe("Test the Model Function for Changing the status to IDLE based on users list passed", function () {
+    let userId0;
```
shreya-mishra, line 229: +1

shreya-mishra (commented):
Please follow DRY concept !

middlewares/validators/userStatus.js
```diff
@@ -118,10 +123,7 @@ const validateMassUpdate = async (req, res, next) => {
 const validateGetQueryParams = async (req, res, next) => {
   const schema = Joi.object()
     .keys({
-      taskStatus: Joi.string()
-        .trim()
-        .valid(userState.IDLE)
-        .error(new Error(`Invalid state value passed for taskStatus.`)),
+      batch: Joi.boolean().valid(true).error(new Error(`Invalid boolean value passed for batch.`)),
```
bharati-21, line 126: When passed as query params `batch` will be of type string. Did you test this if the type is being converted?

models/userStatus.js
```diff
@@ -327,101 +327,111 @@ const updateStatusOnTaskCompletion = async (userId) => {
...
-        const {
-          currentStatus: { state, until },
-        } = data;
+  for (const { userId, expectedState } of users) {
+    let latestStatusData;
+    try {
+      latestStatusData = await getUserStatus(userId);
+    } catch (error) {
+      summary.totalUnprocessedUsers++;
+      return batch;
```
bharati-21, line 350: Why not return error? Why is batch being returned here?

middlewares/validators/userStatus.js
heyrandhir: Yes it gets auto converted. I tested with other strings it fails. for boolean it works as expected.

models/userStatus.js
heyrandhir: Yes this was not needed. I removed it now. earlier I was using map so I needed return statement. Now I am no longer using map so I don't need return. Hence removed.

test/integration/taskBasedStatusUpdate.test.js
heyrandhir: Implemented.

test/unit/models/taskBasedStatusUpdate.test.js
heyrandhir: Done.

models/userStatus.js
bharati-21: Why are we continuing also?

bharati-21: I understood. Resolving this.

bharati-21 (approved):
LGTM

models/userStatus.js
heyrandhir: Thanks

shreya-mishra (dismissed):

shreya-mishra (approved):

vinit717 (dismissed):

apurvkhare (approved):

iamitprakash (approved):

models/userStatus.js
```diff
@@ -327,101 +327,111 @@ const updateStatusOnTaskCompletion = async (userId) => {
...
+      batch.set(newUserStatusRef, newUserStatusData);
+    } else {
+      const {
+        currentStatus: { state: currentState, until },
+      } = data;
+      if (currentState === state) {
+        currentState === userState.ACTIVE ? summary.totalActiveUsersUnAltered++ : summary.totalIdleUsersUnAltered++;
+        continue;
+      }
+      if (currentState === userState.ONBOARDING) {
```
iamitprakash, line 377: here if the current status of the user is ONBOARDING. However, the currentStatus property of the data object is only defined if the user already has a user status. If the user does not have a user status, then the currentStatus property will be undefined.

This means that the if statement will always be true if the user does not have a user status. This will cause the code to update the user status of the user, even if the user does not have a user status.

iamitprakash (changes requested):

models/userStatus.js
heyrandhir: > here if the current status of the user is ONBOARDING. However, the currentStatus property of the data object is only defined if the user already has a user status. If the user does not have a user status, then the currentStatus property will be undefined.
> 
> This means that the if statement will always be true if the user does not have a user status. This will cause the code to update the user status of the user, even if the user does not have a user status.

`here if the current status of the user is ONBOARDING. However, the currentStatus property of the data object is only defined if the user already has a user status. If the user does not have a user status` - This statement is self contradictory. If the user is onboarding then `user does not have a user status.` cant be true. The current status can be either onboarding , idle , active or OOO. currently any user signing up for rds their status gets created automatically and assigned onboarding.

iamitprakash (approved):

PR #1309, opened 2023-07-20
Added member count and creator details in the GET discord-actions/groups API.

bksh05:
## Description 
This pull request adds group creator details, including first name, last name, and profile picture, which will be displayed in the dashboard UI. Additionally, a new field called "count" is introduced to indicate the number of members in the group. These changes are expected to enhance the dashboard user interface and provide more comprehensive information.

Fixes: #1382 

### API response before changes
``` 
{
  "message": "Roles fetched successfully!",
  "groups": [
    {
      "id": "15PPEWydQ84YlK6MlK6z",
      "date": {
        "_seconds": 1689451955,
        "_nanoseconds": 807000000
      },
      "createdBy": "CvUmCEm6vB9pczLBli0z",
      "rolename": "group-Too-Much-Truth",
      "roleid": "1129868137827938324"
    }
  ]
}
```

### API response after changes

```
{
  "message": "Roles fetched successfully!",
  "groups": [
    {
      "id": "mExFFXEJo2YL8XBbdYJ1",
      "date": {
        "_seconds": 1689838197,
        "_nanoseconds": 139000000
      },
      "rolename": "group-developers",
      "roleid": "1131488158979194902",
      "createdBy": "eCITJh08tFpYvq4SDa1d",
      "memberCount": 1,
      "firstName": "Test",
      "lastName": "User1",
      "image": "https://images.unsplash.com/photo-1633332755192-727a05c4013d?ixlib=rb-4.0.3&ixid=M3wxMjA3fDB8MHxzZWFyY2h8Mnx8dXNlcnxlbnwwfHwwfHx8MA%3D%3D&auto=format&fit=crop&w=500&q=60"
    }
  ]
}

```

## Coverage

File                                    | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
----------------------------------------|---------|----------|---------|---------|---------------------------------------------------
 models/discordactions.js                     |   81.33 |     92.3 |   91.66 |   81.69 | 19-20,42-43,57-59,63-64,88-89,142-143    
 models/users.js                              |   43.15 |    26.16 |      32 |   44.76 | ...19,337-338,394-401,411-416,420-432,437-439,461-540,550-564,596-597
services/dataAccessLayer.js                  |   97.22 |    94.73 |     100 |   97.14 | 36  
controllers/discordactions.js                     |   49.12 |        0 |      50 |   49.12 | 18-57,90-91,102-137
  
## Follow-up issue
Add details regarding already joined groups https://github.com/Real-Dev-Squad/website-dashboard/issues/424

### DIFF
--- a/controllers/discordactions.js
+++ b/controllers/discordactions.js
@@ -74,11 +74,12 @@ const getAllGroupRoles = async (req, res) => {
     }, new Set());
     const groupCreatorsDetails = await retrieveUsers({ userIds: Array.from(groupCreatorIds) });
     const groupsWithUserDetails = groupsWithMemberCount.map((group) => {
+      const groupCreator = groupCreatorsDetails[group.createdBy];
       return {
         ...group,
-        firstName: groupCreatorsDetails[group.createdBy].first_name,
-        lastName: groupCreatorsDetails[group.createdBy].last_name,
-        image: groupCreatorsDetails[group.createdBy].picture?.url,
+        firstName: groupCreator.first_name,
+        lastName: groupCreator.last_name,
+        image: groupCreator.picture?.url,
       };
     });
 
--- a/controllers/discordactions.js
+++ b/controllers/discordactions.js
@@ -3,6 +3,7 @@ const admin = require("firebase-admin");
 const config = require("config");
 const jwt = require("jsonwebtoken");
 const discordRolesModel = require("../models/discordactions");
+const { retrieveUsers } = require("../services/dataAccessLayer");
 
 /**
  * Creates a role
@@ -66,9 +67,24 @@ const createGroupRole = async (req, res) => {
 const getAllGroupRoles = async (req, res) => {
   try {
     const { groups } = await discordRolesModel.getAllGroupRoles();
+    const groupsWithMemberCount = await discordRolesModel.getNumberOfMemberForGroups(groups);
+    const groupCreatorIds = groupsWithMemberCount.reduce((ids, group) => {
+      ids.add(group.createdBy);
+      return ids;
+    }, new Set());
+    const groupCreatorsDetails = await retrieveUsers({ userIds: Array.from(groupCreatorIds) });
+    const groupsWithUserDetails = groupsWithMemberCount.map((group) => {
+      return {
+        ...group,
+        firstName: groupCreatorsDetails[group.createdBy].first_name,
+        lastName: groupCreatorsDetails[group.createdBy].last_name,
+        image: groupCreatorsDetails[group.createdBy].picture?.url,
+      };
+    });
+
     return res.json({
       message: "Roles fetched successfully!",
-      groups,
+      groups: groupsWithUserDetails,
     });
   } catch (err) {
     logger.error(`Error while getting roles: ${err}`);
--- a/models/discordactions.js
+++ b/models/discordactions.js
@@ -114,10 +114,41 @@ const updateDiscordImageForVerification = async (userDiscordId) => {
   }
 };
 
+const getNumberOfMemberForGroups = async (groups = []) => {
+  try {
+    if (!groups.length) {
+      return [];
+    }
+    const roleIds = groups.map((group) => group.roleid);
+
+    const snapshots = await memberRoleModel.where("roleid", "in", roleIds).get();
+    const roleCount = {};
+
+    snapshots.forEach((doc) => {
+      const roleToMemberMapping = doc.data();
+
+      if (roleCount[roleToMemberMapping.roleid]) {
+        roleCount[roleToMemberMapping.roleid] += 1;
+      } else {
+        roleCount[roleToMemberMapping.roleid] = 1;
+      }
+    });
+
+    return groups.map((group) => ({
+      ...group,
+      count: roleCount[group.roleid] || 0,
+    }));
+  } catch (err) {
+    logger.error("Error while counting members for each group", err);
+    throw err;
+  }
+};
+
 module.exports = {
   createNewRole,
   getAllGroupRoles,
   addGroupRoleToMember,
   isGroupRoleExists,
   updateDiscordImageForVerification,
+  getNumberOfMemberForGroups,
 };
--- a/models/discordactions.js
+++ b/models/discordactions.js
@@ -136,7 +136,7 @@ const getNumberOfMemberForGroups = async (groups = []) => {
 
     return groups.map((group) => ({
       ...group,
-      count: roleCount[group.roleid] || 0,
+      memberCount: roleCount[group.roleid] || 0,
     }));
   } catch (err) {
     logger.error("Error while counting members for each group", err);
--- a/models/users.js
+++ b/models/users.js
@@ -582,6 +582,7 @@ const fetchUsersWithToken = async () => {
     return users;
   } catch (err) {
     logger.error(`Error while fetching all users with tokens field: ${err}`);
+    return [];
   }
 };
 /**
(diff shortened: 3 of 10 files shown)


controllers/discordactions.js
```diff
@@ -66,9 +67,24 @@ const createGroupRole = async (req, res) => {
...
+    const groupsWithMemberCount = await discordRolesModel.getNumberOfMemberForGroups(groups);
+    const groupCreatorIds = groupsWithMemberCount.reduce((ids, group) => {
+      ids.add(group.createdBy);
+      return ids;
+    }, new Set());
+    const groupCreatorsDetails = await retrieveUsers({ userIds: Array.from(groupCreatorIds) });
+    const groupsWithUserDetails = groupsWithMemberCount.map((group) => {
+      return {
+        ...group,
+        firstName: groupCreatorsDetails[group.createdBy].first_name,
```
apurvkhare, line 79: since ` groupCreatorsDetails[group.createdBy]` is being used in multiple places then it can be stored in a const.
It would look clean.
```
createdBy = groupCreatorsDetails[group.createdBy]
```

bksh05: Yes. That's a good catch.
I somehow missed it.

Thank you

controllers/discordactions.js
```diff
@@ -66,9 +67,25 @@ const createGroupRole = async (req, res) => {
 const getAllGroupRoles = async (req, res) => {
```
iamitprakash, line 67: here retrieveUsers function is expecting an array of userIds as input, but the code is passing a Set of userIds. This will cause the retrieveUsers function to throw an error.

To fix the bug, you need to change the line of code that calls the retrieveUsers function to the following:

const groupCreatorsDetails = await retrieveUsers({ userIds: Array.from(groupCreatorIds) });
This will convert the Set of userIds to an array, which is the correct format for the retrieveUsers function.

models/users.js
```diff
@@ -572,6 +572,32 @@ const fetchAllUsers = async () => {
...
+ * @returns Object containing the details of the users whose userId was provided.
+ */
+const fetchUserByIds = async (userIds = []) => {
+  if (userIds.length === 0) {
+    return {};
+  }
+  try {
+    const users = {};
+    const usersRefs = userIds.map((docId) => userModel.doc(docId));
+    const documents = await firestore.getAll(...usersRefs);
```
iamitprakash, line 587: here in the line of code where you are getting the user data from Firestore. The code is:

const documents = await firestore.getAll(...usersRefs);
The problem with this code is that it is using the getAll() method to get all of the documents in the usersRefs array. However, the getAll() method can be very expensive, especially if there are a lot of documents in the array.

To fix this bug, you can use the get() method instead of the getAll() method. The get() method will only get the document that you specify, so it will be much faster.

iamitprakash (changes requested):

models/users.js
bksh05: Regarding the getAll method, it fetches only the details specified in the parameters, as documented [here](https://googleapis.dev/nodejs/firestore/latest/Firestore.html#getAll).

Additionally, it's important to note that there is no get method available directly on the Firestore object.

If you were referring to using the get method of models, that won't work in this case because it requires conditions using where, and the where clause works with document keys (e.g., discordId in [...]).

Since the requirement is to fetch user data with document IDs, using getAll seems to be the appropriate approach in this scenario.

Note: this is my first time working with Firestore, so I might have limited knowledge. If you have any other suggestions or more efficient ways to handle this, I'm open to learning and making improvements.

Thank you for your feedback and patience!

models/discordactions.js
```diff
@@ -114,10 +114,41 @@ const updateDiscordImageForVerification = async (userDiscordId) => {
...
+      if (roleCount[roleToMemberMapping.roleid]) {
+        roleCount[roleToMemberMapping.roleid] += 1;
+      } else {
+        roleCount[roleToMemberMapping.roleid] = 1;
+      }
+    });
+
+    return groups.map((group) => ({
+      ...group,
+      count: roleCount[group.roleid] || 0,
```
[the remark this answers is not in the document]
bksh05: It's for a case where roleCount[group.roleId] is undefined.
Let's say there is a new role with no members. Then in memberRoleModel will not have any entry for the new role. 
In that case, roleCount[group.roleid] will be undefined.

models/discordactions.js
```diff
@@ -114,10 +114,41 @@ const updateDiscordImageForVerification = async (userDiscordId) => {
...
+    const roleCount = {};
+
+    snapshots.forEach((doc) => {
+      const roleToMemberMapping = doc.data();
+
+      if (roleCount[roleToMemberMapping.roleid]) {
+        roleCount[roleToMemberMapping.roleid] += 1;
+      } else {
+        roleCount[roleToMemberMapping.roleid] = 1;
+      }
```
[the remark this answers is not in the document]
bksh05: After considering both approaches, I believe the choice between using the ternary operator and the if-else statement here is a matter of personal preference. In my opinion, my approach may result in better readability, even though it requires more lines of code. While your approach is shorter and concise, it may slightly compromise readability.

For now, I will stick with my current implementation. However, if you strongly disagree and would like to discuss this further, please let me know, and we can have a discussion about it.

Thank you.

iamitprakash (approved):

iamitprakash:
Other PR related details are missing ....

bksh05:
> Other PR related details are missing ....

I have added the "Follow-up Issue" section to the Pull Request (PR), assuming that's what you were referring to.

controllers/discordactions.js
```diff
@@ -66,9 +67,25 @@ const createGroupRole = async (req, res) => {
...
+      ids.add(group.createdBy);
+      return ids;
+    }, new Set());
+    const groupCreatorsDetails = await retrieveUsers({ userIds: Array.from(groupCreatorIds) });
+    const groupsWithUserDetails = groupsWithMemberCount.map((group) => {
+      const groupCreator = groupCreatorsDetails[group.createdBy];
+      return {
+        ...group,
+        firstName: groupCreator.first_name,
+        lastName: groupCreator.last_name,
```
RitikJaiswal75, line 81: I dont think we need this here only user id is enough.

bksh05: I need this to show first and last names in UI (as expected in the issue). 
It's necessary for that. 

Instead of passing the user id and calling another API with all the userIDs to get all the details. It's better to add it here in a single API as the information is not much and this API gives all the collective details required for the UI to work.

bksh05: Resolving this for review. Let me know if you think otherwise and we can have a discussion on the API structure. 
Thank you for the review.

models/discordactions.js
```diff
@@ -114,10 +114,41 @@ const updateDiscordImageForVerification = async (userDiscordId) => {
...
+    });
```
RitikJaiswal75, line 135: why do we need this the query snapshot returned by Firestore is already size, we can just query based on single role id and create a map

bksh05: Can you explain in more detail? I am not able to understand what you are trying to say here.

bksh05: I am following the below documentation for where with "in" condition. 
https://firebase.google.com/docs/firestore/query-data/queries#in_not-in_and_array-contains-any

And then going through each array to check which ids are present and how many times.

bksh05: Resolving this conversation in favor of the above comments. 
Please reopen if it feels there is something we can improve on this.

Thanks for reviewing.

yesyash (approved):

iamitprakash (approved):

PR #1319, opened 2023-07-22
Feat: progress handling for non working days ( sundays )

khalatevarun:
…nt on sundays

### DIFF
--- a/controllers/progresses.js
+++ b/controllers/progresses.js
@@ -48,6 +48,12 @@ const createProgress = async (req, res) => {
     body: { type },
   } = req;
   try {
+    const date = new Date();
+    if (date.getDay() === 0) {
+      return res.status(500).json({
+        message: "Progress document cannot be created on a Sunday",
+      });
+    }
     const data = await createProgressDocument({ ...req.body, userId: req.userData.id });
     return res.status(201).json({
       data,
--- a/utils/progresses.js
+++ b/utils/progresses.js
@@ -172,7 +172,10 @@ const getProgressRecords = async (query, queryParams) => {
   const currentDate = new Date(startDate);
   while (currentDate <= new Date(endDate)) {
     const date = currentDate.toISOString().slice(0, 10);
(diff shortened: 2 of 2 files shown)


controllers/progresses.js
```diff
@@ -48,6 +52,12 @@ const createProgress = async (req, res) => {
     body: { type },
   } = req;
   try {
+    const date = new Date();
+    if (date.getDay() === 0) {
+      return res.status(400).json({
+        message: PROGRESS_DOCUMENT_NON_WORKING_DAYS,
+      });
+    }
```
heyrandhir, line 60: can you try to handle it in validator. As validation checks are majorly handled in validator file.

heyrandhir (changes requested):

controllers/progresses.js
khalatevarun: done, thanks for pointing it out

middlewares/validators/progresses.js
```diff
@@ -1,7 +1,14 @@
...
+const { VALID_PROGRESS_TYPES, RESPONSE_MESSAGES } = require("../../constants/progresses");
 
 const validateCreateProgressRecords = async (req, res, next) => {
+  // Check if the request is being made on a Sunday
+  const today = new Date();
+  if (today.getDay() === 0) {
+    // Sunday corresponds to day index 0
+    res.boom.badRequest(RESPONSE_MESSAGES.PROGRESS_DOCUMENT_NON_WORKING_DAYS);
+  }
+
```
heyrandhir, line 11: there is a catch here @khalatevarun . Actually on Sundays too until 6 am IST  we will allow users to update. As for the current day users can submit their standup up till next day 6 am. Similarly someone can send sunday standup on Monday 5 am. This logic will miss both these cases.

heyrandhir (changes requested):

middlewares/validators/progresses.js
khalatevarun: got it, will add support to these two cases

Issue #1310, opened 2023-07-21
modify status change on task status update


heyrandhir:
Include other status such as needs review, in review for marking users idle. Currently this is happening only for assigned and in progress and completed status.

PR #1321, opened 2023-07-23
Include all tasks state such as NEEDS_REVIEW , IN_REVIEW etc. for Marking Users Idle/ Active

heyrandhir:
Closes 
- #1310

We currently update user status only for tasks when they are marked as completed. However, we also need to update user status for other statuses like NEEDS_REVIEW,  IN_REVIEW etc. to mark users as idle/active.

This feature is hidden behind a feature flag.

Testing Stats: 

Controllers 

image

### DIFF
--- a/test/integration/taskBasedStatusUpdate.test.js
+++ b/test/integration/taskBasedStatusUpdate.test.js
@@ -83,9 +83,38 @@ describe("Task Based Status Updates", function () {
         expect(res.body.userStatus.data.currentStatus).to.equal(userState.IDLE);
       });
 
-      it("Should not change the ACTIVE state to IDLE if no other task is assigned to the user.", async function () {
+      it("Should change the ACTIVE state to IDLE if no other task is assigned to the user.", async function () {
         const statusData = generateStatusDataForState(userId, userState.ACTIVE);
         await firestore.collection("usersStatus").doc("userStatus").set(statusData);
+        const res = await chai
+          .request(app)
+          .patch(`/tasks/self/taskid123?userStatusFlag=true`)
+          .set("cookie", `${cookieName}=${userJwt}`)
+          .send(reqBody);
+        expect(res.body.userStatus.status).to.equal("success");
+        expect(res.body.userStatus.message).to.equal("The status has been updated to IDLE");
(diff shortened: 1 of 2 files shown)


iamitprakash (changes requested):
test case stats missing

test/integration/taskBasedStatusUpdate.test.js
```diff
@@ -148,9 +177,38 @@ describe("Task Based Status Updates", function () {
         expect(res.body.userStatus.data.currentStatus).to.equal(userState.ACTIVE);
       });
 
-      it("Should change to ACTIVE state if the user is not ACTIVE.", async function () {
+      it("Should change to ACTIVE state if the user is not ACTIVE. ", async function () {
```
vikhyat187, line 180: `the user is not ACTIVE.` @heyrandhir if the user isn't active then do we change the status to ACTIVE?

heyrandhir: yes @vikhyat187 if the user has other task assigned on their name. then we will mark them active if their current status is IDLE.

heyrandhir:
> test case stats missing

I have just updated a line in the controller file. We don't have unit tests for controllers file. We check them in integration file. The stats for controller file coverage has been attached. Let me know if I need to add any other details. Shall I attach models file coverage as well ?

test/integration/taskBasedStatusUpdate.test.js
vikhyat187: Got it.

vikhyat187: NIT: If we can create a visual diagram, then it will be easier for a new person to understand the flow much easier. 
@heyrandhir

heyrandhir: > NIT: If we can create a visual diagram, then it will be easier for a new person to understand the flow much easier. @heyrandhir

will do it @vikhyat187 . Thanks for suggesting

vikhyat187 (commented):
There is a comment regarding the test cases, please check that from @iamitprakash

sahsisunny (approved):

vinit717 (approved):

iamitprakash (approved):

Issue #1305, opened 2023-07-19
Cache new response iff successful


Aryex82:
Description of the issue: 
1. For CacheResponse Middleware: The request will timeout if there's any error on the api's where the cache middleware is used.
2. For InvalidateCache Middleware: It will invalidate the cache without checking if the next actions are successful / unsuccessful.

PR #1322, opened 2023-07-24
Cache new response if successful 

Aryex82:
Fixes: #1305

Description: 
For caching: 
1. This PR adds a check to verify if the controller layer sets the status code in the range 2xx before caching the response.

For Invalidation: 
1. Currently we were invalidating the cache before passing the control to controller function. 
2. This PR makes necessary changes to make the invalidation after the controller function is executed. It also makes sure the call was successful by checking the status code. 

Test coverage: 
Screenshot 2023-07-24 at 9 33 09 PM

### DIFF
--- a/utils/cache.js
+++ b/utils/cache.js
@@ -160,6 +160,11 @@ const cacheResponse = (options = {}) => {
         const oldSend = res.send;
 
         res.send = (body) => {
+          if (res.statusCode < 200 || res.statusCode >= 300) {
+            res.send = oldSend;
+            return res.send(body);
+          }
+
           const cacheValue = {
             priority: priority,
             response: body,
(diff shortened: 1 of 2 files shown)


iamitprakash:
@RitikJaiswal75  can you please look here once

utils/cache.js
```diff
@@ -160,6 +160,11 @@ const cacheResponse = (options = {}) => {
         const oldSend = res.send;
 
         res.send = (body) => {
+          if (res.statusCode < 200 || res.statusCode >= 300) {
+            res.send = oldSend;
+            return res.send(body);
+          }
```
RitikJaiswal75, line 166: from where in the code do we get this res.statusCode (middleware/controller/somewhere else)?

Aryex82:
```javascript
 /**
         * As we do not have data in our cache we call the next middleware,
         * intercept the response being sent from middleware and store it in cache.
         *  */
        const oldSend = res.send;


        res.send = (body) => {
    
          if (res.statusCode < 200 || res.statusCode >= 300) {
            res.send = oldSend;
            return res.send(body);
          }

          const cacheValue = {
            priority: priority,
            response: body,
            expiry: new Date().getTime() + minutesToMilliseconds(expiry),
            size: Buffer.byteLength(body),
          };
          pool.set(key, cacheValue);
          if (modelKey) {
            cachedKeys.addCachedKey(modelKey, key);
          }
          res.send = oldSend;
          return res.send(body);
        };

        next();
``` 

In our `cacheResponse` middleware we are just assigning a new method to `res.send` 

This will be called in controller using res.send or res.json or res.boom.. etc..

RitikJaiswal75: > This will be called in controller using res.send or res.json or res.boom.. etc..

if this is being called in the controller and then all the code from the controller will be made then the db calls will be made. werent we suppossed to protect that?

RitikJaiswal75 (approved):
Thanks for taking the time and explaining it over the call.

yesyash (approved):

PR #1327, opened 2023-07-25
Add filter by query for task

vinit717:
Filtering the task based on search term query 
image

Utils
image

validator
image

Follow-up issue:

- [ ] https://github.com/Real-Dev-Squad/website-backend/issues/1365


test/integration/tasks.test.js
```diff
@@ -264,6 +264,29 @@ describe("Tasks", function () {
   });
 
   describe("GET /tasks/:id/details", function () {
+    it("Should get tasks filtered by search term", function (done) {
+      const searchTerm = "search";
+      chai
+        .request(app)
+        .get(`/tasks?q=${encodeURIComponent(searchTerm)}`)
```
heyrandhir, line 271: here everything you type in q will become search term. instead we should use `q=searchTerm:golang` so that later we can modify it to `q=searchTerm:golang+assignee:vinit` and so on

controllers/tasks.js
```diff
@@ -128,16 +128,24 @@ const fetchPaginatedTasks = async (query) => {
...
         message: "Tasks returned successfully!",
         ...paginatedTasks,
       });
     }
+    if (queryString) {
+      const filterTasks = await tasks.fetchTasks(searchTerm);
+      const tasksWithRdsAssigneeInfo = await fetchTasksWithRdsAssigneeInfo(filterTasks);
+      return res.json({
+        message: "Filter tasks returned successfully!",
+        tasks: tasksWithRdsAssigneeInfo.length > 0 ? tasksWithRdsAssigneeInfo : [],
```
heyrandhir, line 146: if the array is empty the message can be updated to `no tasks found` instead of `Filter tasks returned successfully`

test/unit/middlewares/tasks-validator.test.js
```diff
@@ -333,13 +333,14 @@ describe("getTasks validator", function () {
     expect(nextMiddlewareSpy.callCount).to.be.equal(1);
   });
 
-  it("should pass the request when correct parameters are passed: page, dev, status and size", async function () {
+  it("should pass the request when correct parameters are passed: page, dev, status, sixe and searchterm", async function () {
```
heyrandhir, line 336: can you add a test that it passes w/o the query term as well.

test/unit/models/tasks.test.js
```diff
@@ -91,6 +91,14 @@ describe("tasks", function () {
         expect(task).to.contain.all.keys(sameTask);
       });
     });
+    it("should fetch tasks filtered by search term", async function () {
+      const searchTerm = "task";
+      const tasksSnapshot = await tasksModel.get();
+      const result = await getBuiltTasks(tasksSnapshot, searchTerm);
```
heyrandhir, line 97: can we check the length here as well ! to be sure that its not an empty array and the below assertions are working.

heyrandhir (changes requested):

controllers/tasks.js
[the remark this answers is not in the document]
vinit717: yes for dev=true it only return paginated tasks

controllers/tasks.js
```diff
@@ -128,16 +128,24 @@ const fetchPaginatedTasks = async (query) => {
 
 const fetchTasks = async (req, res) => {
   try {
-    const { dev, status, page, size, prev, next } = req.query;
+    const { dev, status, page, size, prev, next, q: queryString } = req.query;
     const transformedQuery = transformQuery(dev, status, size, page);
-
+    const searchTerm = queryString || "";
```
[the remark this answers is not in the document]
vinit717: yes we can check in condition also

test/unit/models/tasks.test.js
vinit717: done

test/unit/middlewares/tasks-validator.test.js
vinit717: done

controllers/tasks.js
vinit717: done

test/integration/tasks.test.js
vinit717: done

middlewares/validators/tasks.js
```diff
@@ -164,6 +164,19 @@ const getTasksValidator = async (req, res, next) => {
         })
       ),
     size: joi.number().integer().positive().min(1).max(100).optional(),
+    q: joi
+      .string()
+      .optional()
+      .custom((value, helpers) => {
+        if (value && value.includes(":")) {
+          const [key] = value.split(":");
+          const allowedKeywords = ["searchterm", "assignee", "status"];
```
heyrandhir, line 173: is assignee and status handled ?

vinit717: not now because the scope of the pr is focused on the task search term only but can create another issue for it

test/integration/tasks.test.js
```diff
@@ -261,6 +261,45 @@ describe("Tasks", function () {
...
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Filter tasks returned successfully!");
+          expect(res.body.tasks).to.be.a("array");
+
+          const matchingTasks = res.body.tasks;
```
heyrandhir, line 278: can we check the matching task length

heyrandhir (changes requested):

middlewares/validators/tasks.js
vinit717: done

test/integration/tasks.test.js
vinit717: done

heyrandhir (approved):

sahsisunny (approved):

PR #1328, opened 2023-07-26
Fixoldtaskstatus

chandrareddy7:
### Closes issue #1298 

Added a new route `/updateOldTaskStatus/all` to run a function which updates old task status as following.
'COMPLETED' -> 'DONE'
'unassigned' -> 'UNASSIGNED'.


models/tasks.js
```diff
@@ -101,7 +101,7 @@ const getBuiltTasks = async (tasksSnapshot) => {
   const promises = tasks.map(async (task) => fromFirestoreData(task));
   const updatedTasks = await Promise.all(promises);
   const taskPromises = updatedTasks.map(async (task) => {
-    task.status = TASK_STATUS[task.status.toUpperCase()] || task.status;
+    // task.status = TASK_STATUS[task.status.toUpperCase()] || task.status;
```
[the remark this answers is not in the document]
chandrareddy7: Removed it.

controllers/tasks.js
```diff
@@ -401,6 +401,32 @@ const assignTask = async (req, res) => {
...
+    }
+    return res.json({ message: `Updated ${updatedTasks.length} tasks`, updatedTasks });
+  } catch (error) {
+    return res.json({
+      message: "error while updating old task's status",
+      error: error,
+    });
+  }
+};
+
```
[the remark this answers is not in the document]
chandrareddy7: This is a one time route which is used to update old task status.
Once this route is hit, we won't be using this anymore. ( This route can be removed after running the script).
So, should I write tests?

controllers/tasks.js
```diff
@@ -401,6 +401,27 @@ const assignTask = async (req, res) => {
   }
 };
 
+const updateOldTaskStatus = async (req, res) => {
+  try {
+    const allTasks = await tasks.fetchTasks();
+    const allOldTasks = allTasks.filter(
+      (task) => task.status.toUpperCase() === "COMPLETED" || task.status === "unassigned"
+    );
```
heyrandhir, line 409: why only capital case for COMPLETED . There can be tasks with status COMPLETED in small case.

controllers/tasks.js
```diff
@@ -401,6 +401,27 @@ const assignTask = async (req, res) => {
...
+    const updatedTasks = [];
+    for (const task of allOldTasks) {
+      if (task.status.toUpperCase() === "COMPLETED") {
```
heyrandhir, line 412: since COMPLETED is used at multiple places. can we make it into a constant or if it already exist can it be imported from there

controllers/tasks.js
```diff
@@ -401,6 +401,27 @@ const assignTask = async (req, res) => {
...
+        updatedTasks.push(await tasks.updateTask({ status: "DONE" }, task.id));
+      } else if (task.status === "unassigned") {
+        updatedTasks.push(await tasks.updateTask({ status: "UNASSIGNED" }, task.id));
+      }
+    }
+    return res.json({ message: `Updated Old tasks`, tasks: updatedTasks });
```
heyrandhir, line 418: can we provide a more meaningful response message please

controllers/tasks.js
```diff
@@ -401,6 +401,27 @@ const assignTask = async (req, res) => {
...
+  } catch (error) {
+    logger.error(`Error while Updating tasks ${error}`);
+    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
+  }
+};
+
```
heyrandhir, line 424: This is business logic. Should ideally happen in model file instead of controller

controllers/tasks.js
```diff
@@ -401,6 +401,27 @@ const assignTask = async (req, res) => {
...
+    const allTasks = await tasks.fetchTasks();
+    const allOldTasks = allTasks.filter(
+      (task) => task.status.toUpperCase() === "COMPLETED" || task.status === "unassigned"
+    );
+    const updatedTasks = [];
+    for (const task of allOldTasks) {
+      if (task.status.toUpperCase() === "COMPLETED") {
+        updatedTasks.push(await tasks.updateTask({ status: "DONE" }, task.id));
+      } else if (task.status === "unassigned") {
+        updatedTasks.push(await tasks.updateTask({ status: "UNASSIGNED" }, task.id));
```
heyrandhir, line 415: can we use all the task state constants from constants please instead of hardcoding

models/tasks.js
```diff
@@ -5,7 +5,7 @@ const dependencyModel = firestore.collection("taskDependencies");
 const userUtils = require("../utils/users");
 const { fromFirestoreData, toFirestoreData, buildTasks } = require("../utils/tasks");
 const { TASK_TYPE, TASK_STATUS, TASK_STATUS_OLD, TASK_SIZE } = require("../constants/tasks");
-const { IN_PROGRESS, BLOCKED, SMOKE_TESTING, COMPLETED } = TASK_STATUS;
+const { IN_PROGRESS, BLOCKED, SMOKE_TESTING, DONE } = TASK_STATUS;
```
heyrandhir, line 8: will this change affect the frontend sites which are currently using the old statuses ? will it not cause breaking changes in the frontend repos !

test/integration/tasks.test.js
```diff
@@ -820,4 +820,38 @@ describe("Tasks", function () {
...
+        .patch("/tasks/updateOldTaskStatus/all")
+        .set("cookie", `${cookieName}=${superUserJwt}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.equal("Updated Old tasks");
+          expect(res.body.tasks).to.be.a("array");
```
heyrandhir, line 837: asserting type to be an array doesn't assures much we should also check the content in the array.

heyrandhir (changes requested):

controllers/tasks.js
chandrareddy7: Just didn't want to write `task.status === 'completed' || task.status === 'COMPLETED'` so made it into `task.status.toUpperCase() === 'COMPLETED'`.

Before navigating to the route.
image

After navigating to the route.
image

heyrandhir: can we do the same for `unassigned` ?

routes/tasks.js
```diff
@@ -39,5 +39,6 @@ router.patch(
   assignTask
 );
 router.patch("/assign/self", authenticate, invalidateCache({ invalidationKeys: [ALL_TASKS] }), tasks.assignTask);
+router.patch("/updateOldTaskStatus/all", authenticate, authorizeRoles([SUPERUSER]), tasks.updateOldTaskStatus);
```
heyrandhir, line 42: the path name should be a resource. Please modify it accordingly

heyrandhir (changes requested):

models/tasks.js
chandrareddy7: The change in tasks constants file could potentially break the status page. As the status page renders tasks with status as 'COMPLETED' and I have removed it from constants (as we are updating all tasks with 'COMPLETED' status to 'DONE').

heyrandhir: In that case it should be handled in those sites as well so that they don't start breaking.

controllers/tasks.js
chandrareddy7: image

Changed it.

routes/tasks.js
```diff
@@ -39,5 +39,6 @@ router.patch(
   assignTask
 );
 router.patch("/assign/self", authenticate, invalidateCache({ invalidationKeys: [ALL_TASKS] }), tasks.assignTask);
+router.patch("/oldtasks/all", authenticate, authorizeRoles([SUPERUSER]), tasks.updateOldTaskStatusController);
```
heyrandhir, line 42: can we keep just /migrate ?

heyrandhir (changes requested):
Please add tests for model function as well

chandrareddy7:
### When user with no SUPERUSER role hits the route.
image

### When user with SUPERUSER role hits the route and no tasks are found to update.
image

### When user with SUPERUSER role hits the route.
image

controllers/tasks.js
```diff
@@ -401,6 +401,29 @@ const assignTask = async (req, res) => {
...
+      return res.json({ message: message });
+    }
+    if (updatedTasksCount > 0) {
+      message = `Updated ${updatedTasksCount} task(s) with old status as ${Array.from(oldStatus).join(", ")}`;
+    }
+    return res.json({ message: message, tasks: updatedTasks });
+  } catch (error) {
+    logger.error(`Error while Updating tasks ${error}`);
+    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
+  }
```
heyrandhir, line 424: can you write test that covers this condition as well

heyrandhir (changes requested):

heyrandhir (commented):
I think first the changes in Frontend needs to go live with feature flag. then this should get merge. so that once the script is triggered. frontend will be still usable with feature flag. if the script is directly merged then frontend will break. as I dont see any open PR in frontend.

Issue #1214, opened 2023-06-30
Add logs of archive details in the archiving process


gauravsinhaweb:
Closes - #1465
Currently, there is a need to enhance the process of assigning users to the `archived` status in the website backend. The current implementation lacks the necessary details, which makes it difficult to track and understand why a developer has been placed in the archived status.

### Solution

To improve this process, we need to introduce the following details while assigning users to the `archived` status:

1. **Reason**: Include a field to capture a brief explanation or reason for putting the developer in the archived status. This will help provide context and assist in future reference or auditing purposes.

2. **Who (ID)**: Add a field to specify the name of the assignee responsible for assigning the developer to the archived status. This will allow for better accountability and facilitate communication between team members.

3. **Superuser ID**: Introduce a field to capture the name of the superuser who is performing the action. This will help identify the authority behind the decision to archive a developer and provide transparency within the system.

heyrandhir:
> ## Ticket Description
> ### Issue
> Currently, there is a need to enhance the process of assigning users to the `archived` status in the website backend. The current implementation lacks the necessary details, which makes it difficult to track and understand why a developer has been placed in the archived status.
> 
> ### Proposed Solution
> To improve this process, we need to introduce the following details while assigning users to the `archived` status:
> 
> 1. **Message**: Include a field to capture a brief explanation or reason for putting the developer in the archived status. This will help provide context and assist in future reference or auditing purposes.
> 2. **Who**: Add a field to specify the name of the assignee responsible for assigning the developer to the archived status. This will allow for better accountability and facilitate communication between team members.
> 3. **Superuser Name**: Introduce a field to capture the name of the superuser who is performing the action. This will help identify the authority behind the decision to archive a developer and provide transparency within the system.
> 
> ETA: will add if the ticket got approved.

We should store the superuser id and the user archived id not the name.

PR #1329, opened 2023-07-26
Feat: add logs of archive details in archiving process

gauravsinhaweb:
This PR is linked to issue/ ticket #1214 

**API contract link: https://github.com/Real-Dev-Squad/website-api-contracts/pull/145**
**API data modal link: https://github.com/Real-Dev-Squad/website-data-models/pull/3**

## To add the archive details of the users in the archiving process.

**Logs structure**

```js
const archivedUserDetailsModal = [
  {
    type: "archived-details",
    meta: { },
    body: {
      reason: "test reason",
      archived_user: { user_id: "R5kljdsleH4Gr2t7tvr0Z", username: "testUser1" },
      archived_by: {
        user_id: "ReMyuklislajwooncVL",
        roles: {
          in_discord: true,
          super_user: false,
          member: true,
          archived: false,
        },
      },
    },
    timestamp: {
      _seconds: 1657193216,
      _nanoseconds: 912000000,
    },
  },
]
  ```
 **API endpoints**

- GET `/logs/archivedUsers` - This endpoint will fetch all the logs of archive details
- GET `/logs/archivedUsers?username=test` - This endpoint will fetch the logs of specific archive details based on the query `username`

**Test Cases**

**GET** `/logs/archivedUsers` 

- [x] Should return an empty array if no logs found
- [x] Should fetch all archived logs

**GET** `/logs/archivedUsers?username=test` 

- [x] Should fetch all archived logs for a given username
- [x] Should throw response status 404, if username is incorrect in the query

**Test Case Result**
Screenshot 2023-08-10 183359
Screenshot 2023-08-10 183218

**Test Coverage**

Screenshot 2023-08-10 214941

These updates enhance the archiving process, allowing us to store and retrieve valuable archive details for users efficiently.

### DIFF
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -3,6 +3,7 @@ const members = require("../models/members");
 const tasks = require("../models/tasks");
 const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
 const dataAccess = require("../services/dataAccessLayer");
+const { addLog } = require("../models/logs");
 /**
  * Fetches the data about our members
  *
@@ -82,18 +83,23 @@ const moveToMembers = async (req, res) => {
 const archiveMembers = async (req, res) => {
   try {
     const { username } = req.params;
-    const user = await fetchUser({ username });
-    const userId = user.user.id;
+    const user = await dataAccess.retrieveUsers({ username });
     const superUserId = req.userData.id;
     const body = req.body;
+    const meta = {
+      userId: user?.user?.id,
+      superUserId: superUserId,
+      username: username,
+    };
     if (!body?.reason || /^\s*$/.test(body?.reason)) {
       return res.boom.badRequest("Reason is required");
     }
     if (user?.userExists) {
-      const successObject = await members.addArchiveRoleToMembers(userId, username, superUserId, body?.reason);
+      const successObject = await members.addArchiveRoleToMembers(user.user.id);
       if (successObject.isArchived) {
         return res.boom.badRequest("User is already archived");
       }
+      addLog("archiveDetails", meta, { body: body?.reason });
       return res.status(204).send();
     }
     return res.boom.notFound("User doesn't exist");
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -1,7 +1,7 @@
 const ROLES = require("../constants/roles");
 const members = require("../models/members");
 const tasks = require("../models/tasks");
-const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
+const { SOMETHING_WENT_WRONG, INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
 const dataAccess = require("../services/dataAccessLayer");
 const { addLog } = require("../models/logs");
 /**
@@ -99,13 +99,13 @@ const archiveMembers = async (req, res) => {
       if (successObject.isArchived) {
         return res.boom.badRequest("User is already archived");
       }
-      addLog("archiveDetails", meta, { body: body?.reason });
+      addLog("archive-details", meta, { body: body?.reason });
       return res.status(204).send();
     }
     return res.boom.notFound("User doesn't exist");
   } catch (err) {
     logger.error(`Error while retriving contributions ${err}`);
-    return res.boom.badImplementation(SOMETHING_WENT_WRONG);
+    return res.status(500).json({ message: INTERNAL_SERVER_ERROR });
   }
 };
 
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -1,9 +1,8 @@
 const ROLES = require("../constants/roles");
 const members = require("../models/members");
 const tasks = require("../models/tasks");
-const { fetchUser } = require("../models/users");
 const { SOMETHING_WENT_WRONG } = require("../constants/errorMessages");
-
+const dataAccess = require("../services/dataAccessLayer");
 /**
  * Fetches the data about our members
  *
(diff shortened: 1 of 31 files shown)

### CODE REVIEW

File: controllers/members.js [hunk 1]
```diff
@@ -83,11 +84,22 @@ const archiveMembers = async (req, res) => {
   try {
     const { username } = req.params;
     const user = await dataAccess.retrieveUsers({ username });
+    const superUserId = req.userData.id;
```
Pratiyushkumar, line 87: are we sure here that `req.userData.id` will give here super user id??

controllers/members.js
```diff
@@ -83,11 +84,22 @@ const archiveMembers = async (req, res) => {
...
+    const body = req.body;
+    const meta = {
+      userId: user?.user?.id,
+      superUserId: superUserId,
+      username: username,
+    };
+    if (!body?.reason || /^\s*$/.test(body?.reason)) {
+      return res.boom.badRequest("Reason is required");
```
Pratiyushkumar, line 95: `|| /^\s*$/.test(body?.reason)`--> why??

controllers/members.js
```diff
@@ -83,11 +84,22 @@ const archiveMembers = async (req, res) => {
   try {
     const { username } = req.params;
     const user = await dataAccess.retrieveUsers({ username });
+    const superUserId = req.userData.id;
+    const body = req.body;
+    const meta = {
+      userId: user?.user?.id,
+      superUserId: superUserId,
+      username: username,
```
Pratiyushkumar, line 92: what is meta here?? when I read this variable name first thought I get is of meta tag..is it something related to that...??

test/unit/models/logs.test.js
Pratiyushkumar, line 1: what if unauthorized user tries to access the API, can we also test that..??
what if we have an internal server error, in that case what is the response for API, can we also test that??

Pratiyushkumar (changes requested):

controllers/members.js
gauravsinhaweb: yes, if the user has access to put the users to achieve, they are authorized user

controllers/members.js
gauravsinhaweb: Just in case the reason a super user will mention should not be `empty space`, it should have proper text

controllers/members.js
gauravsinhaweb: Since in our `logs.js`, we have a function to `addLogs`, which takes three param - `type`, `meta` and `body`, 
so passing the `meta` here into that function as a param and keeping the object name the same for reference i.e. `meta`

test/unit/models/logs.test.js
gauravsinhaweb: Those test cases have already been implemented

controllers/members.js
Pratiyushkumar: `!body?.reason ` itself not handle empty space condition??

gauravsinhaweb: no, because empty space has length 1, so it will not validate properly
Also, the current implementation can handle multiple empty spaces to make validation properly

test/unit/models/logs.test.js
gauravsinhaweb: you'll find it in develop brach, https://github.com/Real-Dev-Squad/website-backend/blob/develop/test/integration/members.test.js

gauravsinhaweb: Done

- line no. 322 `Should return 401 if user is not a super use` in the above link
- Updated: test handling internal server error

Pratiyushkumar:
Please add Test Coverage as well

gauravsinhaweb:
> Please add Test Coverage as well

Added https://github.com/Real-Dev-Squad/website-backend/pull/1329#issue-1822371683

controllers/members.js
```diff
@@ -83,17 +84,28 @@ const archiveMembers = async (req, res) => {
...
+    };
+    if (!body?.reason || /^\s*$/.test(body?.reason)) {
```
sahsisunny, line 94:
```suggestion
    const isReasonNullOrUndefined = !body?.reason;
    const isReasonEmptyOrWhitespace = /^\s*$/.test(body?.reason);
    if (isReasonNullOrUndefined || isReasonEmptyOrWhitespace) {
```

test/fixtures/logs/archievedUsers.js
sahsisunny, line 1: Data should look like the original data please fix it

test/unit/models/logs.test.js
```diff
@@ -37,4 +48,80 @@ describe("Logs", function () {
...
+
+    it("Should return an object with status 500 and an error message", async function () {
+      addLogsStub = Sinon.stub(logsQuery, "fetchLogs");
+      addLogsStub.throws(new Error(INTERNAL_SERVER_ERROR));
+
+      addUser(userToBeMadeMember).then(() => {
+        const res = chai.request(app).get("/logs/archive-details").set("cookie", `${cookieName}=${jwt}`).send();
+
+        // expect(res).to.have.status(500);
+        // expect(res.body).to.have.property("message").that.is.a("string");
```
sahsisunny, line 72: ?

gauravsinhaweb: Updated

test/fixtures/logs/archievedUsers.js
gauravsinhaweb: Updated

iamitprakash (approved):

sahsisunny:
If changes are made, would it be possible to update the data model as well?

sahsisunny (approved):

gauravsinhaweb:
> If changes are made, would it be possible to update the data model as well?

the current structure follows the same data structure as https://github.com/Real-Dev-Squad/website-data-models/pull/3,
the changes are instead of passing details in `meta`, we are passing details in the `body`, and the `meta` will remain empty.
So there is no need to update the data modal. 
example structure is in the description.

Pratiyushkumar (approved):

Issue #1233, opened 2023-07-05
Archived 'false' for  in_discord 'true'


j24m:
### Issue Description :
Initially when a user joins discord server of RDS, the inDiscord property of user should be true and the archived property should be false, but that is not the case, some users are still archived.

### Steps to Resolve :
Check if the inDiscord property is true and if so, then add a property **`archived : false`**  .

### Acceptance Criteria :
100% test coverage 

### Related Files : 
- external-accounts.js
- external-accounts.test.js

PR #1335, opened 2023-07-27
Script to Update 'archived' Role When 'in_discord' is True

RitikJaiswal75:
# Script to Update 'archived' Role When 'in_discord' is True

closes: #1233 

### Expected Behavior
When the 'in_discord' flag is set to true for a user, their 'archived' role should be updated to true as well.

### Current Behavior
Currently, there is no automated process to update the 'archived' role when the 'in_discord' flag is set to true for a user. The 'archived' role remains unaffected even if 'in_discord' is true.

## Proposed Solution
Create a script that runs periodically and checks the 'in_discord' flag for each user. If the 'in_discord' flag is true for a user, the script should update their 'archived' role to true.

## Script Implementation
The script should be written in a programming language with access to the necessary database or user management system. The script should perform the following steps:

1. Connect to the user database or user management system.
2. Retrieve a list of all users.
3. For each user in the list, check if the 'in_discord' flag is true.
4. If the 'in_discord' flag is true for a user, update their 'archived' role to true.
5. Log the changes made to the 'archived' role for auditing purposes.
6. Handle any errors or exceptions gracefully to ensure the script's reliability.

### Controllers
image

### Models
image

### DIFF
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -1,7 +1,7 @@
 const externalAccountsModel = require("../models/external-accounts");
 const { SOMETHING_WENT_WRONG, INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
 const { getDiscordMembers } = require("../services/discordService");
-const { addOrUpdate } = require("../models/users");
+const { addOrUpdate, getUsersByRole } = require("../models/users");
 const { retrieveDiscordUsers } = require("../services/dataAccessLayer");
 const logger = require("../utils/logger");
 
@@ -56,6 +56,7 @@ const syncExternalAccountData = async (req, res) => {
     const rdsUserDataMap = {};
     const updateUserDataPromises = [];
     const userUpdatedWithInDiscordFalse = [];
+    const updateArchivedPromises = [];
 
     rdsUserData.forEach((rdsUser) => {
       rdsUserDataMap[rdsUser.discordId] = {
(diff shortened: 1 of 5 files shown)


controllers/external-accounts.js
```diff
@@ -90,10 +91,26 @@ const syncExternalAccountData = async (req, res) => {
...
+          roles: {
+            ...user.roles,
+            archived: false,
+          },
+        };
+        updateArchivedPromises.push(addOrUpdate(userData, user.id));
+      }
+    });
+
+    await Promise.all(updateArchivedPromises);
```
pallabez, line 107: Do we need two `Prosime.all`? One here and other in line 92

test/unit/models/users.test.js
```diff
@@ -257,4 +257,26 @@ describe("users", function () {
...
+      members.forEach((member) => {
+        expect(member.roles.member).to.be.equal(true);
+      });
+    });
+    it("throws an error", async function () {
+      await users.getUsersByRole(32389434).catch((err) => {
+        expect(err).to.be.instanceOf(Error);
+      });
+    });
+  });
```
pallabez, line 281: Thanks for the test

models/users.js
```diff
@@ -622,6 +622,25 @@ const removeGitHubToken = async (users) => {
     throw err;
   }
 };
+
+const getUsersByRole = async (role) => {
+  try {
+    const usersRef = await userModel.where(`roles.${role}`, "==", true).get();
```
pallabez, line 628: What do we get here is role is invalid? Error or empty array?

controllers/external-accounts.js
RitikJaiswal75: Yes as its an array of promises and if we wait inside loop for each user then the api wait time will increase drastically, 
promise.all resolves all promises without waiting for previpus one to finish, i.e. dosent matter which promise resolves first as soon as all the promises are resolved the await on promise.all will move ahead in the thread.

models/users.js
RitikJaiswal75: throw an error which is handled with 5xx response.

RitikJaiswal75: in the test you can see that i have entered a random role and i check for the error to occur.

gauravsinhaweb (approved):

controllers/external-accounts.js
```diff
@@ -1,7 +1,7 @@
 const externalAccountsModel = require("../models/external-accounts");
 const { SOMETHING_WENT_WRONG, INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
 const { getDiscordMembers } = require("../services/discordService");
-const { addOrUpdate } = require("../models/users");
```
gauravsinhaweb, line 4: `addOrUpdate` name can be more descriptive

gauravsinhaweb (approved):
LGTM

iamitprakash (approved):

vvaibhavdesai (approved):

Issue #1344, opened 2023-07-29
Add timestamp to most used collections


sahsisunny:
# Add timestamp to all collections

1. user:
    - created_at : first time document created -> 
    - updated_at : last modified time

2. user-status - Onboarding:
    - from : when status add to onboarding

3. Tasks : 
    - created_at : first time document created
    - updated_at : last modified time

4. applicants:
    - created_at : first time document created

5. groups: change date object to timestamp
    - created_at : first time document created
    - updated_at : last modified time

1610ayush:
Hey @sahsisunny Can i start working on this issue?

PR #1346, opened 2023-07-29
FEATURE: add timestamp to user model collection

sahsisunny:
## Parent TIcket - #1344

**Description:**
In this PR we have added two new fields in the user collection. These fields are created_at and updated_at. These fields will be used to keep track of when the document is created and when it is last modified. We have also updated the existing records to have these fields.

** Databases Changes:**

- [x] Yes

**Note:** This PR is only for new users not for migration-related changes.

**Test Coverage**
Controllers
image
Models
image
Services
image


sahsisunny:
Temporary closed as per discussion with @RitikJaiswal75.

controllers/auth.js
```diff
@@ -53,6 +53,9 @@ const githubAuthCallback = (req, res, next) => {
       userData = {
         github_id: user.username,
         github_display_name: user.displayName,
+        // github_account_created_at: user.created_at,
```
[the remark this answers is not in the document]
sahsisunny: Someone is working on it so we can keep this comment code as of now it also  helps to understand the created_at field is for RDS user document not  for GitHub account created date.

iamitprakash (approved):

models/users.js
```diff
@@ -53,6 +54,7 @@ const addOrUpdate = async (userData, userId = null) => {
         isNewUser: false,
         userId: user.docs[0].id,
         incompleteUserDetails: user.docs[0].data().incompleteUserDetails,
```
iamitprakash, line 56: what if docs is null

iamitprakash (changes requested):

models/users.js
sahsisunny: This change is not part of my PR.

sahsisunny: > what if docs is null

https://github.com/Real-Dev-Squad/website-backend/blob/2dc69673d6bb4899a250425a75827381954308de/models/users.js#L50C1-L51C1

iamitprakash: userId: user?.docs[0]?.id,

iamitprakash (approved):

controllers/auth.js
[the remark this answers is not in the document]
heyrandhir: for now fine but ideally we should remove as they can always look in git history to find it.

test/fixtures/user/user.js
```diff
@@ -329,5 +329,19 @@ module.exports = () => {
...
+      github_id: githubUserInfo[0].username,
+      github_display_name: githubUserInfo[0].displayName,
+      roles: {
+        member: true,
+        in_discord: true,
+      },
+      incompleteUserDetails: false,
+      updated_at: Date.now(),
+      created_at: Date.now(),
+    },
```
heyrandhir, line 345: Ankur is already the user in the array. can we rename the username to avoid confusion ?

test/integration/auth.test.js
```diff
@@ -90,7 +91,7 @@ describe("auth", function () {
     expect(res.headers.location).to.equal(rdsUrl);
   });
 
-  it("should redirect the realdevsquad.com if non RDS URL provided, any url that is other than *.realdevsqud.com is invalid", async function () {
+  it("should redirect the realdevsquad.com if non RDS URL provided, any url that is other than *.realdevsquad.com is invalid", async function () {
```
heyrandhir, line 94: thanks for fixing the typos ❤️

heyrandhir (changes requested):

test/fixtures/user/user.js
sahsisunny: Done

heyrandhir (approved):

PR #1350, opened 2023-07-31
feat: made `users/githubInfo/userId` endpoint to get user github profile

ravikumar1002:
### What is the change?
Created route to get user github data through RDS id 

The process to write endpoint:
 - Created a route for that.
 - In a controller, I have written a function that will be called when someone accesses that route.
-  In the user modal, I have a write function that will basically collect user data through rds id and after that, it will call GitHub API.
-  In the service file I have written the GitHub app call part.

### TDD?
 No, I will do it 

### Database changes:
 No

Demo: 

https://github.com/Real-Dev-Squad/website-backend/assets/86847625/d75e68ca-ecd1-4a68-a99b-38a83a54fca9

### DIFF
--- a/models/users.js
+++ b/models/users.js
@@ -19,6 +19,7 @@ const userStatusModel = firestore.collection("usersStatus");
 const photoVerificationModel = firestore.collection("photo-verification");
 const { ITEM_TAG, USER_STATE } = ALLOWED_FILTER_PARAMS;
 const admin = require("firebase-admin");
+const { fetchUserGithubInfo } = require("../services/githubService");
 
 /**
  * Adds or updates the user data
@@ -266,6 +267,25 @@ const fetchUser = async ({ userId = null, username = null, githubUsername = null
   } catch (err) {
     logger.error("Error retrieving user data", err);
     throw err;
+  }
+};
+
+const fetchGithubInfoByUserId = async (userId) => {
+  if (!userId) {
+    throw new Error("User Id is missing");
+  }
+  try {
+    const userData = await fetchUser({ userId: userId });
+    const { userExists, user } = userData;
+    if (!userExists || !user.github_id) {
+      throw new Error("User not found");
+    }
+    const githubData = await fetchUserGithubInfo(user.github_id);
+    const { data } = githubData;
+    return data;
+  } catch (err) {
+    logger.error(`Error while fetching user github information: ${err}`);
+    throw new Error(err);
   }
 };
 
@@ -645,6 +665,7 @@ module.exports = {
   addOrUpdate,
   fetchPaginatedUsers,
   fetchUser,
+  fetchGithubInfoByUserId,
   setIncompleteUserDetails,
   initializeUser,
   updateUserPicture,
--- a/routes/users.js
+++ b/routes/users.js
@@ -21,6 +21,7 @@ router.get("/chaincode", authenticate, users.generateChaincode);
 router.get("/search", userValidator.validateUserQueryParams, users.filterUsers);
 router.patch("/remove-tokens", authenticate, authorizeRoles([SUPERUSER]), users.removeTokens);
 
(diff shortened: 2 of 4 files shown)


heyrandhir (changes requested):
Please add tests. PR's w/o tests needs special approval from Ankush to get merged.

routes/users.js
```diff
@@ -21,6 +21,7 @@ router.get("/chaincode", authenticate, users.generateChaincode);
 router.get("/search", userValidator.validateUserQueryParams, users.filterUsers);
 router.patch("/remove-tokens", authenticate, authorizeRoles([SUPERUSER]), users.removeTokens);
 
+router.get("/githubInfo/:userId", users.userGithubInfo);
```
RitikJaiswal75, line 24: please use a better api name githubInfo, seems bit odd. also we do not use camel case in URLs.

models/users.js
```diff
@@ -269,6 +270,25 @@ const fetchUser = async ({ userId = null, username = null, githubUsername = null
   }
 };
 
+const fetchGithubInfoByUserId = async (userId) => {
+  if (!userId) {
+    throw new Error("User Id is missing");
```
RitikJaiswal75, line 275: this error isn't caught anywhere

models/users.js
```diff
@@ -269,6 +270,25 @@ const fetchUser = async ({ userId = null, username = null, githubUsername = null
...
+  }
+  try {
+    const userData = await fetchUser({ userId: userId });
+    const { userExists, user } = userData;
+    if (!userExists || !user.github_id) {
+      throw new Error("User not found");
+    }
```
RitikJaiswal75, line 282: Can we make this funtction simpler models only need to query database, here first we are fetching user id then passing the user id to wait again for github info. make a function that fetches github info based on github id. the other part of extracting user id can be done some place else.

services/githubService.js
```diff
@@ -106,6 +106,25 @@ function getFetch(url) {
 function getFetchWithAuthToken(url, params = null, headers = null) {
   return utils.fetch(url, "get", params, null, headers);
 }
+
+/**
+ * Fetches the user github informations by userId using GitHub API
+ * @param githubUsername {string} - githubUsername String
+ */
+
+const fetchUserGithubInfo = async (githubUsername) => {
```
RitikJaiswal75, line 115: can we rename it to fetchGithubInfo?

and since the function already says github info lets rename the parameter as username.

services/githubService.js
```diff
@@ -106,6 +106,25 @@ function getFetch(url) {
...
+  try {
+    if (!githubUsername) {
+      throw new Error("github username missing");
+    }
```
RitikJaiswal75, line 119: if block will throw one error that will be caught in catch block and catch block will send another message for this condition logger wont be abe to log error thus we will get errors. i.e. here the error.response.data.message will be undefined.

use separate try catch

ravikumar1002:
> Please add tests. PR's w/o tests needs special approval from Ankush to get merged.

Ok


> Please add tests. PR's w/o tests needs special approval from Ankush to get merged.

yeah sure, from next time I will make a PR with test

routes/users.js
ravikumar1002: ok, I will keep in mind from next time

models/users.js
ravikumar1002: ok

models/users.js
ravikumar1002: Yes I had a chat with Ankush he told me to store GitHub created on value in user modal so I don't need these things
that's why I changed this PR into a draft.

services/githubService.js
ravikumar1002: Ok next time I will try to  follow these steps

prakashchoudhary07:
Closing this, as it's incomplete

Issue #1314, opened 2023-07-21
Create levels of access in the recently added data access layer


mailmesriza98:
Parent issue: 
- https://github.com/Real-Dev-Squad/website-backend/issues/1196

In the current data access layer we are removing all the sensitive informational details like **phone, email, chaincode, tokens.** while fetching users data.
However, going forward we would be providing an option to provide different levels of access in the dataAccess layer so that 
the module utilizing the data access layer can pass on its requirement (level 1 or 2 etc) for the granularity of information needed in the sensitive user data.

The levels will be categorized as below:
- level 0: all user-sensitive data is removed(phone,email,chaincode,tokens)
- level 1:  email is disclosed along with non-sensitive data
- level 2 : phone and email is disclosed along with non-sensitive data
- level 3 : phone, email and tokens are disclosed alongside non-sensitive data

The module using the dataAccess layer can pass on the levels required as arguments depending on the requirements. 
Also a privileged role like superusers can be able to pass on a certain level at a time if required to gather sensitive information.

Please let me know thoughts and comments on this issue.

heyrandhir:
This approach looks good to me, However I have some doubts 

Will we have some restriction on the layered authorized access. Will layer 0 data be accessible to all user or only super user.  How will we be managing access levels authorization. Implementing robust access control mechanisms rules for user privileges will be essential to avoid data breaches and unauthorized access. 

Additionally, providing detailed documentation on access level guidelines and enforcing best practices will help users and developers understand and adhere to the defined data access policies effectively.

vikhyat187:
Hi @mailmesriza98 if you have any rough approach in mind, can you just create a diagram for the approach that you will follow to implement this?

vikhyat187:
I think granular level based permissions should be implemented, as this can help to protect the system, say a malicious actor tries to cause harm to the resources or misuse the data, we can put the restriction on user level, in case of the role based access we will have to create a new role to restrict him / her.

@heyrandhir @prakashchoudhary07 what are your thoughts

prakashchoudhary07:
I think, we should go ahead with role based approach. We are already having role based system in place.

heyrandhir:
[...]

PR #1363, opened 2023-08-03
Create levels of access via data access layer

mailmesriza98:
closes https://github.com/Real-Dev-Squad/website-backend/issues/1314

coverage


services/dataAccessLayer.js
```diff
@@ -70,11 +91,37 @@ const removeSensitiveInfo = function (obj) {
...
+  if (level === AccessLevel.CONFIDENTIAL) {
+    user.chaincode = data.chaincode;
+  }
+  return user;
+};
+
+const levelSpecificAccess = (user, level = AccessLevel.PUBLIC, role = null) => {
+  const unFilteredData = JSON.parse(JSON.stringify(user));
+  removeSensitiveInfo(user);
+  if (level === AccessLevel.PUBLIC) {
```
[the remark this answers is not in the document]
iamitprakash: that's the issues no, as soon level will get matched with public the return will get executed,  am i correct...?

iamitprakash (approved):

services/dataAccessLayer.js
```diff
@@ -1,41 +1,62 @@
...
+    const result = await userQuery.fetchUser({ userId: userdata.id });
+    return levelSpecificAccess(result.user, level, role);
   }
 };
 
-const retrieveDiscordUsers = async () => {
+const retrieveDiscordUsers = async (level = AccessLevel.PUBLIC, role = null) => {
   const users = await userQuery.getDiscordUsers();
+  const userdata = [];
   users.forEach((element) => {
```
heyrandhir, line 55: can we rename element to something more meaningful such as `userData`

services/dataAccessLayer.js
```diff
@@ -1,41 +1,62 @@
...
-    return userdata;
+    const result = await userQuery.fetchUser({ userId: userdata.id });
+    return levelSpecificAccess(result.user, level, role);
   }
 };
 
-const retrieveDiscordUsers = async () => {
+const retrieveDiscordUsers = async (level = AccessLevel.PUBLIC, role = null) => {
   const users = await userQuery.getDiscordUsers();
+  const userdata = [];
```
heyrandhir, line 54: this is storing all the user data can we call it in plural form `usersData` instead.

services/dataAccessLayer.js
```diff
@@ -70,11 +91,37 @@ const removeSensitiveInfo = function (obj) {
...
+
+const levelSpecificAccess = (user, level = AccessLevel.PUBLIC, role = null) => {
+  const unFilteredData = JSON.parse(JSON.stringify(user));
+  removeSensitiveInfo(user);
+  if (level === AccessLevel.PUBLIC) {
+    return user;
+  }
+  if (!role.super_user) {
+    return "unauthorized";
+  }
```
heyrandhir, line 113:
```suggestion
  if (role === null || !role.super_user) {
    return "unauthorized";
  }
```

test/fixtures/user/user.js
```diff
@@ -267,12 +267,6 @@ module.exports = () => {
       linkedin_id: "testuser1",
       github_id: "testuser1",
       github_display_name: "Test User",
-      phone: "1234567890",
-      email: "tu@gmail.com",
-      chaincode: "1234",
-      tokens: {
-        githubAccessToken: "githubAccessToken",
-      },
```
heyrandhir, line 275: why are we removing these fields. I know these are confidential but they still needed to be added in the DB. Right ? then while accessing we will remove this data.

test/unit/services/dataAccessLayer.test.js
```diff
@@ -62,7 +65,7 @@ describe("Data Access Layer", function () {
       const query = { page: 1 };
       const result = await retrieveUsers({ query });
       removeSensitiveInfo(userData[12]);
-      result.allUsers.forEach((element) => {
+      result.users.forEach((element) => {
```
heyrandhir, line 68: can we use more reasonable name instead of `element`

test/unit/services/dataAccessLayer.test.js
```diff
@@ -146,4 +149,53 @@ describe("Data Access Layer", function () {
...
+    it("should set only email for INTERNAL access", function () {
+      const result = privilegedAccess(userData[11], data, AccessLevel.INTERNAL);
+      expect(result).to.have.property("email");
+    });
+
+    it("should set email and phone for PRIVATE access", function () {
+      const result = privilegedAccess(userData[11], data, AccessLevel.PRIVATE);
+      expect(result).to.have.property("email");
+      expect(result).to.have.property("phone");
+    });
```
heyrandhir, line 168: some users in the db dont have phone no added. can we write test for them too.

test/unit/services/dataAccessLayer.test.js
```diff
@@ -146,4 +149,53 @@ describe("Data Access Layer", function () {
...
+
+    it("should set email, phone, and chaincode for CONFIDENTIAL access", function () {
+      const result = privilegedAccess(userData[11], data, AccessLevel.CONFIDENTIAL);
+      expect(result).to.have.property("email");
+      expect(result).to.have.property("phone");
+      expect(result).to.have.property("chaincode");
+    });
+  });
+
+  describe("levelSpecificAccess", function () {
```
heyrandhir, line 178: why 2 different describe blocks . the previous describe block too was testing based on levels

heyrandhir (changes requested):
Can we write even integration tests for the this. I only see the old integration test modified. but I don't see any new added.

services/dataAccessLayer.js
heyrandhir: Better to handle null checks explicitly for clarity.

services/dataAccessLayer.js
```diff
@@ -70,11 +91,44 @@ const removeSensitiveInfo = function (obj) {
...
+  if (level === ACCESS_LEVEL.CONFIDENTIAL) {
+    if ("chaincode" in data) {
+      user.chaincode = data.chaincode;
+    }
+  }
+  return user;
+};
+
+const levelSpecificAccess = (user, level = ACCESS_LEVEL.PUBLIC, role = null) => {
+  const unFilteredData = JSON.parse(JSON.stringify(user));
```
prakashchoudhary07, line 112: Isn't this negating each other?

services/dataAccessLayer.js
```diff
@@ -70,11 +91,44 @@ const removeSensitiveInfo = function (obj) {
   }
 };
 
+const privilegedAccess = (user, data, level) => {
+  if ("email" in data) {
```
prakashchoudhary07, line 95: what is user and data here?

prakashchoudhary07 (changes requested):

services/dataAccessLayer.js
heyrandhir: you are stringifying and parsing it back again. both these operations will nullify each other.

bksh05 (approved):
LGTM

constants/userDataLevels.js
```diff
@@ -0,0 +1,21 @@
...
+  confidential: ["super_user"],
+};
+
+const ROLE_ACCESS = {
+  public: ["email", "phone", "chaincode"],
+  internal: ["phone", "chaincode"],
+  private: ["chaincode"],
+  confidential: [],
+};
+
```
heyrandhir, line 20: any specific reason for keeping the keys casing all caps in `ACCESS_LEVEL` and small case in `ROLE_LEVEL` & `ROLE_ACCESS`. we should maintain consistent across all 3 enums.

heyrandhir (changes requested):

heyrandhir (dismissed):

heyrandhir (approved):

iamitprakash (approved):

heyrandhir (approved):

prakashchoudhary07 (approved):

PR #1378, opened 2023-08-07
API endpoint to update Discord usernames for backend users

### LINKED ISSUE #1338: Update Discord servernames with rds usersname
# Create API endpoint to update Discord usernames for backend users

## Description
Currently, there is a need to synchronize the usernames of rds users in our Discord server with the usernames they choose for the backend. To achieve this, we need to create a new API endpoint in our backend, which can only be accessed by super users. This API will run a script that utilizes an external API to update the Discord usernames.

## Acceptance Criteria
1. Create a new API endpoint in the backend that can be accessed only by super users.
2. The API should send a PATCH request to the following endpoint: `rdsServerlessBaseUrl/guild/member`.
3. The request headers should include an `Authorization` field with the value `Bearer <Auth Token>` for authentication.
4. The request body should include the following parameters:
   - `discordId`: The Discord ID of the user whose username needs to be updated.
   - `userName`: The RDS username of the user.
5. When the super user runs this script, the backend should check if the user document contains a `discordId`.
6. If the `discordId` is present, the backend should call the external API mentioned above, passing the `discordId` and the corresponding `userName` to update the user's nickname in the Discord server.
7. The API should handle the following responses from the external API:
   - If the response contains the string "🚫 Bad Request Signature," it should be treated as a missing signature/auth header error.
   - If the response contains the string "Oops! We have encountered an internal Server Error," it should be treated as an internal server error.
   - If the response contains the string "User nickname changed successfully," it should be treated as a successful update.
   - In case the API returns any other response, it should be handled appropriately as well.
8. Since there are no status codes provided by the external API, we will rely on the response strings mentioned above for error and success detection.
9. Proper error handling should be implemented for different scenarios, and appropriate messages should be logged.
10. Unit tests should be created to verify the functionality of the new API endpoint and the script.

Please ensure that the external API (`discord-slash-commands` repository) mentioned in the acceptance criteria is properly set up and functional before integrating it with the backend. Additionally, implement super user authorization for accessing the new API endpoint to restrict it to authorized users only.

kotesh-arya:
### Issue
Closes #1338 

### Description:
Currently, there is a need to synchronize the usernames of rds users in our Discord server with the usernames they choose for the backend. To achieve this, we need to create a new API endpoint in our backend, which can only be accessed by super users. This API will run a script that utilizes an external API to update the Discord usernames.

###  Changes made:
- Added a API route to update the Nickname of the user in RDS discord server which runs a script and gets a user with matching userId that we pass as a parameter.
- The API sends a PATCH request to the following endpoint: `rdsServerlessBaseUrl/guild/member`.
- The request headers should include an `Authorization` field with the value `Bearer <Auth Token>` for authentication.

### Testing:
- The script and API endpoint have undergone comprehensive testing to ensure accurate identification of users with a linked Discord account and proper updating of their nickname with the username in RDS-backend. Testing includes various scenarios, such as users who are verified and unverified, to ensure the script behaves as expected.

### Is Development Tested?

- [x] Yes
- [ ] No

#### Before:
nickname-before

#### After:

updated-nickname

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -319,20 +319,13 @@ const verifyUserImage = async (req, res) => {
 
 const updateDiscordUserNickname = async (req, res) => {
   try {
-    // const token = req.cookies[config.get("userToken.cookieName")];
-    // console.log("TOKEN", token);
-    // console.log("params", req.params.userId);
     const userToBeUpdated = await dataAccess.retrieveUsers({ id: req.params.userId });
-    // console.log(userToBeUpdated);
     const { discordId, username: userName } = userToBeUpdated.user;
-    // console.log("DISCORD-ID", discordId, "user-name", userName);
     if (!discordId) {
       throw new Error("user not verified");
     }
     const response = await setUserDiscordNickname(userName, discordId);
-    // console.log(`discordId:${req.params.discordId} userName:${userToBeUpdated.username}`);
 
-    // Handling errors based on different responses is yet to be implemented after testing with a real user
     return res.json({
       userAffected: {
         userId: req.params.userId,
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -618,25 +618,6 @@ const setInDiscordScript = async (req, res) => {
   }
 };
 
-const removeTokens = async (req, res) => {
-  try {
-    const users = await userQuery.fetchUsersWithToken();
-
-    if (!users.length) {
-      return res.status(404).json({ message: "No users found with github Token!" });
-    }
-
-    await userQuery.removeGitHubToken(users);
-
-    return res.status(200).json({
-      message: "Github Token removed from all users!",
-      usersFound: users.length,
-    });
-  } catch (err) {
-    return res.boom.badImplementation({ message: INTERNAL_SERVER_ERROR });
-  }
-};
-
 const updateRoles = async (req, res) => {
   try {
     const result = await dataAccess.retrieveUsers({ id: req.params.id });
@@ -736,7 +717,6 @@ module.exports = {
   nonVerifiedDiscordUsers,
   setInDiscordScript,
   markUnverified,
-  removeTokens,
   updateRoles,
   updateDiscordUserNickname,
   archiveUserIfNotInDiscord,
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -323,18 +323,19 @@ const verifyUserImage = async (req, res) => {
  */
 
 const updateDiscordUserNickname = async (req, res) => {
+  const { userId } = req.params;
   try {
-    const userToBeUpdated = await dataAccess.retrieveUsers({ id: req.params.userId });
-    const { discordId, username: userName } = userToBeUpdated.user;
+    const userToBeUpdated = await dataAccess.retrieveUsers({ id: userId });
+    const { discordId, username } = userToBeUpdated.user;
     if (!discordId) {
       throw new Error("user not verified");
     }
-    const response = await setUserDiscordNickname(userName, discordId);
+    const response = await setUserDiscordNickname(username, discordId);
 
     return res.json({
       userAffected: {
-        userId: req.params.userId,
-        username: userName,
+        userId,
+        username,
         discordId,
       },
       message: response,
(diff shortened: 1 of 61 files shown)


controllers/users.js
```diff
@@ -315,6 +315,35 @@ const verifyUserImage = async (req, res) => {
...
 };
 
+/**
+ * Patch Update user nickname
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const updateDiscordUserNickname = async (req, res) => {
```
iamitprakash, line 325: setUserDiscordNickname() function expects the discordId parameter to be a string. However, the value of the discordId variable is the result of the userToBeUpdated.user.discordId expression, which could be null or undefined.

services/discordService.js
```diff
@@ -68,9 +68,28 @@ const removeRoleFromUser = async (roleId, discordId) => {
   }
 };
 
+const setUserDiscordNickname = async (userName, discordId) => {
```
iamitprakash, line 71: DISCORD_BASE_URL environment variable is not defined. This means that the fetch() function will not be able to connect to the Discord API.

To fix the bug, you need to define the DISCORD_BASE_URL environment variable. You can do this by adding the following line to your .env file:

iamitprakash (changes requested):

services/discordService.js
kotesh-arya: I am accessing it from the `services.discordBot.baseUrl` in the **config/local.js** file.

controllers/users.js
kotesh-arya: I am handling this case of `discordId` being null/undefined with this check and showing error:
 ```
if (!discordId) {
      throw new Error("user not verified");
    }
```

File: controllers/users.js [hunk 3]
```diff
@@ -315,6 +315,35 @@ const verifyUserImage = async (req, res) => {
...
+  try {
+    const userToBeUpdated = await dataAccess.retrieveUsers({ id: req.params.userId });
```
DashDeipayan, line 327: We can save this as a variable and use it 
`const {userId}=req.params`

`

     return res.json({
       userAffected: {
         userId
         username: userName,
         discordId,
       },
       message: response,
     });
`

controllers/users.js
```diff
@@ -315,6 +315,35 @@ const verifyUserImage = async (req, res) => {
...
+    const { discordId, username: userName } = userToBeUpdated.user;
+    if (!discordId) {
+      throw new Error("user not verified");
+    }
+    const response = await setUserDiscordNickname(userName, discordId);
+
+    return res.json({
+      userAffected: {
+        userId: req.params.userId,
+        username: userName,
```
DashDeipayan, line 337: why making `username` as `userName` when it is being used as it is by default?

middlewares/authenticate.js
```diff
@@ -41,6 +41,7 @@ const checkRestricted = async (req, res, next) => {
  */
 module.exports = async (req, res, next) => {
   try {
+    // console.log("from authenticate middleware");
```
DashDeipayan, line 44: remove console log comments

test/integration/users.test.js
```diff
@@ -1751,4 +1752,76 @@ describe("Users", function () {
...
+          status: 200,
+          json: () =>
+            Promise.resolve({
+              userAffected: {
+                userId: "X0H3paYveEWh7Q2fPhor",
+                username: "test-name-007",
+                discordId: "1123566257019568232",
+              },
+              message: "User nickname changed successfully",
+            }),
```
DashDeipayan, line 1776: put this in a fixture please

DashDeipayan (changes requested):
added comments

controllers/users.js
kotesh-arya: I thought about changing it's casing into camelCase, But we actually don't need to change it.
reverted it back to `username`

middlewares/authenticate.js
kotesh-arya: Done 👍🏻

test/integration/users.test.js
kotesh-arya: Created a fixture for this 👍🏻

DashDeipayan (approved):

iamitprakash (approved):

PR #1380, opened 2023-08-07
Add new route for Unique Username Generation 

### LINKED ISSUE #1377: Add new route for Unique Username Generation
# Add new route for Unique Username Generation 

## Overview

This ticket focuses on creating a new route, /user/generateUniqueUsername, that enables the system to generate a unique username based on user-provided first and last names. If the desired username is already in use, the system will iteratively append numerical suffixes until an available username is found. The generated username will then be returned to the frontend of user signup flow.
 

## User stories

- As a user, I want the system to automatically generate a unique username for me based on my first and last names
- As a developer, I want to create a new route to facilitate the generation of unique usernames without disrupting the existing production route (i.e. route GET `/users/isUsernameAvailable/:username`)

## Acceptance Criteria

- [x] TDD 
- [x] Develop a new HTTP GET route : `/users/isUsernameAvailable/` (end point may be get change(<a href="https://github.com/Real-Dev-Squad/website-api-contracts/issues/148">RFC</a>)
- [x] If the requested username (concatenated firstname-lastname) is available, respond with a JSON object containing the generated username.
- [x] If the username is already taken, generate a unique username by incrementally appending numerical suffixes ("-1", "-2", etc.) until an available username is found.
- [x] Return the generated unique username as part of the JSON response.
- [x] Ensure that the existing production route for checking username availability (`/users/isUsernameAvailable/:username`) remains functional and unaffected by this implementation.
- [x] Update the system's API documentation to include details about the newly added route 

## Estimated Time to Complete:

- [x] New Route Implementation: 2-4 hours
- [x] Username Generation Logic: 3-5 hours
- [x] Test: 2 hours

shubhamsigdar1:
#####  Closes  #1377

Issue dependent on this pr: - https://github.com/Real-Dev-Squad/website-my/issues/474

- RFC for API contract: https://github.com/Real-Dev-Squad/website-api-contracts/issues/148

### What is the change?

- Develop a new HTTP GET route : /users/username/ (end point may be get change(<a href= "https://github.com/Real-Dev-Squad/website-api-contracts/issues/148">RFC </a> ))
- Sending from frontend firstname and lastname as a query parameter i.e. `https://api.realdevsquad.com/users/username?firstname=shubham&lastname=sigdar`
- Added logic to generate unique username 
- return the generated unique username as part of the JSON response.

### Is Development Tested?

- [x] Yes
- [ ] No

### Note to Reviewers:
- The feature is under a feature flag.

### Test stats

**Before:** 

 website-backend/models/users.js
image

**After:** 

 website-backend/models/users.js
image

### After Change Video

https://github.com/Real-Dev-Squad/website-backend/assets/107163260/a63be4a8-21e0-476d-96c8-875a2bd7801f

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -15,6 +15,11 @@ const { addRoleToUser, getDiscordMembers } = require("../services/discordService
 const { fetchAllUsers } = require("../models/users");
 const { getQualifiers } = require("../utils/helper");
 const { getFilteredPRsOrIssues } = require("../utils/pullRequests");
+const {
+  USERS_PATCH_HANDLER_ACTIONS,
+  USERS_PATCH_HANDLER_ERROR_MESSAGES,
+  USERS_PATCH_HANDLER_SUCCESS_MESSAGES,
+} = require("../constants/users");
 
 const verifyUser = async (req, res) => {
   const userId = req.userData.id;
@@ -101,9 +106,10 @@ const getUsers = async (req, res) => {
     }
 
     const data = await dataAccess.retrieveUsers({ query: req.query });
+
     return res.json({
       message: "Users returned successfully!",
-      users: data.allUsers,
+      users: data.users,
       links: {
         next: data.nextId ? getPaginationLink(req.query, "next", data.nextId) : "",
         prev: data.prevId ? getPaginationLink(req.query, "prev", data.prevId) : "",
@@ -232,10 +238,9 @@ const getUsername = async (req, res) => {
 const getSelfDetails = async (req, res) => {
   try {
     if (req.userData) {
-      if (req.query.private) {
-        return res.send(req.userData);
-      }
-      const user = await dataAccess.retrieveUsers({ userdata: req.userData });
+      const user = await dataAccess.retrieveUsers({
+        userdata: req.userData,
+      });
       return res.send(user);
     }
     return res.boom.notFound("User doesn't exist");
@@ -434,6 +439,7 @@ const updateUser = async (req, res) => {
 const generateChaincode = async (req, res) => {
   try {
     const { id } = req.userData;
+
     const chaincode = await chaincodeQuery.storeChaincode(id);
     await userQuery.addOrUpdate({ chaincode }, id);
     return res.json({
@@ -596,9 +602,9 @@ const filterUsers = async (req, res) => {
   }
 };
 
-const nonVerifiedDiscordUsers = async (req, res) => {
+const nonVerifiedDiscordUsers = async () => {
   const data = await dataAccess.retrieveDiscordUsers();
-  return res.json(data);
+  return data;
 };
 
 const setInDiscordScript = async (req, res) => {
@@ -652,6 +658,58 @@ const updateRoles = async (req, res) => {
   }
 };
 
+const archiveUserIfNotInDiscord = async () => {
+  try {
+    const data = await userQuery.archiveUserIfNotInDiscord();
+
+    if (data.totalUsers === 0) {
+      return {
+        message: USERS_PATCH_HANDLER_ERROR_MESSAGES.ARCHIVE_USERS.NO_USERS_DATA_TO_UPDATE,
+        summary: data,
+      };
+    }
+
+    return {
+      message: USERS_PATCH_HANDLER_SUCCESS_MESSAGES.ARCHIVE_USERS.SUCCESSFULLY_UPDATED_DATA,
+      summary: data,
+    };
+  } catch (error) {
+    logger.error(`Error while updating the archived role: ${error}`);
+    throw Error(INTERNAL_SERVER_ERROR);
+  }
+};
+
+async function usersPatchHandler(req, res) {
+  try {
+    const { action } = req.body;
+    let response;
+
+    if (action === USERS_PATCH_HANDLER_ACTIONS.NON_VERFIED_DISCORD_USERS) {
+      const data = await nonVerifiedDiscordUsers();
+      response = data;
+    }
+
+    if (action === USERS_PATCH_HANDLER_ACTIONS.ARCHIVE_USERS) {
+      const debugQuery = req.query.debug?.toLowerCase();
+      const data = await archiveUserIfNotInDiscord();
+
+      if (debugQuery === "true") {
+        data.summary.updatedUserDetails = data.summary.updatedUserDetails.slice(-3);
+        response = data;
+      } else {
+        delete data.summary.updatedUserDetails;
+        delete data.summary.failedUserDetails;
+        response = data;
+      }
+    }
+
+    return res.status(200).json(response);
+  } catch (error) {
+    logger.error("Error while handling the users common patch route:", error);
+    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
+  }
+}
+
 module.exports = {
   verifyUser,
   generateChaincode,
@@ -679,4 +737,6 @@ module.exports = {
   markUnverified,
   removeTokens,
   updateRoles,
+  archiveUserIfNotInDiscord,
+  usersPatchHandler,
 };
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -201,7 +201,7 @@ const getUsernameAvailabilty = async (req, res) => {
   }
 };
 
-const getUsername = async (req, res) => {
+const generateUsername = async (req, res) => {
   try {
     const { firstname, lastname, dev } = req.query;
     if (dev === "true") {
@@ -719,7 +719,7 @@ module.exports = {
   getSelfDetails,
   getUser,
   getUsernameAvailabilty,
-  getUsername,
+  generateUsername,
   getSuggestedUsers,
   postUserPicture,
   updateUser,
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -201,20 +201,16 @@ const getUsername = async (req, res) => {
     if (dev) {
       const baseUsername = `${firstname.toLowerCase()}-${lastname.toLowerCase()}`;
       let newUsername = baseUsername;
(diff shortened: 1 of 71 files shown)


test/integration/users.test.js
```diff
@@ -719,6 +719,27 @@ describe("Users", function () {
...
+            return done();
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.username).to.equal("ankur-narkhede");
+
+          return done();
+        });
+    });
+  });
```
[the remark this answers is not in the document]
shubhamsigdar1: Done please check my latest commit

controllers/users.js
```diff
@@ -195,6 +195,37 @@ const getUsernameAvailabilty = async (req, res) => {
...
 
+const getUsername = async (req, res) => {
+  try {
+    const { firstname, lastname, dev } = req.query;
+    if (dev) {
+      const baseUsername = `${firstname.toLowerCase()}-${lastname.toLowerCase()}`;
+      let newUsername = baseUsername;
+      let count = 1;
+
+      while (true) {
```
apurvkhare, line 206: is there no other way then `while(true)` ?

shubhamsigdar1: Done with changes @apurvkhare...can you please check latest commit

shubhamsigdar1: But @apurvkhare I am going to change my approach because in this approach I am accessing the database every time to check for the existence of a username but accessing database every time is not cost effective 
So my new approach is  access database once and fetch all user that matches the given firstname and lastname.

apurvkhare: What I see now is `for` with no end condition which is the same, are you working on updating this or that's the final commit?

khalatevarun (approved):

routes/users.js
```diff
@@ -17,6 +17,7 @@ router.patch("/self", authenticate, userValidator.updateUser, users.updateSelf);
 router.get("/", userValidator.getUsers, users.getUsers);
 router.get("/self", authenticate, users.getSelfDetails);
 router.get("/isUsernameAvailable/:username", authenticate, users.getUsernameAvailabilty);
```
rohan09-raj, line 19: The API name can be corrected to `/users/username/:username` to better

routes/users.js
```diff
@@ -17,6 +17,7 @@ router.patch("/self", authenticate, userValidator.updateUser, users.updateSelf);
...
+router.get("/generateusername", authenticate, users.getUsername);
```
rohan09-raj, line 20: I think it would be better if this is a POST route as here we are doing some action. GET is generally used for querying the data and giving it in a specific format

Route can be `/users/username`

routes/users.js
shubhamsigdar1: sorry for late response. My health is not good last week. I am OOO

Regarding your comment, This API route will get removed

routes/users.js
shubhamsigdar1: I want to discuss with you regarding this comment.

Hi @rohan09-raj first of all thank you for reviewing my pr, If you check this RFC After change API contract (<a href=" https://github.com/Real-Dev-Squad/website-api-contracts/issues/148">RFC</a>) I am not modifying the database or creating any new resources when generating a unique username. Instead, I am performing a read-only operation to calculate and provide a username to frontend based on existing data. In such a case, using a GET request in route is consistent with the principles of RESTful API design, where GET requests are used for retrieving resources without making any modifications.

Thank you for route endpoint name

controllers/users.js
```diff
@@ -201,6 +201,23 @@ const getUsernameAvailabilty = async (req, res) => {
...
 
+const getUsername = async (req, res) => {
+  try {
+    const { firstname, lastname, dev } = req.query;
+    if (dev === "true") {
+      const username = await userQuery.generateUniqueUsername(firstname, lastname);
+      return res.json({ username });
+    } else {
+      return res.status(404).json({
+        message: "Data Not Found",
```
vinit717, line 212: message: "UserName Not Found"

test/integration/users.test.js
vinit717, line 1: where is unit test?

shubhamsigdar1: Is it important to right unit test because I already written integration test.

test/integration/users.test.js
```diff
@@ -694,6 +694,45 @@ describe("Users", function () {
...
+            return done();
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.username).to.equal("shubham-sigdar");
+
+          return done();
+        });
+    });
+
```
prakashchoudhary07, line 717: PLease write test for case when username is already present in DB

controllers/users.js
```diff
@@ -201,6 +201,23 @@ const getUsernameAvailabilty = async (req, res) => {
   }
 };
 
+const getUsername = async (req, res) => {
```
iamitprakash, line 204: here dev query parameter is a string. But the === operator only compares the values of two variables, not their types. So, the expression dev === "true" will always evaluate to true, even if the value of dev is not actually the string "true".

models/users.js
```diff
@@ -715,6 +715,27 @@ const getUsersByRole = async (role) => {
   }
 };
 
+const generateUniqueUsername = async (firstname, lastname) => {
```
iamitprakash, line 718: I cannot find any bugs in the code snippet you have provided. The code seems to be working as intended.

The function generateUniqueUsername() takes two parameters: the user's first name and last name. It first checks if there is already a user with the same first name and last name in the database. If there is, it increments the user count by 1 and appends that number to the username. Otherwise, it simply returns the username as-is.

The code snippet you have provided does not have any errors. However, there are some potential improvements that could be made. For example, the code could be made more efficient by using a parameterized query instead of a hardcoded query. Additionally, the code could be made more robust by handling errors more gracefully.

Here is an example of how the code could be improved:

const generateUniqueUsername = async (firstname, lastname) => {
  try {
    const query = userModel.where({
      first_name: firstname,
      last_name: lastname,
    });
    const count = await query.count();
    const initialUsername = `${firstname}-${lastname}`;
    const uniqueUsername = count === 0 ? initialUsername : `${initialUsername}-${count + 1}`;
    return uniqueUsername;
  } catch (err) {
    logger.error(`Error while generating unique username: ${err.message}`);
    throw err;
  }
};

iamitprakash (changes requested):

controllers/users.js
shubhamsigdar1: Hi @iamitprakash Thank you for reviewing my pr

=== Operator  only returns true if both the value of the variables (operands) and the data type of the variables are the same, else false. 
If the types of the operands are different, the === operator will return false
true === "true" (false)

models/users.js
shubhamsigdar1: Hi tejas, the code snippet you provided is not actually running and its give error
then I research and found firestore does not support parameterized query in the way you described and actually there is no concept of parameterized query in firestore (both hardcoded and parameterized query idea same in firestore)

I want to achieve this
RFC: https://github.com/Real-Dev-Squad/website-backend/issues/1407
image

To achieve above 

I found code directly from firestore docs- https://firebase.google.com/docs/firestore/query-data/aggregation-queries
image

Firestore has a billing optimization that allows you to be charged for one document read operation if your count() operation matches between 0 and 1000 index entries.

controllers/users.js
shubhamsigdar1: Done Can you please check latest commit changes

test/integration/users.test.js
shubhamsigdar1: Done with changes can you please check latest commit

test/integration/users.test.js
shubhamsigdar1: @vinit717 written unit test can you please check latest commit

prakashchoudhary07 (approved):

vinit717 (dismissed):

controllers/users.js
pallabez, line 204: I'd suggest 
```suggestion
const generateUsername = async (req, res) => {
```

routes/users.js
```diff
@@ -17,6 +17,7 @@ router.patch("/self", authenticate, userValidator.updateUser, users.updateSelf);
 router.get("/", userValidator.getUsers, users.getUsers);
 router.get("/self", authenticate, users.getSelfDetails);
 router.get("/isUsernameAvailable/:username", authenticate, users.getUsernameAvailabilty);
+router.get("/username", authenticate, userValidator.getUsername, users.getUsername);
```
pallabez, line 20: Not a requirement now but would suggest - `/username-generation` for future

prakashchoudhary07 (approved):

controllers/users.js
shubhamsigdar1: Done with changes can you please check latest commit

middlewares/validators/user.js
```diff
@@ -281,6 +281,32 @@ async function validateUsersPatchHandler(req, res, next) {
...
 }
 
+/**
+ * Validates query params for the username route
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ * @param next {Object} - Express middelware function
+ */
+const getUsername = async (req, res, next) => {
```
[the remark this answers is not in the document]
shubhamsigdar1: Done with changes can you please check latest commit

routes/users.js
shubhamsigdar1: Okk @pallabez Thank you

vinit717 (approved):

Achintya-Chatterjee (approved):
LGTM

pallabez (approved):
LGTM. Thanks for the changes

Issue #1366, opened 2023-08-05
Update Username validator to accommodate new requirement


shubhamsigdar1:
# Update Username validator to accommodate new requirement
## Overview

Modify the existing username validator in the backend to accommodate the new username format.

Issue ticket dependent on this task : 
- https://github.com/Real-Dev-Squad/website-my/issues/465

## User stories

- As a user, my username should be like 'shubham-sigdar', 'shubham-sigdar-1', 'shubham-sigdar-2' 
- As a user, my username contain hypens, in lowercase,  any length without being restricted by a character limit, contain number i.e.  contains only lowercase letters, hyphens, and numbers.

## Acceptance Criteria

- [x] TDD
- [x] The updated username validator must allow hyphens (-) in usernames and not any other special character.
- [x] Usernames can consist only lowercase alphanumeric characters.
- [x] Usernames can contain number.
- [x] There is no specific character limit for usernames; they can be of any length.

PR #1384, opened 2023-08-07
Fix/username validator

Abhay5855:
# Closes:
#1366 

## Overview:
While generating the username It should follow a specific format.

## Changes Proposed:
- As a user, my username should be like 'shubham-sigdar', 'shubham-sigdar-1', 'shubham-sigdar-2' 
- As a user, my username contains hyphens, in lowercase,  any length without being restricted by a character limit, and contain number i.e. contains only lowercase letters, hyphens, and numbers.

## Test Coverage:
**Integration Test**
image

**Unit Test**
image

## Reviewer Checklist:
- [ ] Ensure the API handles error cases appropriately and provides informative error messages.
- [ ] Test the validations once regarding the username.
- [ ] Review the code for adherence to coding standards and best practices.

### DIFF
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -17,7 +17,7 @@ const updateUser = async (req, res, next) => {
         .string()
         .optional()
         .regex(/^[a-z0-9-]+$/)
-        .message("Username must be lowercase only hypen, numbers are allowed"),
+        .message("Username must be lowercase only hypen, numbers are allowed."),
       first_name: joi.string().optional(),
       last_name: joi.string().optional(),
       yoe: joi.number().min(0).optional(),
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -16,10 +16,8 @@ const updateUser = async (req, res, next) => {
       username: joi
         .string()
         .optional()
-        .min(4)
-        .max(20)
-        .regex(/^[a-zA-Z0-9]+$/)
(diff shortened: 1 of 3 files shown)


middlewares/validators/user.js
```diff
@@ -16,10 +16,8 @@ const updateUser = async (req, res, next) => {
       username: joi
         .string()
         .optional()
-        .min(4)
-        .max(20)
-        .regex(/^[a-zA-Z0-9]+$/)
-        .message("Username must be between 4 and 20 characters long and contain only letters or numbers."),
+        .regex(/^[a-z0-9-]+$/)
```
shubhamsigdar1, line 19: Hello @Abhay5855, I want that username validator should enforce the specific format of usernames
i.e. Usernames should start with first name,  followed by a hyphen, then the last name.
ex:  username validator enforces the required username format of "firstname-lastname" or "firstname-lastname-1"

Is it possible
Like this below regex
image

So if hypen is before firstname it is invalid username or number before firstname is invalid username

shubhamsigdar1 (commented):
great work @Abhay5855, Please check one of my comment if you confused please ping me on discord

middlewares/validators/user.js
Abhay5855: @shubhamsigdar1 That's great thinking I have made the validation as required such that firstname-lastname-1 or firstname-lastname can be included rest will throw an error.

Issue #1374, opened 2023-08-07
[Live-Site] Feat/event kickout peer api


prerana1821:
# Parent Story

https://github.com/Real-Dev-Squad/website-www/issues/343

# Description
Need an API to allow host to kick out peer from the event.

<html>
<body>

Method | Route | Description
-- | -- | --
PATCH | :id/peers/kickout | Removes peer from the event & update the database.

</body>
</html>

PR #1375, opened 2023-08-07
[Live-Site] feat: added api to kickout peer

prerana1821:
# Issues

Fixes #1374 

# Added API:

- [x] PATCH  - /:id/peers/kickout - Removes peer from the event & update the database.

PR #1387, opened 2023-08-08
[Live-Site] Test/event kickout peer api

prerana1821:
# Description:
Tests for kickout peer API for live site.

PATCH - /:id/peers/kickout - Removes peer from the event & update the database.

# Main PR
https://github.com/Real-Dev-Squad/website-backend/pull/1375

# More details:
Fixes: https://github.com/Real-Dev-Squad/website-backend/issues/1374


iamitprakash (changes requested):
Change accordingly as per main PR

prerana1821:
> 

Yes, done. Thank you

Pratiyushkumar (approved):

vvaibhavdesai (changes requested):

iamitprakash (approved):

controllers/events.js
```diff
@@ -193,11 +195,77 @@ const endActiveEvent = async (req, res) => {
...
+    });
+    return res.status(200).json({
+      data,
+      message: `Selected Participant is removed from event.`,
+    });
+  } catch (error) {
+    logger.error({ error });
+    return res.status(500).json({
+      error: error.code,
+      message: "You can't remove selected Participant from Remove, Please ask Admin or Host for help.",
```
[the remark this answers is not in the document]
prerana1821: Yes, I have fixed it.

controllers/events.js
```diff
@@ -193,11 +200,77 @@ const endActiveEvent = async (req, res) => {
...
+ */
+const kickoutPeer = async (req, res) => {
+  const { id } = req.params;
+  const payload = {
+    peer_id: req.body.peerId,
+    reason: req.body.reason,
+  };
+
+  try {
+    await apiService.post(`/active-rooms/${id}/remove-peers`, payload);
```
vvaibhavdesai, line 253: I will suggest to keep all the strings in constant 
1. easier to maintain
2. easier to replace
3. you remember constant names more often than long strings so they are easier to navigate and search

controllers/events.js
```diff
@@ -193,11 +200,77 @@ const endActiveEvent = async (req, res) => {
...
+    await eventQuery.kickoutPeer({ eventId: id, peerId: payload.peer_id, reason: req.body.reason });
+    return res.status(200).json({
+      message: SUCCESS_MESSAGES.CONTROLLERS.KICKOUT_PEER,
```
vvaibhavdesai, line 256: nicely done 👌

controllers/events.js
prerana1821: Thanks a lot for your review, @vvaibhavdesai.
I have made all the strings as constant.

controllers/events.js
```diff
@@ -15,21 +23,21 @@ const apiService = new EventAPIService(tokenService);
...
+ * @returns {Object} If an error occurs while creating the event document.
  */
 const createEvent = async (req, res) => {
   const { name, description, region, userId } = req.body;
   const payload = { name, description, region };
   try {
-    const eventData = await apiService.post("/rooms", payload);
+    const eventData = await apiService.post(API_URLS.CREATE_EVENT, payload);
     const event = removeUnwantedProperties(UNWANTED_PROPERTIES_FROM_100MS, eventData);
     const eventDataFromDB = await eventQuery.createEvent({ room_id: eventData.id, created_by: userId, ...event });
```
iamitprakash, line 34: do we need to save room_id in DB..?

controllers/events.js
```diff
@@ -86,7 +94,7 @@ const getAllEvents = async (req, res) => {
  * @param {Object} req - The Express request object.
  * @param {Object} res - The Express response object.
  * @returns {Object} An object containing the generated token and a success message in JSON format.
- * @throws {Error} If an error occurs while generating the token.
+ * @returns {Object} If an error occurs while generating the token.
  */
 const joinEvent = async (req, res) => {
```
iamitprakash, line 99: function is not checking the role property of the req.body object. This means that the function will allow users to join events with any role, even if they are not supposed to have access to the event.

is my understanding correct..?

controllers/events.js
```diff
@@ -173,22 +182,86 @@ const updateEvent = async (req, res) => {
  * @param {Object} req - The Express request object.
  * @param {Object} res - The Express response object.
  * @returns {Promise<Object>} The JSON response with a message indicating the session has ended.
- * @throws {Object} The JSON response with an error message if an error occurred while ending the event.
+ * @returns {Promise<Object>} The JSON response with an error message if an error occurred while ending the event.
  */
 const endActiveEvent = async (req, res) => {
```
iamitprakash, line 187: endActiveEvent() function is not checking the lock property of the req.body object. This means that the function will allow users to end active events, even if the event is locked.

 to fix this change to below
if (payload.lock === true) {
  return res.status(403).json({
    error: 'Forbidden',
    message: 'The event is locked and cannot be ended',
  });
}

iamitprakash (changes requested):

PR #1400, opened 2023-08-11
Verify status validation1

shreya-mishra:
- [x] created new route called /get-qrcode-device-info?user_id = {user_id}
- [x] functions related to that which is filtering according to the authorization status
- [x] Test


DashDeipayan (approved):

lakshayman (approved):
Looks good to me

DashDeipayan (approved):

PR #1411, opened 2023-08-13
feat/oooCommand: route added to get user by discordId for oooCommand

palakgupta2712:
For /ooo command to work in discord-slash-command, have created a new route -
`router.get("/discordId/:discordId", users.getUserByDiscordId);`
it will take discordId and will be returning data of that particular discordId user.

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -146,6 +146,24 @@ const getUser = async (req, res) => {
   }
 };
 
+const getUserByDiscordId = async (req, res) => {
+  try {
+    const discordId = Number(req.params.discordId);
+    const result = await userQuery.fetchUser({ discordId });
+    const user = result.user;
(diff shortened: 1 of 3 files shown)


sahsisunny:
Add test coverage of your changes and link issue ticket as well.

controllers/users.js
```diff
@@ -146,6 +146,24 @@ const getUser = async (req, res) => {
   }
 };
 
+const getUserByDiscordId = async (req, res) => {
+  try {
+    const discordId = Number(req.params.discordId);
```
lakshayman, line 151: NIT: Are we not using any validator, as here we are converting it explicitly?

lakshayman (changes requested):

lakshayman:
Few tests are failing

PR #1414, opened 2023-08-13
Refactor: keeping body optional in archiving details in the archiving process

gauravsinhaweb:
**Why do we need to keep the body (reason) optional to the issue/ ticket #1214 ?**

- If we don't keep the body optional, the current archiving process will break because we don't have to Input field in the front-end for the body `reason`,

In the future scope, we will make the body(reason) a required field.

This PR aims to make the 
body an optional field to not break the current working process in prod, and the body (reason) can be passed optional, and yet the authorized users can fetch the logs of `archived-details`.

**Test case coverage and result:**

Screenshot 2023-08-14 043647
Screenshot 2023-08-12 004131
Screenshot 2023-08-12 004109
Screenshot 2023-08-14 042811

### DIFF
--- a/controllers/members.js
+++ b/controllers/members.js
@@ -87,18 +87,13 @@ const archiveMembers = async (req, res) => {
     const superUserId = req.userData.id;
     const { reason } = req.body;
     const roles = req?.userData?.roles;
-    const isReasonNullOrUndefined = !reason;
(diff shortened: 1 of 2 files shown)


iamitprakash (approved):

bksh05 (approved):

bhtibrewal (approved):
LGTM

PR #1417, opened 2023-08-14
added a timestamp to tasks model

kavyarishi19:


models/tasks.js
```diff
@@ -27,9 +27,13 @@ const updateTask = async (taskData, taskId = null) => {
         taskData = { ...taskData, endsOn: Math.floor(Date.now() / 1000) };
       }
       const { dependsOn, ...taskWithoutDependsOn } = taskData;
+      const updatedAt = Date.now();
+      const createdAt = task.data().createdAt ? { createdAt: task.data().createdAt } : {};
```
ankushdharkar, line 31: Why is this an object? Should it not be a string?

PR #1434, opened 2023-08-21
[Live-Site] feat: create get event codes API

satyam73:
PR for live-site get event code for particular API.
More description will be added after it gets done.


middlewares/validators/events.js
```diff
@@ -165,6 +165,24 @@ const generateEventCode = async (req, res, next) => {
...
+  const { id } = req.params;
+
+  const schema = joi.object({
+    id: joi.string().required(),
+  });
+
+  const validationOptions = { abortEarly: false };
+
+  try {
+    await schema.validateAsync({ id }, validationOptions);
```
iamitprakash, line 178: next() function is called before the res.boom.badRequest() function. This means that the res.boom.badRequest() function will not be executed if the validation fails.

models/events.js
```diff
@@ -237,11 +237,46 @@ const createEventCode = async (eventCodeData) => {
   }
 };
 
+const getEventById = async ({ id }) => {
```
iamitprakash, line 240: eventSnapshotData variable is not being checked for null before it is returned. This means that if the event with the given ID does not exist, the eventSnapshotData variable will be null and the function will return null.

models/events.js
```diff
@@ -237,11 +237,46 @@ const createEventCode = async (eventCodeData) => {
...
+    const eventSnapshot = await eventRef.get();
+    const eventSnapshotData = eventSnapshot.data();
+    return eventSnapshotData;
+  } catch (error) {
+    logger.error("Error in getting event by id from the database.", error);
+    throw error;
+  }
+};
+
+const getEventCodes = async ({ id }) => {
```
iamitprakash, line 252: eventCodesArrayPromises array is not being checked for empty before it is passed to the Promise.all() function. This means that if the array is empty, the Promise.all() function will still wait for all the promises to resolve, even though there are no promises.

iamitprakash (changes requested):

middlewares/validators/events.js
satyam73: but the validateAsync function will throw error if validation fails resulting in going to catch block

models/events.js
satyam73: Yes it will return null but if you will see the ln 255 I have added a check for if eventData is null then it will throw an error.

models/events.js
satyam73: resolved this one

iamitprakash (approved):

Issue #1435, opened 2023-08-22
Role check middleware for events


prerana1821:
## authorizeEventRoles

**Host**
  request --> check user collectin roles if( issuperuser ) --> join
**Moderator**
  request --> user.roles.moderator === true --> join 
**Maven**  
  check code ["", "" , ""].includes('req.body.eventCode');  --> join
**Guest**   
  select role --> join

PR #1436, opened 2023-08-22
[Live-Site] authorizeEventRoles middleware 

prerana1821:
## Closes PR
https://github.com/Real-Dev-Squad/website-backend/issues/1435

## Description
Added authorizeEventRoles & getAllEventCodes model

### DIFF
--- a/test/unit/middlewares/authorizeEventRoles.test.js
+++ b/test/unit/middlewares/authorizeEventRoles.test.js
@@ -36,7 +36,7 @@ describe("authorizeEventRoles", function () {
 
   it("should call next if user has valid role and valid body role", async function () {
     req.body.role = ROLES.HOST;
-    req.userData.roles = { super_user: true, member: true };
+    req.userData.roles = { super_user: true, moderator: true };
 
     await authorizeEventRoles(["host", "moderator"])(req, res, next);
 
@@ -45,7 +45,7 @@ describe("authorizeEventRoles", function () {
 
   it("should call next if user has valid role and valid body role (MODERATOR)", async function () {
     req.body.role = ROLES.MODERATOR;
-    req.userData.roles = { super_user: true, member: true };
+    req.userData.roles = { super_user: true, moderator: true, member: true };
 
     await authorizeEventRoles(["host", "moderator"])(req, res, next);
 


models/events.js
```diff
@@ -237,10 +238,35 @@ const createEventCode = async (eventCodeData) => {
   }
 };
 
+const getAllEventCodes = async (roomId) => {
```
iamitprakash, line 241: getAllEventCodes function is declared as a const function, but it is using the async and await keywords. This is not allowed, because const functions cannot be asynchronous.

iamitprakash (changes requested):

models/events.js
prerana1821: As per I know, I think we can use const. You can check in the examples here as well:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/async_function

Please let me know if I'm missing anything.

iamitprakash (approved):

controllers/events.js
```diff
@@ -96,10 +97,16 @@ const joinEvent = async (req, res) => {
   const payload = { roomId, userId, role };
```
satyam73, line 97: we will not need roomId  from body now because we're getting from the API itself.
So send that roomId in payload

routes/events.js
```diff
@@ -3,15 +3,45 @@ const router = express.Router();
...
+router.post("/join", authorizeEventRoles([ROLES.MAVEN, ROLES.GUEST]), eventsValidator.joinEvent, events.joinEvent);
 
-router.post("/", authenticate, eventsValidator.createEvent, events.createEvent);
-router.get("/", eventsValidator.getAllEvents, events.getAllEvents);
-router.post("/join", eventsValidator.joinEvent, events.joinEvent);
 router.get("/:id", eventsValidator.getEventById, events.getEventById);
-router.patch("/", authenticate, eventsValidator.updateEvent, events.updateEvent);
-router.patch("/end", authenticate, eventsValidator.endActiveEvent, events.endActiveEvent);
+
+router.patch("/", authenticate, authorizeRoles([SUPERUSER]), eventsValidator.updateEvent, events.updateEvent);
```
satyam73, line 27: moderator can also have the ability to update the event

routes/events.js
```diff
@@ -3,15 +3,45 @@ const router = express.Router();
...
+
+router.patch("/end", authenticate, authorizeRoles([SUPERUSER]), eventsValidator.endActiveEvent, events.endActiveEvent);
```
satyam73, line 29: moderator can also have the ability to update the event

models/events.js
satyam73, line 1: I already merged PR for this one why this is created again?

middlewares/authorizeEventRoles.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+
+    if (req.body.role === ROLES.MAVEN) {
+      const roomId = req.body.roomId ?? req.params.id;
+      const allowedEventCodes = await getAllEventCodes(roomId);
+      if (allowedEventCodes.includes(req.body.eventCode) && userHasPermission(req.body.role, allowedRoles)) {
+        return next();
+      }
+    }
```
satyam73, line 38: take latest pull and use the changes of mine that has been merged before for getEventCodes

satyam73 (changes requested):

Issue #1432, opened 2023-08-21
Goals Backend Authentication API Creation


ivinayakg:
# Brief
- ![image](https://github.com/Real-Dev-Squad/website-goals/assets/63191499/ebd5e82d-e96c-44c6-8cb9-0af0af576c2c)
- This is the flow for the authentication of the goals-site.
- A authenticated user will request on the backend (main backend) and from there the backend will call the goals-site backend internally authenticate itself and send user info like `user_id, roles` with a `GOAL_BACKEND_SECRET_KEY` all in the body only and then will return the token for goals-backend `goal-site-token`, which will be used by the user to talk to goal site directly.

## What we need
- We need a simple API, which will call the goal site API with `user_id, roles, GOAL_BACKEND_SECRET_KEY`, and then return the token which can be used to access he goal site backend.

pallabez:
Only personal goal backend is deployed, so we still need to figure out how we can make the connection.

ivinayakg:
# ETA - 28th August 2023

pallabez:
New issue - Tests are failing cause nock is not working

skv93-coder:
Please assign this task to me

PR #1445, opened 2023-08-23
feat: goal site auth api (initial)

ivinayakg:
# RFC - #1432 

## Brief
- We needed a simple API, that would call the goal site API with user_id, roles, GOAL_BACKEND_SECRET_KEY, and then return the token which can be used to access the goal site backend.

### Integration tests
image

### Unit tests
image


controllers/goals.js
```diff
@@ -0,0 +1,44 @@
...
+    const { roles, id: userId } = req.userData;
+    const goalSiteConfig = config.services.goalAPI;
+
+    const goalApiResponse = await fetch(`${goalSiteConfig.baseUrl}/auth_token`, {
+      method: "POST",
+      body: JSON.stringify({
+        user_id: userId,
+        roles,
+        goal_api_secret_key: goalSiteConfig.secretKey,
+      }),
```
pallabez, line 15: Why are we using stringify here?

ivinayakg: Using the browser fetch api, there we have to send json, right?

controllers/goals.js
```diff
@@ -0,0 +1,36 @@
...
+            roles: roles,
+          },
+        },
+      }),
+      headers: { "Content-Type": "application/vnd.api+json", "Rest-Key": goalSiteConfig.secretKey },
+    });
+
+    const goalApiData = await goalApiResponse.json();
+
+    if (goalApiData) {
```
iamitprakash, line 24: what about below code

if (goalApiResponse.status === 200) {
  const goalApiData = await goalApiResponse.json();
  return res.status(200).json({ message: "success", user: { ...data.attributes, id: data.id } });
} else {
  return res.status(goalApiResponse.status).json({ message: "error" });
}

iamitprakash (changes requested):

controllers/goals.js
ivinayakg: sure

pallabez (approved):
LGTM

ivinayakg:
> LGTM. The workflow test is failing tho.

just checked.
they are running on the local, not sure why failing here

shubhamsigdar1:
Test is failing @ivinayakg

PR #140, opened 2021-01-24
Added /isUsernameAvailable/:username route in users main route

saishankarmanugula:
Added route to check user availability. Added swagger definition and integration test for it.

image

image

PR #139, opened 2021-01-23
Added the tasks integration tests

ankurnarkhede:
- Added the tasks integration tests back in the project that were removed in https://github.com/Real-Dev-Squad/website-backend/pull/137

PR #1457, opened 2023-08-28
Fetch User Details from Discord Id

j24m:
## Issue Ticket : 
- https://github.com/Real-Dev-Squad/discord-slash-commands/issues/106
## PR 
- To be checked along with this feature :  
https://github.com/Real-Dev-Squad/discord-slash-commands/pull/132
## Description
- This pull request adds a new feature to the **users** module that allows users to be retrieved by their Discord ID.
- The following changes have been made : 
-> The ``getUsers( )`` middleware has been updated to support the ``discordId`` query parameter.
-> The ``retrieveUsers( )`` function checks if the discordId parameter is present in the request data.
-> The ``fetchUser( )`` service has been updated to support the ``discordId`` option.
## Tests
- Unit and integration tests have been updated to cover the new functionality.
**Unit test stats(Before) :**
image

**Unit test stats(After) :** 
image
**Integration test stats(Before) :**
image

**Integration test stats(After) :** 
image

## Working proof 
- API to fetch User details by passing discord Id as query params : localhost:3000/users/?**dev=true&discordId=504855562094247953**

https://github.com/Real-Dev-Squad/website-backend/assets/53934353/a9682cb2-e3eb-46cc-b544-4c22148ade89


iamitprakash (changes requested):
Why coverage is so low, can you please add before code change stats as well

iamitprakash (changes requested):
Why coverage is so low, can you please add before code change stats as well

iamitprakash (changes requested):
where is working example proof ..?

Pratiyushkumar (approved):

controllers/users.js
```diff
@@ -128,6 +128,30 @@ const getUsers = async (req, res) => {
       }
     }
 
+    // getting user details by discord id if present.
+    const dev = req.query.dev === "true";
+    if (dev) {
+      if (req.query.discordId) {
+        const discordId = req.query.discordId;
```
iamitprakash, line 135: in the line const discordId = req.query.discordId;. The req.query object is a JavaScript object that contains the query parameters passed in the URL. In this case, the query parameter is discordId. However, the code is trying to access the discordId property of the req.query object before it has been initialized. This will cause an error.

middlewares/validators/user.js
```diff
@@ -158,6 +159,9 @@ async function getUsers(req, res, next) {
       id: joi.string().optional().messages({
         "string.empty": "id value must not be empty",
       }),
+      discordId: joi.string().optional().messages({
```
iamitprakash, line 162: messages object. The string.empty message is only triggered if the discordId value is empty. However, the optional() modifier means that the discordId value is allowed to be empty. This means that the string.empty message will never be triggered.

test/unit/models/users.test.js
```diff
@@ -117,6 +117,24 @@ describe("users", function () {
     });
   });
 
+  describe("fetch user details based on discord id", function () {
```
iamitprakash, line 120: here expect(result.user).to.haveOwnProperty("state");. The result.user object is the result of the users.fetchUser() function, which returns a user object. The state property is not part of the user object, so the expect() assertion will fail.

To fix this, you need to remove the expect(result.user).to.haveOwnProperty("state"); assertion.

models/users.js
```diff
@@ -259,6 +259,17 @@ const fetchUser = async ({ userId = null, username = null, githubUsername = null
         id = doc.id;
         userData = doc.data();
       });
+    } else if (discordId) {
+      const user = await userModel.where("discordId", "==", discordId).limit(1).get();
+      user.forEach((doc) => {
+        id = doc.id;
+        userData = doc.data();
+      });
+      const userStatusDocs = await userStatusModel.where("userId", "==", id).limit(1).get();
```
iamitprakash, line 268: here const userStatusDocs = await userStatusModel.where("userId", "==", id).limit(1).get();. The userStatusDocs variable is a list of documents. However, the userStatusDoc variable is only initialized with the first document in the list. This means that the data variable will only contain the data from the first document.

iamitprakash (changes requested):

controllers/users.js
```diff
@@ -128,6 +128,30 @@ const getUsers = async (req, res) => {
...
+
+        let result, user;
+        try {
+          result = await dataAccess.retrieveUsers({ discordId: discordId });
```
DashDeipayan, line 139: can be written as `result = await dataAccess.retrieveUsers({ discordId });`

controllers/users.js
```diff
@@ -128,6 +128,30 @@ const getUsers = async (req, res) => {
...
+          user = result.user;
+        } catch (error) {
+          logger.error(`Error while fetching user: ${error}`);
+          return res.boom.serverUnavailable(SOMETHING_WENT_WRONG);
```
DashDeipayan, line 143: I think an error of INTERNAL SERVER ERROR is a more meaningful error here

services/dataAccessLayer.js
```diff
@@ -47,6 +48,9 @@ const retrieveUsers = async ({
       users.push(user);
     });
     return { users, nextId, prevId };
+  } else if (discordId !== null) {
+    result = await userQuery.fetchUser({ discordId: discordId });
```
DashDeipayan, line 52: same as above

DashDeipayan (changes requested):

test/unit/models/users.test.js
j24m: image
I am adding the state property in the beforeEach block.

j24m: This is the updated user object after adding the state property in the beforeEach block. 
image

models/users.js
prakashchoudhary07, line 268: Why getting userStatus details in retrieve user by discordId

iamitprakash (approved):

DashDeipayan (dismissed):

iamitprakash (changes requested):
please change as per Prakash change request

services/dataAccessLayer.js
```diff
@@ -47,6 +48,9 @@ const retrieveUsers = async ({
       users.push(user);
     });
     return { users, nextId, prevId };
+  } else if (discordId !== null) {
+    result = await userQuery.fetchUser({ discordId });
+    return result;
```
prakashchoudhary07, line 53: Please filter the result according to the levels here
Do not directly return the result

prakashchoudhary07 (changes requested):

services/dataAccessLayer.js
```diff
@@ -12,8 +13,8 @@ const retrieveUsers = async ({
   role = null,
   userIds = null,
 }) => {
+  let result = {};
```
prakashchoudhary07, line 16: Why initialize result as an empty object?

iamitprakash:
@j24m  please add working proof

gauravsinhaweb (approved):

gauravsinhaweb (changes requested):
Add updated test result

Achintya-Chatterjee:
Please add test stats

controllers/users.js
```diff
@@ -130,6 +132,38 @@ const getUsers = async (req, res) => {
       }
     }
 
+    // getting user details by discord id if present.
+    const discordId = req.query.discordId;
+
+    const dev = req.query.dev === "true";
+    if (req.query.discordId) {
```
iamitprakash, line 139: can we merge both condition #139 and #140

controllers/users.js
```diff
@@ -130,6 +132,38 @@ const getUsers = async (req, res) => {
...
+      if (dev) {
```
iamitprakash, line 140: dev seems to sting into line 138, then why it's like bool

controllers/users.js
```diff
@@ -130,6 +132,38 @@ const getUsers = async (req, res) => {
...
+        let result, user;
+        try {
+          result = await dataAccess.retrieveUsers({ discordId });
+          user = result.user;
```
iamitprakash, line 144: what if result is null or empty

iamitprakash, line 144: please add test case to see user array is null

controllers/users.js
j24m: done

controllers/users.js
j24m: done

gauravsinhaweb (approved):

iamitprakash:
@j24m  
image

sahsisunny (approved):

iamitprakash (approved):

PR #1469, opened 2023-09-01
group-idle role on idle members

### LINKED ISSUE #1422: Add group-idle role
# What to do?
- [x] group-idle should be applied to the members who become idle
- [x] group-idle should be removed when a member is not idle

lakshayman:
closes #1422

### DIFF
--- a/models/discordactions.js
+++ b/models/discordactions.js
@@ -52,6 +52,33 @@ const getAllGroupRoles = async () => {
  * @param roleData { Object }: Data of the new role
  * @returns {Promise<discordRoleModel|Object>}
  */
+const getGroupRole = async (rolename) => {
+  try {
+    if (!rolename) return { roleExists: false };
+    const data = await discordRoleModel.where("rolename", "==", rolename).limit(1).get();
+    if (data.empty) {
+      return {
+        roleExists: false,
+      };
+    }
+    return {
+      roleExists: true,
+      role: {
+        id: data.docs[0].id,
+        ...data.docs[0].data(),
+      },
+    };
+  } catch (err) {
+    logger.error("Error in getting role", err);
+    throw err;
+  }
+};
+
+/**
+ *
+ * @param roleData { Object }: Data of the new role
+ * @returns {Promise<discordRoleModel|Object>}
+ */
 
 const isGroupRoleExists = async (rolename) => {
   try {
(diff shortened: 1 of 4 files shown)


models/discordactions.js
```diff
@@ -47,6 +47,33 @@ const getAllGroupRoles = async () => {
   }
 };
 
+/**
+ *
+ * @param roleData { Object }: Data of the new role
```
iamitprakash, line 52: roleData is never used

iamitprakash (changes requested):

models/userStatus.js
```diff
@@ -17,7 +17,65 @@ const { TASK_STATUS } = require("../constants/tasks");
 const userStatusModel = firestore.collection("usersStatus");
 const tasksModel = firestore.collection("tasks");
 const { userState } = require("../constants/userStatus");
+const { getGroupRole, addGroupRoleToMember, removeGroupRoleFromMember } = require("./discordactions");
 const usersCollection = firestore.collection("users");
+const DISCORD_BASE_URL = config.get("services.discordBot.baseUrl");
+const jwt = require("jsonwebtoken");
+
+const removeGroupRoleFromDiscordUser = async ({ userId, roleName }) => {
```
iamitprakash, line 25: roleid property of the dataForDiscord object is not being set correctly.

The removeGroupRoleFromDiscordUser function takes a userId and roleName parameter and removes the role with the given name from the user's Discord account. The function first calls the getGroupRole function to get the role data, and then calls the removeGroupRoleFromMember function to remove the role from the user's Discord guild. Finally, the function calls the fetch function to make a DELETE request to the Discord API to remove the role from the user's Discord account.

iamitprakash (changes requested):

models/userStatus.js
lakshayman: First, it removes the role data from our firestore database.
Yes, getGroupRole is for getting the ID of the role.
removeGroupRoleFromMember is for removing the role from our firestore database.
fetch call is for deleting it from the user on Discord.

models/discordactions.js
lakshayman: resolved in latest commit

iamitprakash (approved):

PR #1484, opened 2023-09-04
Remove Unverified Role

### LINKED ISSUE #1225: create a script for removing unverified role tag from them once they are verified

manish591:
### closes #1225

### DIFF
--- a/services/discordService.js
+++ b/services/discordService.js
@@ -86,10 +86,15 @@ const setUserDiscordNickname = async (userName, discordId) => {
   }
 };
 
+const removeUnverifiedRole = (discordId) => {
+  // Script here
+};
+
 module.exports = {
   getDiscordMembers,
   setInDiscordFalseScript,
   addRoleToUser,
   removeRoleFromUser,
   setUserDiscordNickname,
+  removeUnverifiedRole,
 };


services/discordService.js
```diff
@@ -85,10 +85,27 @@ const setUserDiscordNickname = async (userName, discordId) => {
   }
 };
 
+const markUserVerified = async (discordId) => {
```
Achintya-Chatterjee, line 88: can you please explain what this code does

controllers/external-accounts.js
```diff
@@ -87,10 +88,19 @@ const syncExternalAccountData = async (req, res) => {
...
         };
       }
+
+      const unverifiedRoleId = config.get("discordUnverifiedRoleId");
+      const isUserHasUnverifiedRole = discordUser && discordUser.roles.includes(unverifiedRoleId);
+
+      if (isUserHasUnverifiedRole) {
+        markUsersVerifiedPromises.push(markUserVerified(discordUser.user.id));
+      }
+
```
Achintya-Chatterjee, line 98: Can you please explain what the code does

services/discordService.js
```diff
@@ -85,10 +85,27 @@ const setUserDiscordNickname = async (userName, discordId) => {
...
+  try {
+    const unverifiedRoleId = config.get("discordUnverifiedRoleId");
+
+    if (!discordId) {
+      throw Error;
+    }
```
fakhruddinkw, line 94: Do we need to have this if check here?
I think, if the function accepts the discordId as parameter, and if we omit that parameter, we wont be able to call the function right?
And i think this if check is checking for whether discordId is there or not. 

I think we will never actually hit this condition the way this function is implemented, hence we do not need this here. Please mention if i have missed something.

services/discordService.js
```diff
@@ -85,10 +85,27 @@ const setUserDiscordNickname = async (userName, discordId) => {
   }
 };
 
+const markUserVerified = async (discordId) => {
+  try {
+    const unverifiedRoleId = config.get("discordUnverifiedRoleId");
+
```
fakhruddinkw, line 91: Suggestion: Instead of getting unverifiedRoleId at each function call, can we not just get the value of this once at the top of our file?

controllers/external-accounts.js
manish591: ` const isUserHasUnverifiedRole = discordUser && discordUser.roles.includes(unverifiedRoleId);`

For the first condition, we check if a user is in Discord and has run the /verify command. For the second condition, we check if the user after running the /verify command has an unverified role.

Then, if we find such users we call a script to remove the unverified role by passing their discord id in the script.

Thanks

gauravsinhaweb (changes requested):
could you add test cases for the 2nd case i.e `For old user already verified by running the verify command`

PR #1489, opened 2023-09-06
Fix Group Idle Sync Function

lakshayman:


models/userStatus.js
```diff
@@ -52,92 +51,59 @@ const getGroupRole = async (rolename) => {
...
-      }
-      const USER_DISCORD_ID = discordId || userData.discordId;
+    const groupRoleObj = await getGroupRole("group-idle");
+    if(groupRoleObj?.roleExists) {
+      const groupIdleRoleId = groupRoleObj.role.roleid;
+      const user = await usersCollection.doc(userId).get();
+      const discordId = user.data().discordId;
+      if(discordId) {
+        // Add role to firestore collection
+        const alreadyHasRole = await memberRoleModel
```
iamitprakash, line 92: if the user already has the groupIdleRoleId role, but the where clause is missing the array_contains operator. The correct code should be:

const alreadyHasRole = await memberRoleModel
  .where("roleid", "==", groupIdleRoleId)
  .where("userid", "==", discordId)
  .where("array_contains", "roles", groupIdleRoleId)
  .limit(1)
  .get();

iamitprakash (changes requested):

models/userStatus.js
lakshayman: The member-discord-roles collection has a document that has a single role like
{
  roleid: 122333,
  userid: 2321312
}

Document is for single role only there is no array in it.

iamitprakash (approved):

PR #1504, opened 2023-09-11
Update usernames to lowercase

manish591:
## Description

Fix the user's search not working on the status site.

## Problem

When searching for some users, the `users?search=username` API is not returning the correct details. This is because of Firestore case-sensitive queries. Please refer to the issue for more details.

## Solution

The solution is explained below steps:

We will write a script to make all the usernames lowercase and then use the search functionality defined before.

## Dev Tested

- [x] Yes

## Test Coverage

Controllers:

Screenshot (81)

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -762,6 +762,21 @@ async function usersPatchHandler(req, res) {
   }
 }
 
+const updateUsernamesToLowercase = async (req, res) => {
+  try {
+    const data = await userQuery.updateUsernameToLowercase();
+    return res.json({
+      message: "Updated usernames to lowercase successfully!",
+      data: {
+        totalUsersAffected: data,
+      },
+    });
+  } catch (error) {
+    logger.error(`Error while updating username to lowercase: ${error}`);
+    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
+  }
+};
+
 module.exports = {
   verifyUser,
   generateChaincode,
@@ -791,4 +806,5 @@ module.exports = {
   updateDiscordUserNickname,
   archiveUserIfNotInDiscord,
   usersPatchHandler,
+  updateUsernamesToLowercase,
 };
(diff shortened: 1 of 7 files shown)


controllers/users.js
```diff
@@ -762,6 +762,21 @@ async function usersPatchHandler(req, res) {
   }
 }
 
+const updateUsernamesToLowercase = async (req, res) => {
```
iamitprakash, line 765: the data variable returned from the updateUsernameToLowercase() function is not being checked for errors before it is passed to the res.json() function. This means that if the updateUsernameToLowercase() function fails, the data variable may contain an error message, which will be returned to the client.

models/users.js
```diff
@@ -829,6 +829,26 @@ const fetchUsersListForMultipleValues = async (documentKey, valueList) => {
   }
 };
 
+const updateUsernameToLowercase = async () => {
```
iamitprakash, line 832: the updateUserNameToLowercasePromises array may contain rejected promises if any of the addOrUpdate() functions fail. If this happens, the Promise.all() function will reject, and the return updateUserNameToLowercasePromises.length; statement will never be executed.

models/users.js
```diff
@@ -829,6 +829,26 @@ const fetchUsersListForMultipleValues = async (documentKey, valueList) => {
...
+  try {
+    const allUsers = await fetchAllUsers();
```
iamitprakash, line 834: the updateUsernameToLowercase() function is not returning a value. This means that the caller of the function will not know how many users were successfully updated.

services/users.js
```diff
@@ -41,6 +41,17 @@ const archiveUsers = async (usersData) => {
   }
 };
 
+const isUsernameLowercase = (username) => {
```
iamitprakash, line 44: this function will return false for usernames that contain non-ASCII characters, even if those characters are lowercase. For example, the function would return false for the username "bonjour", even though it is lowercase.

PR #1506, opened 2023-09-11
refact: added assignee filter in the overdue tasks api

ivinayakg:
## Brief
- Added a `assignee` filter in overdue tasks API
- this may need us to create additional indexes on the firstore

### Indexes needed on firestore
image


sahsisunny:
Completed in #1508

PR #1509, opened 2023-09-12
Fix/nickname sync api

kotesh-arya:
### Changes done:
- Added a try-catch block to handle the error case efficiently, where we try to update nicknames of users who exist on the database and do not exist in the Discord server.

### Test-stats for the extra changes:
API-refactor-test-stats


lakshayman (approved):
LGTM

iamitprakash (approved):

Achintya-Chatterjee (approved):

PR #1514, opened 2023-09-13
Fix/discord nickname sync api

kotesh-arya:
changes Done:
- Refactored Array.map to a for-loop
- Test case expect statements updated as per the response object keys

Before:

before-fixedAPI

After:

after-API-update

Successful API response:

after-API-JSON-response

### DIFF
--- a/controllers/discordactions.js
+++ b/controllers/discordactions.js
@@ -203,7 +203,7 @@ const updateDiscordNicknames = async (req, res) => {
 
     const discordServerUsers = await dataAccess.retrieveDiscordUsers();
     const nonSuperUsers = discordServerUsers.filter((user) => !user.roles.super_user);
-    let errorsArr = [];
+    const errorsArr = [];
     let successCounter = 0;
     let errorCounter = 0;
     const batchUpdate = nonSuperUsers.map(async (user) => {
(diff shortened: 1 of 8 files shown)


MayurLund (approved):

shubhamsigdar1 (approved):

Achintya-Chatterjee (approved):

controllers/discordactions.js
```diff
@@ -203,27 +202,33 @@ const updateDiscordNicknames = async (req, res) => {
...
-    const batchUpdate = nonSuperUsers.map(async (user) => {
-      const { discordId, username } = user;
+
+    let counter = 0;
+    for (let i = 0; i < nonSuperUsers.length; i++) {
+      const { discordId, username } = nonSuperUsers[i];
       try {
-        await setUserDiscordNickname(username, discordId);
+        if (counter % 10 === 0 && counter !== 0) {
+          await new Promise((resolve) => setTimeout(resolve, 4500));
```
prakashchoudhary07, line 215: Why is this?

yesyash (approved):

controllers/discordactions.js
```diff
@@ -203,27 +202,33 @@ const updateDiscordNicknames = async (req, res) => {
...
+        }
+        await setUserDiscordNickname(username.toLowerCase(), discordId);
+        counter++;
```
prakashchoudhary07, line 218: So if it allows 10 requests at once, then why not do 10 requests all at once and then wait for 3s and then next 10 together instead of waiting for each one?

Issue #1515, opened 2023-09-14
Fetch extension request logs of a given taskID for super user


Maheima:
## Issue description:

Summary:
The extension requests API 's GET method gives us all the extension requests and the logs API's GET method with type `extensionRequests` give us the logs of extension requests created, updated, timestamp, taskID etc. We need to get the extenion logs related to a particular taskID such that when we try to call GET method on `extensionRequests/logs` , we get logs of all the extension requests related user events of that task

Steps to do it:
Add a query param that gives the extension logs related to a particular taskID in the extension request model.

PR #1516, opened 2023-09-14
Get logs w.r.t the taskID

Maheima:
## Description
The extension requests API 's GET method gives us all the extension requests and the logs API's GET method with type extensionRequests give us the logs of extension requests created, updated, timestamp, taskID etc. We need to get the extenion logs related to a particular taskID such that when we try to call GET method on extensionRequests/logs , we get logs of all the extension requests related user events of that task

## Issue
https://github.com/Real-Dev-Squad/website-backend/issues/1515

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing

- [x] Is dev tested
- [ ] Is not dev tested

## Files modified

models/logs.js => Added query params to get the task with respect to a taskID

### DIFF
--- a/test/unit/models/logs.test.js
+++ b/test/unit/models/logs.test.js
@@ -1,19 +1,33 @@
 const chai = require("chai");
+
 const { expect } = chai;
 
 const cleanDb = require("../../utils/cleanDb");
+
 const logsQuery = require("../../../models/logs");
+
 const cacheData = require("../../fixtures/cloudflareCache/data");
+
 const logsData = require("../../fixtures/logs/archievedUsers");
+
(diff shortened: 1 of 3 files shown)


test/unit/models/logs.test.js
Pratiyushkumar, line 1: NIT - Also is it possible to check if meta key is not found then what will happen or in case of any error what can happen

PR #1530, opened 2023-09-17
FEATURE: Add more field overdue user response

sahsisunny:


controllers/users.js
```diff
@@ -167,18 +167,45 @@ const getUsers = async (req, res) => {
...
+          }
+        }
+
         return res.json({
-          message: "Users returned successfully!",
-          count: users.size,
-          users: Array.from(users),
+          message: "Users and their tasks returned successfully!",
+          count: usersData.length,
+          users: usersData,
```
vinit717, line 204: Here if the feature flag is not enabled then still it returns this message which is inappropriate as tasks are not there

controllers/users.js
```diff
@@ -167,18 +167,45 @@ const getUsers = async (req, res) => {
     if (transformedQuery?.filterBy === OVERDUE_TASKS) {
       try {
         const tasksData = await getOverdueTasks(days);
-        const users = new Set();
+        const userIds = new Set();
+        const usersData = [];
+
         tasksData.forEach((task) => {
-          users.add(task.assignee);
+          userIds.add(task.assignee);
```
iamitprakash, line 174: what if task doesn't have assignee prop

controllers/users.js
```diff
@@ -167,18 +167,45 @@ const getUsers = async (req, res) => {
     if (transformedQuery?.filterBy === OVERDUE_TASKS) {
       try {
         const tasksData = await getOverdueTasks(days);
```
iamitprakash, line 169: tasksData array to only include tasks that are assigned to the current user. However, the problem is that the tasksData array may contain tasks that are assigned to multiple users. This means that the userTasks array may contain duplicate tasks.

iamitprakash (changes requested):

controllers/users.js
sahsisunny: We are fetching those tasks to have an assignee.

controllers/users.js
```diff
@@ -167,18 +167,45 @@ const getUsers = async (req, res) => {
...
+              usersData.push(userData);
+            }
+          }
+        }
+
         return res.json({
-          message: "Users returned successfully!",
-          count: users.size,
-          users: Array.from(users),
+          message: "Users and their tasks returned successfully!",
```
sahsisunny, line 202:
```suggestion
          message: "Users returned successfully!",
```

test/integration/users.test.js
```diff
@@ -697,7 +697,7 @@ describe("Users", function () {
           }
           expect(res).to.have.status(200);
           expect(res.body).to.be.an("object");
-          expect(res.body.message).to.equal("Users returned successfully!");
+          expect(res.body.message).to.equal("Users and their tasks returned successfully!");
```
sahsisunny, line 700:
```suggestion
          expect(res.body.message).to.equal("Users returned successfully!");
```

controllers/users.js
sahsisunny: Done

vinit717 (approved):

iamitprakash (approved):

Pratiyushkumar (approved):
LGTM

yesyash (approved):

PR #1533, opened 2023-09-18
REFACTOR: filtering users to be effected as developers

kotesh-arya:
changes done:
- Filtering users with roles as `Mavens`, `super_users` and users whose nicknames which were already synced.
Before:
nickname-sync-api-v3-before
After:
nickname-sync-api-v3-after

API response after sync, when tested in Local:
nickname-sync-api-v-3-response


controllers/discordactions.js
```diff
@@ -200,35 +201,64 @@ const updateDiscordNicknames = async (req, res) => {
...
+            const isSuperuser = foundUserWithDiscordId.user.roles.super_user;
+            if (isDeveloper && !isMaven && !isUsernameMatched && !isBot && !isSuperuser) {
+              usersToBeEffected.push({
+                discordId: foundUserWithDiscordId.user.discordId,
+                username: foundUserWithDiscordId.user.username,
+                first_name: foundUserWithDiscordId.user.first_name,
+              });
+            }
+          }
+        } catch (error) {
```
iamitprakash, line 225: try/catch block. If an error occurs when fetching the user with the matching Discord ID, the code will throw a new error. However, this will cause the outer Promise.all() call to fail, even if all of the other user fetch operations were successful.

To fix this bug, you can simply remove the throw statement from the catch block. This will allow the outer Promise.all() call to succeed, even if some of the user fetch operations failed.

iamitprakash (changes requested):

controllers/discordactions.js
kotesh-arya: Fixed it, thanks for pointing it out 🙌🏻

controllers/discordactions.js
```diff
@@ -200,35 +201,63 @@ const updateDiscordNicknames = async (req, res) => {
...
-          await new Promise((resolve) => setTimeout(resolve, 4500));
+          await new Promise((resolve) => setTimeout(resolve, 3100));
+        }
+        if (!discordId) {
+          throw new Error("user not verified");
+        } else if (!username) {
+          throw new Error(`does not have a username`);
+        }
+        const response = await setUserDiscordNickname(username.toLowerCase(), discordId);
+        if (response.message) {
```
iamitprakash, line 247: response variable is of type Promise, so it may not have a message property yet. If the response.message property is not defined, the if statement will evaluate to false, and the code inside the block will not be executed. This means that the counter and totalNicknamesUpdated.count variables will not be incremented, even though the Discord nickname was successfully updated.

To fix this bug, you can use the following code:


if (response) {
  const message = await response.message;
  if (message) {
    counter++;
    totalNicknamesUpdated.count++;
  }
}

iamitprakash (changes requested):

iamitprakash (approved):

PR #1544, opened 2023-09-22
Feat: OOO nickname update

bharati-21:
# Feat: 1119 Update User Discord Nickname During OOO

### Closes: https://github.com/Real-Dev-Squad/website-backend/issues/1119

- The following PR contains changes related to the RDS discord server nickname change when a user is on OOO or out of OOO.
- The updates are reflected when the `POST /discord-actions/nickname/status` API is triggered by the cron job.
- The route takes `lastNicknameUpdate` timestamp from the cron job and uses it to fetch those status with latest update.

### Test Cases
- Please find the tests in this PR: https://github.com/Real-Dev-Squad/website-backend/pull/1545


controllers/discordactions.js
```diff
@@ -272,6 +272,27 @@ const updateDiscordNicknames = async (req, res) => {
...
 };
 
+/**
+ * Update all user Discord nickname based on status
+ *
+ * @param req {Object} - Express request object
+ * @param res {Object} - Express response object
+ */
+
+const updateUsersNicknameStatus = async (req, res) => {
```
iamitprakash, line 282: Please improve this code block on below points

1. Use a prepared statement: This will help to prevent SQL injection attacks.
2. Use a transaction: This will ensure that all of the changes are committed or rolled back together.
3. Use a batch update: This will improve performance if there are a large number of users to update.
4. Use a worker pool: This will improve performance if the code is being called concurrently by multiple threads.

middlewares/validators/discordactions.js
```diff
@@ -28,7 +28,22 @@ const validateMemberRoleBody = async (req, res, next) => {
   }
 };
 
+const validateUpdateUsersNicknameStatusBody = async (req, res, next) => {
+  const schema = Joi.object({
+    lastNicknameUpdate: Joi.alternatives().try(Joi.string().trim(), Joi.number().unit("milliseconds")).required(),
```
iamitprakash, line 33: lastNicknameUpdate field is allowed to be either a string or a number, but the unit("milliseconds") validator only works with numbers. This means that if the lastNicknameUpdate field is a string, the validation will fail.

middlewares/validators/discordactions.js
```diff
@@ -28,7 +28,22 @@ const validateMemberRoleBody = async (req, res, next) => {
   }
 };
 
+const validateUpdateUsersNicknameStatusBody = async (req, res, next) => {
```
iamitprakash, line 31: improve this code block on below points

1. Use a cache: You can cache the schema object to avoid having to recompile it on every request.
2. Use a middleware: You can create a middleware function that validates the request body using the schema object. This will allow you to reuse the validation code in multiple routes.
3. Use asynchronous validation: You can use the validateAsync() method to validate the request body asynchronously. This will improve the performance of your application.

models/discordactions.js
```diff
@@ -336,6 +339,90 @@ const updateIdleUsersOnDiscord = async () => {
...
+      nicknameUpdateBatches.push(usersStatusDocs.slice(startIndex, end));
+      startIndex = end;
+    }
+
+    for (let i = 0; i < nicknameUpdateBatches.length; i++) {
+      const promises = [];
+      const usersStatusDocsBatch = nicknameUpdateBatches[i];
+      usersStatusDocsBatch.forEach((document) => {
+        const doc = document.data();
+        const userId = doc.userId;
```
iamitprakash, line 383: userId variable is assigned to a value from the doc object, which is untrusted. This means that an attacker could control the value of the userId variable by injecting malicious data into the doc object.

models/discordactions.js
```diff
@@ -336,6 +339,90 @@ const updateIdleUsersOnDiscord = async () => {
   };
 };
 
+const updateUsersNicknameStatus = async (lastNicknameUpdate) => {
```
iamitprakash, line 342: improve/ refactor this function on below points 
1. Use a cache: You can cache the results of the usersCurrentStatus and usersFutureStatus queries to avoid having to run them every time the function is called.
2. Use a batch update: You can use a batch update to update the nicknames of multiple users at once. This will improve the performance of the function if there are a large number of users to update.
3. Use a worker pool: You can use a worker pool to update the nicknames of multiple users in parallel. This will further improve the performance of the function, especially if there are a large number of users to update.

utils/users.js
```diff
@@ -218,6 +221,56 @@ const parseSearchQuery = (queryString) => {
   return searchParams;
 };
 
+/**
+ * Generates discord nickname for a user
+ *
+ * @param {string} username - The discord username of the user.
+ * @returns {string} - Nickname of the user.
+ */
+const generateOOONickname = (username = "", from, until) => {
```
iamitprakash, line 230: here  the generateOOONickname() function. The function uses the substring() method to truncate the username to fit within the Discord nickname length limit. However, the substring() method does not take into account the length of the OOO message. This can result in the username being truncated too short, and the OOO message being truncated or omitted entirely.

iamitprakash, line 230: improvement points 

1. Use a cache: You can cache the results of the dataAccessLayer.retrieveUsers() call to avoid having to call it every time the function is called.
2. Use a batch update: If you are updating the nicknames of multiple users at once, you can 3. use a batch update to improve performance.
4. Use a worker pool: If you are updating the nicknames of a large number of users, you can use a worker pool to improve performance by updating the nicknames of multiple users in parallel.

iamitprakash (changes requested):

controllers/discordactions.js
bharati-21: 1. I am not sure I understand this. I thought the attacks occur only when SQL commands are built and instead of user input malicious SQL statements are provided. Can you please help me with this? Because when I searched on this in the context of firestore, I was unable to find anything about injection attacks. Also could not find ways to add prepared statements like in an SQL database.
2. Sure, I have made the changes to ensure that either all the nicknames are updated together or none of them are.
3. In the `models/discordactions.js`, I am updating the nickname in the batches of 4 Ensuring that there's a delay between each batch's nickname update.
4. For this I would require time to read and understand about worker pool. Please tell me if it's okay if I can take this as a separate extension and enhancement of this task.

middlewares/validators/discordactions.js
bharati-21: Yes, thank you for pointing it out. There's no need to use `string()` validation. The validation to check if the numbers are sent properly as ms will be handled automatically by `number().unit()`. Even if they are strings, using `number()`, the validation is applied after trying to parse the value to number.

middlewares/validators/discordactions.js
bharati-21: 1. Are you referring to caching the nicknames of the users that are being recompiled? Or caching the result of the request? 
2. I have split the validation code and the validator. Please confirm if this is what was expected.
3. I did use `validateAsync()`. Am I missing something here?

models/discordactions.js
bharati-21: Are you implying that it is more secure when userId is directly retrieved as `document.data().userId`?

utils/users.js
bharati-21: Yes, this was discussed with before with Ankush during discovery. It was decided that when a user is in OOO, the OOO dates should be given priority even if that results in the username being temporarily shortened. Hence the username (and thus it's length) in the nickname is decided on the basis of the length of OOO message. 
```
 const nicknameLen = discordNicknameLength - oooMessage.length - 1;
 return `${username.substring(0, nicknameLen)} ${oooMessage}`;
  ```

models/discordactions.js
bharati-21: For 1 & 3, Again, as I've asked above: for I would require time to read and understand about worker pool. Please tell me if it's okay if I can take this as a separate extension and enhancement of this task.

2. This function is already being called separately in batches of 4 for users.

utils/users.js
bharati-21: For 1 & 3, Again, as I've asked above: for I would require time to read and understand about worker pool. Please tell me if it's okay if I can take this as a separate extension and enhancement of this task.

2. This function is already being called separately in batches of 4 for users.

shubhamsigdar1 (approved):
good work @bharati-21, not find anything in code to comment

iamitprakash (approved):

Achintya-Chatterjee (dismissed):

iamitprakash (approved):

PR #1555, opened 2023-09-25
Refactor to Retrieve Users in Discord only

sahsisunny:
**Description:**
To improve our user management, we need to refactor the code to return only those users who are in our Discord server .


shreya-mishra (approved):
Please add description !

j24m (approved):

PR #1558, opened 2023-09-25
sync discord groups in firestore

### LINKED ISSUE #1494: Sync discord groups and groups saved in firestore.
This API will sync the groups in Firestore and discord with ID and if the group is manually deleted from the discord then upon syncing it will delete that group from firestore as well.

MehulKChaudhari:
**Description:**

This pull request introduces a new feature: Sync discord group in Firestore. This API will check if any group `id` is changed it will update it and if there is any new role which is not stored in the Firestore will add that role. 

**Changes Made:**

* Added a new discordActions controller function, `syncDiscordGroupRolesInFirestore`, which gets all the discord roles in Discord.
* Then checks if any role with same name exists in Firestore `discord-roles` collection but its ID is not updated then it updates the ID
* But there are any roles which are not saved in the Firestore then it creates documents of those roles in Firestore collection. 
* it gives all the roles saved in Firestore in the response.
* Added detailed JSDoc documentation for the `updateGroupRole` and `getGroupRoleByName` functions. 
image

Testing: 
image

### DIFF
--- a/controllers/discordactions.js
+++ b/controllers/discordactions.js
@@ -3,10 +3,12 @@ const admin = require("firebase-admin");
 const config = require("config");
 const jwt = require("jsonwebtoken");
 const discordRolesModel = require("../models/discordactions");
-const { setUserDiscordNickname, getDiscordMembers, getDiscordRoles } = require("../services/discordService");
+const discordServices = require("../services/discordService");
 const { fetchAllUsers } = require("../models/users");
 const discordDeveloperRoleId = config.get("discordDeveloperRoleId");
 const discordMavenRoleId = config.get("discordMavenRoleId");
+
+const { setUserDiscordNickname, getDiscordMembers } = discordServices;
 
 /**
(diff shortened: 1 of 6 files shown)


iamitprakash (changes requested):
1. Add test case stats
2. Working proof

controllers/discordactions.js
```diff
@@ -270,6 +273,47 @@ const updateDiscordNicknames = async (req, res) => {
   }
 };
 
+const syncDiscordGroupRolesInFirestore = async (req, res) => {
+  try {
+    const value = await discordServices.getDiscordRoles();
```
prakashchoudhary07, line 278: NIT: roles or discordRoles

prakashchoudhary07 (approved):

Achintya-Chatterjee (dismissed):

iamitprakash (approved):

Issue #1503, opened 2023-09-10
Unified Token System [RFC] - Sign In with QR Code Enhancement


sahsisunny:
## Issue Description

**Problem:**
Our mobile app offers two distinct sign-in methods: "Sign In with QR Code" and "Sign In with GitHub." There is a discrepancy in how tokens are handled for these methods. "Sign In with QR Code" does not store any token for API authentication, while "Sign In with GitHub" stores a token. This separation of token handling has led to potential API failures and a lack of consistency in our authentication system.

**Desired Solution:**
We need to create a unified token system that can be used for both "Sign In with QR Code" and "Sign In with GitHub." This will streamline our authentication process, reduce the risk of API failures, and enhance the user experience by providing a consistent sign-in mechanism.

**Proposed Changes for "Sign In with QR Code":**
1. Develop a single token generation and management system that can be used for both sign-in methods.
2. Update the authentication flow for "Sign In with QR Code" to utilize the unified token.

**Expected Benefits:**
1. Reduced risk of API failures due to a consistent token system.
2. Improved user experience through a unified sign-in process.
3. Easier maintenance and scalability of the authentication system.

## Proposed Solution for "Sign In with QR Code"

To address the issue of having separate token handling for "Sign In with QR Code" and "Sign In with GitHub" and create a unified token system for "Sign In with QR Code," we can follow the proposed solution outlined below:

#### a. Token Generation
Utilize an existing function on the server-side that generates tokens upon successful authentication for "Sign In with QR Code."

#### b. Token Storage
- For "Sign In with QR Code," introduce token storage.
- When a user successfully signs in with a QR code, generate a unified token as described above and store it securely on the mobile device.
- This token will be used for subsequent API requests.

#### c. Token Validation
- Implement a token validation mechanism on the server-side to verify the authenticity and validity of the tokens during API requests.
- This validation should check the token's signature, expiration time, and any relevant user information associated with it.

#### a. "Sign In with QR Code"
1. When a user successfully signs in with a QR code, send the user's information (e.g., user ID) to the server.
2. The server generates a unified token and associates it with the user's account.
3. Send the unified token back to the mobile app, which stores it securely.
4. Use the unified token for subsequent API requests.

PR #1561, opened 2023-09-26
REFACTOR: return rds-session token in response

kotesh-arya:
#### issue addressed: https://github.com/Real-Dev-Squad/website-backend/issues/1503
This PR addresses the **Token Generation** part of the above mentioned issue as follows:
Initially a unique token was being sent when user authenticate themselves using **Scan with QR**, but while hitting the API endpoints we do need the session cookie of the user to Authorize each request.
### Changes done:
Instead of sending a unique token, rds-session cookie is being sent as a part of the response, which can be used to make successful API calls.
 
rds-session-token-respose_censored

Test coverage stats:
image

image

### DIFF
--- a/utils/generateUniqueToken.js
+++ b/utils/generateUniqueToken.js
@@ -1,6 +1,0 @@
-import { generateAuthToken } from "../services/authService";
-
-export const generateUniqueToken = (userId) => {
-  const token = generateAuthToken({ userId });
-  return token;
-};
--- a/utils/generateUniqueToken.js
+++ b/utils/generateUniqueToken.js
@@ -1,6 +1,6 @@
 import { generateAuthToken } from "../services/authService";
 
-export const generateUniqueToken = async (userId) => {
+export const generateUniqueToken = (userId) => {
   const token = generateAuthToken({ userId });
   return token;
 };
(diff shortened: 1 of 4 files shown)


shreya-mishra (commented):
Test ?

kotesh-arya:
> Test ?

Added test 👍🏻

Ajeyakrishna-k (dismissed):

utils/generateUniqueToken.js
```diff
@@ -1,12 +1,6 @@
...
-export const generateUniqueToken = async () => {
-  const uuidToken = crypto.randomUUID();
-  const randomNumber = Math.floor(Math.random() * 1000000);
-  const generationTime = Date.now();
-  const encoder = new TextEncoder();
-  const encodedString = encoder.encode(uuidToken + randomNumber + generationTime);
-  const hash = await crypto.subtle.digest("SHA-256", encodedString);
-  const token = [...new Uint8Array(hash)].map((x) => x.toString(16).padStart(2, "0")).join("");
+export const generateUniqueToken = (userId) => {
+  const token = generateAuthToken({ userId });
```
Ajeyakrishna-k, line 4: Is this function required? As this just calls and returns `generateAuthToken`

sahsisunny: I don't think

sahsisunny (approved):

Ajeyakrishna-k (approved):

Issue #1418, opened 2023-08-15
Create multiple Extension Requests


MayurLund:
Currently users are restricted to a single extension request, preventing them from creating additional requests as the button gets disabled after 1 extension request.

MayurLund:
**Here is the workflow of it** :
![ExtensionRequestWorkflow-Page-3 (1)](https://github.com/Real-Dev-Squad/website-backend/assets/37043221/3a114a72-93e6-4934-9ed6-02097e643ce7)

![ExtensionRequestWorkflow drawio (1)_pages-to-jpg-0003](https://github.com/Real-Dev-Squad/website-backend/assets/37043221/500018bd-3503-4cf3-9b4d-3fb508f38652)

iamitprakash:
@MayurLund  what param you are passing to GET API...

MayurLund:
> @MayurLund what param you are passing to GET API...

Hi @iamitprakash I am not changing query params. Currently its using 3 params -> status, taskId, assignee and I am just ordering it by timestamp so as to get the latest one

iamitprakash:
> > @MayurLund what param you are passing to GET API...
> 
> Hi @iamitprakash I am not changing query params. Currently its using 3 params -> status, taskId, assignee and while creating new one I am just using ->taskId,assignee as query params and ordering by timestamp so as to get the latest one

@MayurLund that need to mention in above flow-chart

MayurLund:
> > > @MayurLund what param you are passing to GET API...
> > 
> > 
> > Hi @iamitprakash I am not changing query params. Currently its using 3 params -> status, taskId, assignee and while creating new one I am just using ->taskId,assignee as query params and ordering by timestamp so as to get the latest one
> 
> @MayurLund that need to mention in above flow-chart

Thanks @iamitprakash for pointing it out. Will update and add it.

PR #1563, opened 2023-09-26
Allow multiple extension request

MayurLund:
Issue Ticket Number:-
- #1418 

Backend changes
- [x] Yes
- [ ] No

Frontend Changes
- [x] Yes - On mysite and Dashboard
- [ ] No

Database changes
- [x] Yes - Addition of the request number field in extension request data.
- [ ] No

### Deployment notes
Creating an index is necessary before merging it into the main branch.

### Merging these frontend pull requests simultaneously is crucial for completing this feature:
Dashboard PR : https://github.com/Real-Dev-Squad/website-dashboard/pull/519
Mysite PR : https://github.com/Real-Dev-Squad/website-my/pull/496

### Workflow
Previously, the API did not allow the creation of more than one request at a time. To address this limitation, modifications were made under the 'dev' flag. A new function, `fetchLatestExtensionRequest`, was introduced. When triggered, this function retrieves the latest extension request and checks its status. If the status is either 'Approved' or 'Denied,' it allows users to create another extension request.

Additionally, a new field, `requestNumber`, was introduced. This field increments each time a user submits a new request. This increment is reflected both on the user's end within my site and on the dashboard site for super users.

Here is the demo video of it:

[screen-capture (10).webm](https://github.com/Real-Dev-Squad/website-backend/assets/37043221/ce801f71-b826-4f23-be72-b6c8c45c7789)

Here is the workflow of it :
ExtensionRequest

**Changes required in backend :**
1.Creation of index so as to get the latest extension by descending timestamp when multiple extensions are present for a particular user and task.
2.Addition of requestNumber field and its increment and its handling (since its not present and undefined as per current DB fields).

### Test Stats :
image
image


iamitprakash (changes requested):
convert all date into UTC while saving to database

MayurLund:
> convert all date into UTC while saving to database

Hi @iamitprakash Thank you for raising this concern. I understand that there's an issue with the storage of dates .If I just convert extension dates to UTC, it might cause problems with the "taskEndsOn" field and discrepancies on other pages.

shubhamsigdar1 (dismissed):
Good work @MayurLund and Thank you for video and test cases

test/integration/multilpleExtensionRequest.js
```diff
@@ -0,0 +1,520 @@
...
+        lossRate: { [DINERO]: 1 },
+      },
+      {
+        title: "Test task",
+        type: "feature",
+        endsOn: 1234,
+        startedOn: 4567,
+        status: "active",
+        percentCompleted: 10,
+        participants: [],
```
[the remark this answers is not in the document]
MayurLund: Hi @shubhamsigdar1 This field comes up when you select a group on create task page while creating task from dashboard site but I dont think so currently its functional.
image

test/integration/multilpleExtensionRequest.js
```diff
@@ -0,0 +1,520 @@
...
+const cookieName = config.get("userToken.cookieName");
+const userData = require("../fixtures/user/user")();
+const { DINERO, NEELAM } = require("../../constants/wallets");
+const cleanDb = require("../utils/cleanDb");
+const { EXTENSION_REQUEST_STATUS } = require("../../constants/extensionRequests");
+
+chai.use(chaiHttp);
+
+const user = userData[6];
+const appOwner = userData[3];
```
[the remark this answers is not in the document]
MayurLund: Yes, task extension can only be created by the user who has been assigned that task or SUPERUSER

MayurLund:
> Good work @MayurLund and Thank you for video and test cases
Thanks @shubhamsigdar1!

MayurLund:
> convert all date into UTC while saving to database
Hi @iamitprakash its converting into UTC only. I verified it.

controllers/extensionRequests.js
```diff
@@ -132,22 +225,45 @@ const getExtensionRequest = async (req, res) => {
...
+      const { taskId } = req.query;
+
+      if (userId) {
+        let allExtensionRequests = await extensionRequestsQuery.fetchLatestExtensionRequest({
+          taskId,
+        });
+
+        if (allExtensionRequests.length && allExtensionRequests[0].userId !== userId) {
+          allExtensionRequests = [];
+        }
```
prakashchoudhary07, line 242: When does allextensionRequest[0].userId !== userId becomes true

prakashchoudhary07:
NIT: Please break controller in small code, and move that to service or utils

utils/extensionRequests.js
```diff
@@ -16,12 +16,12 @@ const formatExtensionRequest = async (extensionRequest) => {
...
 
   let { timestamp, id, assignee, ...body } = extensionRequest;
-
+  const userId = assignee;
   if (assignee) {
     assignee = await getUsername(assignee);
   }
 
-  return { ...body, id, timestamp, assignee };
+  return { ...body, id, timestamp, assignee, userId };
```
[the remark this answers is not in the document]
MayurLund: No if you see at 21st line the username is assigned but in future username can change so included the userId as well.

controllers/extensionRequests.js
```diff
@@ -14,73 +14,166 @@ const { parseQueryParams } = require("../utils/queryParser");
...
+
+      // logic for extension request count
+      let requestNumber;
+      if (latestExtensionRequest[0] && latestExtensionRequest[0].userId === assigneeId) {
+        if (latestExtensionRequest[0].requestNumber !== undefined && latestExtensionRequest[0].requestNumber > 0) {
+          requestNumber = latestExtensionRequest[0].requestNumber + 1;
+          extensionBody = { ...extensionBody, requestNumber };
+        } else {
+          extensionBody = { ...extensionBody, requestNumber: 2 };
+        }
```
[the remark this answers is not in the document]
MayurLund: Latest DB doesn't contain field requestNumber so there are two ways to handle that either run a script to include requestNumber field in all existing records and the other way is to if requestNumber doesn't exist in previousRecord directly increase the requestNumber 2 so I have followed this approach.

controllers/extensionRequests.js
MayurLund: When assignee changes then Latest extension request will have the userId of previous assignee and when it doesn't matches I need to create [] which will allow at front end to create the extension request

test/integration/multilpleExtensionRequest.js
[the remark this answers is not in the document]
shubhamsigdar1: Okk Thank you

controllers/extensionRequests.js
```diff
@@ -14,73 +14,166 @@ const { parseQueryParams } = require("../utils/queryParser");
...
+      }
+      if (extensionBody.oldEndsOn !== task.endsOn) {
+        extensionBody.oldEndsOn = task.endsOn;
+      }
+
+      const latestExtensionRequest = await extensionRequestsQuery.fetchLatestExtensionRequest({
+        taskId: extensionBody.taskId,
+      });
+
+      if (latestExtensionRequest[0] && latestExtensionRequest[0].status === "PENDING") {
```
Ajeyakrishna-k, line 56: Please use constants here.

controllers/extensionRequests.js
```diff
@@ -14,73 +14,166 @@ const { parseQueryParams } = require("../utils/queryParser");
...
+        return res.boom.badRequest("An extension request for this task already exists.");
+      }
+
+      // dont allow newETA<taskEndsOnETA
+      if (latestExtensionRequest[0]) {
+        if (latestExtensionRequest[0].status === "APPROVED" || latestExtensionRequest[0].status === "DENIED") {
```
Ajeyakrishna-k, line 62: This if-block will always be true as status can only be [PENDING, APPROVED, DENIED] and there's already a if-block before this which returns bad request when the status is PENDING.

controllers/extensionRequests.js
```diff
@@ -14,73 +14,166 @@ const { parseQueryParams } = require("../utils/queryParser");
...
+          if (extensionBody.newEndsOn < task.endsOn) {
```
Ajeyakrishna-k, line 63: There's a same check at line [45](https://github.com/Real-Dev-Squad/website-backend/pull/1563/files#diff-7fda274defc6e98dbba9ad405daf25c6db794428a333e30dc880bdfaa75ceb00R45), and it returns bad request. So this if-block will never be executed.

I think all three nested if-blocks can be removed here.

controllers/extensionRequests.js
```diff
@@ -14,73 +14,166 @@ const { parseQueryParams } = require("../utils/queryParser");
...
+            return res.boom.badRequest("An extension request can't be equal to or lesser than taskEndsOn date");
+          }
+        }
+      }
+
+      // logic for extension request count
+      let requestNumber;
+      if (latestExtensionRequest[0] && latestExtensionRequest[0].userId === assigneeId) {
+        if (latestExtensionRequest[0].requestNumber !== undefined && latestExtensionRequest[0].requestNumber > 0) {
```
Ajeyakrishna-k, line 72:
```suggestion
        if (latestExtensionRequest[0].requestNumber && latestExtensionRequest[0].requestNumber > 0) {
```

controllers/extensionRequests.js
```diff
@@ -14,73 +14,166 @@ const { parseQueryParams } = require("../utils/queryParser");
...
+        if (latestExtensionRequest[0].status === "APPROVED" || latestExtensionRequest[0].status === "DENIED") {
+          if (extensionBody.newEndsOn < task.endsOn) {
+            return res.boom.badRequest("An extension request can't be equal to or lesser than taskEndsOn date");
+          }
+        }
+      }
+
+      // logic for extension request count
+      let requestNumber;
+      if (latestExtensionRequest[0] && latestExtensionRequest[0].userId === assigneeId) {
```
Ajeyakrishna-k, line 71: - Suppose user A creates an ER for task 1. 
- task 1 gets assigned to user B
- user B raises an ER for task 1.
- task 1 gets re-assigned to user A.
- if user A raises ER again wouldn't the er request number get reset to 1 with this implementation?
- Shouldn't it be 2?

To solve this I think we should fetch the `latestExtensionRequest` based on assigneeId and taskId.

controllers/extensionRequests.js
```diff
@@ -132,22 +225,45 @@ const getExtensionRequest = async (req, res) => {
...
-    const { taskId, status } = req.query;
-
-    if (userId) {
-      const allExtensionRequests = await extensionRequestsQuery.fetchExtensionRequests({
-        status,
-        taskId,
-        assignee: userId,
-      });
-      return res.json({ message: "Extension Requests returned successfully!", allExtensionRequests });
+  // Inside dev flag it just shows
```
Ajeyakrishna-k, line 228:
```suggestion
```

models/extensionRequests.js
```diff
@@ -143,10 +143,34 @@ const fetchExtensionRequest = async (extensionRequestId) => {
...
+    });
+    const extensionRequestSnapshot = await extensionRequestsSnapshot.orderBy("timestamp", "desc").limit(1).get();
+
+    if (extensionRequestSnapshot.size === 0) {
+      return [];
+    }
+
+    const request = buildExtensionRequests(extensionRequestSnapshot);
+    const updatedRequests = await formatExtensionRequest(request[0]);
+    return [updatedRequests];
```
Ajeyakrishna-k, line 162: Why are we returning this as an array?

utils/extensionRequests.js
```diff
@@ -16,12 +16,12 @@ const formatExtensionRequest = async (extensionRequest) => {
   if (!extensionRequest) return extensionRequest;
 
   let { timestamp, id, assignee, ...body } = extensionRequest;
-
+  const userId = assignee;
```
Ajeyakrishna-k, line 19: Wouldn't `assigneeId` or `assigneeUserId` make more sense here?

controllers/extensionRequests.js
Ajeyakrishna-k, line 242: Same case as mentioned [here](https://github.com/Real-Dev-Squad/website-backend/pull/1563/files#r1345925185). 

In this case wouldn't the API return an empty array? even though there are older extension request.

controllers/extensionRequests.js
MayurLund: I have changed it now.Thanks!

controllers/extensionRequests.js
MayurLund: Nice catch! I have removed that block of code. Thanks!

controllers/extensionRequests.js
MayurLund: Made the changes.

controllers/extensionRequests.js
MayurLund: I have made the change. Thanks!

controllers/extensionRequests.js
MayurLund: No if assignee is different than previous and if it goes to same user I am making it ER 1.

controllers/extensionRequests.js
```diff
@@ -14,73 +14,166 @@ const { parseQueryParams } = require("../utils/queryParser");
...
+      // dont allow newETA<taskEndsOnETA
+      if (latestExtensionRequest[0]) {
+        if (latestExtensionRequest[0].status === "APPROVED" || latestExtensionRequest[0].status === "DENIED") {
+          if (extensionBody.newEndsOn < task.endsOn) {
+            return res.boom.badRequest("An extension request can't be equal to or lesser than taskEndsOn date");
+          }
+        }
+      }
+
+      // logic for extension request count
```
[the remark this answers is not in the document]
MayurLund: Removed that line.

controllers/extensionRequests.js
MayurLund: Removed that line.

utils/extensionRequests.js
MayurLund: Thanks for the suggestion. I have made the change and it increases the readability of code now.

test/integration/multilpleExtensionRequest.test.js
```diff
@@ -0,0 +1,428 @@
...
+        endsOn: 1234,
+        startedOn: 4567,
+        status: "active",
+        percentCompleted: 10,
+        assignee: appOwner.username,
+        isNoteworthy: true,
+        completionAward: { [DINERO]: 3, [NEELAM]: 300 },
+        lossRate: { [DINERO]: 1 },
+      },
+    ];
```
vinit717, line 86: Please move this on fixtures

MayurLund: I have moved it to fixtures now.

shreya-mishra (approved):

shubhamsigdar1 (dismissed):

iamitprakash (changes requested):
please update SS or Delete this if not needed

controllers/extensionRequests.js
```diff
@@ -14,73 +14,156 @@ const { parseQueryParams } = require("../utils/queryParser");
...
-    logger.error(`Error while creating new extension request: ${err}`);
-    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
+      const extensionLog = {
+        type: "extensionRequests",
+        meta: {
+          taskId: extensionBody.taskId,
+          createdBy: req.userData.id,
+        },
+        body: {
+          extensionRequestId: extensionRequest.id,
```
iamitprakash, line 149: extensionBody object may not contain the id property yet, since it is only assigned after the extensionRequest object is created. This means that if the extensionBody object does not contain the id property, the ...extensionBody spread operator will copy all of the properties from the extensionBody object to the extensionRequest object, except for the id property. This will result in the extensionRequest object being missing the id property.

iamitprakash (changes requested):
Please change error message as discussed

controllers/extensionRequests.js
MayurLund: Since the await is used it would wait for the extensionRequest to create hence it would contain the id property. So null wont be there and  extension request will always contain the id field and all other details.

iamitprakash (approved):

Issue #1569, opened 2023-09-27
Sorting the progress update API modifications


skv93-coder:
@skv93-coder 

### Related issue(s):
https://github.com/Real-Dev-Squad/website-status/issues/876

for overview and context

Implementing the second solution of above problem.

### Goals:-
After this by default latest progress will come first 
By passing orderBy=-date the oldest progress will be returned first

### Not the goal:-
The task will Not involve any frontend integration for now as no of progress are not many.

### Milestone
1. Write test for sorting the progress updates by date.
2. Add a query param orderby=date which i will use to order progress.

PR #1578, opened 2023-10-01
Added query parameter orderBy in progress API

skv93-coder:
Date:02-10-23
Developer Name: Shubham Sharma

PR Number(s):-
Issue Ticket Number:- https://github.com/Real-Dev-Squad/website-backend/issues/1569

Backend changes
Yes

Frontend Changes
No

Database changes
No

No

Deployment notes
Any special instructions for the deployment
No

Tested in staging
No

Also attach their screenshots.

### Images/video of the change:
image

Screenshot from 2023-10-02 12-16-23

image

### Follow-up Issues (if any)
No

### DIFF
--- a/utils/progresses.js
+++ b/utils/progresses.js
@@ -95,13 +95,22 @@ const assertUserOrTaskExists = async (queryParams) => {
  * @returns {Query} A Firestore query object that filters progress documents based on the given parameters.
  */
 const buildQueryToFetchDocs = (queryParams) => {
-  const { type, userId, taskId } = queryParams;
+  const { type, userId, taskId, orderBy } = queryParams;
+  const orderByField = !orderBy ? "date" : orderBy.split("-")[orderBy.split("-").length - 1];
+  const isAscOrDsc = orderBy && orderBy.split("-").length === 1 ? "asc" : "desc";
+
   if (type) {
-    return progressesCollection.where("type", "==", type);
+    return progressesCollection.where("type", "==", type).orderBy(orderByField, isAscOrDsc);
   } else if (userId) {
-    return progressesCollection.where("type", "==", "user").where("userId", "==", userId);
+    return progressesCollection
+      .where("type", "==", "user")
+      .where("userId", "==", userId)
+      .orderBy(orderByField, isAscOrDsc);
   } else {
-    return progressesCollection.where("type", "==", "task").where("taskId", "==", taskId);
+    return progressesCollection
+      .where("type", "==", "task")
+      .where("taskId", "==", taskId)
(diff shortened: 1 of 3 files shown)


iamitprakash (changes requested):
test case stats, missing

utils/progresses.js
```diff
@@ -92,16 +92,26 @@ const assertUserOrTaskExists = async (queryParams) => {
  * @param {string} queryParams.userId - (Optional) The user ID to filter progress documents by.
  * @param {string} queryParams.taskId - (Optional) The task ID to filter progress documents by.
  * @param {string} queryParams.type - (Optional) The type to filter progress documents by.
+ * @param {string} queryParams.orderBy - (Optional) The type to sort the documents.
  * @returns {Query} A Firestore query object that filters progress documents based on the given parameters.
  */
 const buildQueryToFetchDocs = (queryParams) => {
-  const { type, userId, taskId } = queryParams;
+  const { type, userId, taskId, orderBy } = queryParams;
+  const orderByField = !orderBy ? "date" : orderBy.split("-")[orderBy.split("-").length - 1];
```
iamitprakash, line 100: here const orderByField = !orderBy ? "date" : orderBy.split("-")[orderBy.split("-").length - 1];

This line is supposed to extract the field name from the orderBy query parameter, which is in the format field-asc/desc. However, if the orderBy parameter is not in the correct format, or if it is not provided at all, the split() function will return an empty array. In this case, the orderByField variable will be set to undefined.

This will cause an error when the query is executed, because Firestore cannot order documents by a field that does not exist.

can we think link something below ...?

const orderByField = orderBy ? orderBy.split("-")[0] : "date";

iamitprakash:
@skv93-coder  

here is PR format for you

All the PRs must be verified on staging before syncing dev to main or releasing to prod.

Please follow the below template when you raise a dev to main PR

Date:
Developer Name:

PR Number(s):-
Issue Ticket Number:-

Backend changes
Yes
No

Frontend Changes
Yes
No

Database changes
Yes
No

Yes
No

Deployment notes
Any special instructions for the deployment

Tested in staging
Yes
No

List of PRs going in this Sync.

Also attach their screenshots.

utils/progresses.js
skv93-coder: The condition !orderBy will make sure that split will only execute for truthy i.e for string of length more than 0.

Also if a field which does not exist in any of the document is provided the firestore returns zero result not an error.

Through, the suggested change is more readable. so i have applied that

iamitprakash (approved):

Issue #1579, opened 2023-10-02
Fix: From and Until Timestamps when OOO for One Day


bharati-21:
## Problem Statement
- As of now when a user selects OOO from `website-my`, the timestamp is stored as the `00:00:00 UTC` timestamp for the chosen date.
- Problem arises when the user chooses to be on OOO for one day. Suppose a user chooses to be on OOO from `02/10/2023` till `02/10/2023`, then both the `from` and `until` timestamps are chosen as the `00:00:00 UTC` timestamps for both. In this case, `1696204800000`.
- Hence, the `from` and `until` timestamps become equal resulting in a OOO for 0 ms. This would cause problems when updating the OOO dates in the user's discord nickname.
- With this issue, the `from` and `until` timestamp for the day must be set to `00:00:00 UTC` and `23:50:59 UTC` respectively.

Refer to the screenshot higlighting the issue
<img width="1792" alt="image" src="https://github.com/Real-Dev-Squad/website-backend/assets/51514137/f462175c-3e7c-4b42-903a-88bb96292a1e">

## Approach
- First step, a solution design documentation must be submitted highlighting the proposed solution. To proceed with development, the design solution documentation must have at least 3 reviews and approvals.
- Second step, TDD methodology must be used to fix the issue. The first PR submitted should contain the test cases, followed by another PR with the fix.
- The fix PR should only be raised when the PR with test cases have been reviewed.

heyrandhir:
ETA Breakdown:

Design Doc: 1 day
Code Changes + Tests : 1 day
Review+Buffer: 3 day

Completion : 9th Oct 2023

PR #1587, opened 2023-10-04
Update time stamp for OOO

heyrandhir:
Issue Ticket Number:-
- #1579 

Backend changes
- [x] Yes
- [ ] No

Frontend Changes
- [ ] Yes
- [x] No

Database changes
- [ ] Yes 
- [x] No

- [ ] Yes
- [x] No

### Deployment notes
None

### Description
Currently, when a user selects OOO from website-my, the timestamp for the chosen date is stored as 00:00:00 UTC. However, a problem arises when the user chooses OOO for just one day, as both the from and until timestamps are set to 00:00:00 UTC, resulting in them being equal and causing a OOO duration of 0 milliseconds. To address this issue, the from and until timestamps for a single-day OOO should be set to 00:00:00 UTC and 23:50:59 UTC, respectively, to avoid conflicts when updating the user's Discord nickname.

### Testing Stats:

Unit Tests :

models/userStatus.js

image

utils/userStatus.js
image

Integration Tests:

image

### DIFF
--- a/models/userStatus.js
+++ b/models/userStatus.js
@@ -13,6 +13,7 @@ const {
   generateErrorResponse,
   generateNewStatus,
   getNextDayTimeStamp,
+  convertTimestampsToUTC,
 } = require("../utils/userStatus");
 const { TASK_STATUS } = require("../constants/tasks");
 const userStatusModel = firestore.collection("usersStatus");
@@ -201,11 +202,12 @@ const getAllUserStatus = async (query) => {
  * @returns Promise<userStatusModel|Object>
  */
 
-const updateUserStatus = async (userId, newStatusData) => {
+const updateUserStatus = async (userId, updatedStatusData) => {
   try {
+    const newStatusData = convertTimestampsToUTC({ ...updatedStatusData });
     const userStatusDocs = await userStatusModel.where("userId", "==", userId).limit(1).get();
     const [userStatusDoc] = userStatusDocs.docs;
-    const tommorow = getTomorrowTimeStamp();
+    const tomorrow = getTomorrowTimeStamp();
     if (userStatusDoc) {
       const docId = userStatusDoc.id;
       const userStatusData = userStatusDoc.data();
@@ -223,7 +225,7 @@ const updateUserStatus = async (userId, newStatusData) => {
           newStatusData.futureStatus = {};
         }
         if (isNewStateOoo) {
-          if (newStatusData.currentStatus.from >= tommorow) {
+          if (newStatusData.currentStatus.from >= tomorrow) {
             newStatusData.futureStatus = { ...newStatusData.currentStatus };
             delete newStatusData.currentStatus;
           } else {
@@ -245,7 +247,7 @@ const updateUserStatus = async (userId, newStatusData) => {
         filterStatusData(newStatusData);
         const isNewStateOOO = newStatusData.currentStatus.state === userState.OOO;
         if (isNewStateOOO) {
-          if (newStatusData.currentStatus.from >= tommorow) {
+          if (newStatusData.currentStatus.from >= tomorrow) {
             newStatusData.futureStatus = { ...newStatusData.currentStatus };
             delete newStatusData.currentStatus;
           }
--- a/test/unit/utils/userStatus.test.js
+++ b/test/unit/utils/userStatus.test.js
@@ -121,7 +121,6 @@ describe("User Status Functions", function () {
       const isEndOfDay = true;
       const result = convertTimestampToUTCStartOrEndOfDay(timestamp, isEndOfDay);
 
-      // Expected result: 1696463999999 Wed Oct 04 2023 23:59:59 UTC
       expect(result).to.equal(null);
     });
   });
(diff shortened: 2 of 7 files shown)


utils/userStatus.js
```diff
@@ -302,6 +302,52 @@ const getNextDayTimeStamp = (timeStamp) => {
...
 };
 
+/**
+ * Converts a timestamp to either the start or end of a day in UTC time.
+ *
+ * @param {number} timestamp - The timestamp to be converted.
+ * @param {boolean} isEndOfDay - A flag indicating whether to convert to the end of the day (true) or the start of the day (false).
+ * @returns {number} The converted timestamp.
+ */
+const convertTimestampToUTCStartOrEndOfDay = (timestamp, isEndOfDay) => {
```
iamitprakash, line 312: the function does not check if the from and until timestamps are valid before converting them to UTC. This means that the function could throw an error if the timestamps are invalid.

iamitprakash (changes requested):

utils/userStatus.js
heyrandhir: Thanks for bringing this to notice. Will add safety check

heyrandhir: Changes applied. Please review again @iamitprakash

utils/userStatus.js
```diff
@@ -302,6 +302,55 @@ const getNextDayTimeStamp = (timeStamp) => {
...
+      const { from, until } = obj[key];
+      const fromType = typeof from;
+      const untilType = typeof until;
+
+      if ((fromType === "string" || fromType === "number") && String(from).trim() !== "") {
+        obj[key].from = convertTimestampToUTCStartOrEndOfDay(from, false);
+      }
+
+      if ((untilType === "string" || untilType === "number") && String(until).trim() !== "") {
+        obj[key].until = convertTimestampToUTCStartOrEndOfDay(until, true);
```
bharati-21, line 346: Nit, you can assign default value for `isEndOfDay` as true and only pass a value when `isEndOfDay` is false.

models/userStatus.js
```diff
@@ -223,7 +225,7 @@ const updateUserStatus = async (userId, newStatusData) => {
           newStatusData.futureStatus = {};
         }
         if (isNewStateOoo) {
-          if (newStatusData.currentStatus.from >= tommorow) {
+          if (newStatusData.currentStatus.from >= tomorrow) {
```
bharati-21, line 228: Can you please explain why this check is needed? Why are we checking the currentStatus's start time with the next day's date?

test/unit/utils/userStatus.test.js
```diff
@@ -91,4 +96,62 @@ describe("User Status Functions", function () {
...
+      const expectedOutput = {
+        currentStatus: {
+          from: 1696377600000, // October 4, 2023, 00:00:00 UTC
+          until: 1697155199999, // Thu Oct 12, 2023, 23:59:59 UTC
+        },
+        futureStatus: {
+          from: 1696377600000, // October 4, 2023, 00:00:00 UTC
+          until: "", // No conversion for an empty string
+        },
+      };
```
bharati-21, line 151: Please move these to fixtures and import it here.

utils/userStatus.js
```diff
@@ -302,6 +302,55 @@ const getNextDayTimeStamp = (timeStamp) => {
...
+const convertTimestampsToUTC = (obj) => {
+  const statusKeys = ["currentStatus", "futureStatus"];
+
+  for (const key of statusKeys) {
+    if (obj[key]) {
+      const { from, until } = obj[key];
+      const fromType = typeof from;
+      const untilType = typeof until;
+
+      if ((fromType === "string" || fromType === "number") && String(from).trim() !== "") {
```
Ajeyakrishna-k, line 341:
```suggestion
      if ((fromType === "string" || fromType === "number") && from) {
```

utils/userStatus.js
```diff
@@ -302,6 +302,55 @@ const getNextDayTimeStamp = (timeStamp) => {
...
+        obj[key].from = convertTimestampToUTCStartOrEndOfDay(from, false);
+      }
+
+      if ((untilType === "string" || untilType === "number") && String(until).trim() !== "") {
```
Ajeyakrishna-k, line 345:
```suggestion
      if ((untilType === "string" || untilType === "number") && until) {
```

Ajeyakrishna-k:
NIT : I think the two new time modification utils method added here could've been more generic and moved into `utils/time.js` file

utils/userStatus.js
heyrandhir: yeah currently as well if the 2nd parameter is not passed it takes beginning of day as default.

utils/userStatus.js
heyrandhir: It can be string as well as number. the `.trim` only works for string hence 2nd part is needed.

utils/userStatus.js
heyrandhir: It can be string as well as number. the `.trim` only works for string hence 2nd part is needed.

heyrandhir:
> test/unit/utils/userStatus.test.js


The convertTimestampToUTCStartOrEndOfDay function was originally a generic function, so it has been relocated to the `utils/time.js file`. However, the `convertTimestampsToUTC` function has specific requirements, as it expects the input object to adhere to a particular format with specific keys. Therefore, it was not moved along with the generic function.

Ajeyakrishna-k (approved):

bharati-21 (approved):

vinit717 (approved):

PR #1595, opened 2023-10-09
Enhancement : fetch users who have archived : false

j24m:


kotesh-arya (dismissed):
Looks Good to me

samyakshah3008 (approved):
Great work

sahsisunny (dismissed):

MayurLund (approved):

Achintya-Chatterjee (dismissed):

iamitprakash (changes requested):
In case of user not found error should not be 404

j24m:
> In case of user not found error should not be 404

- I searched for this on google, and in MDN as well.
- here is the link : https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
- I found that 404 status code is thrown if user is not found. 
- Can you tell me, what according to you should be the status code ?

iamitprakash:
> > In case of user not found error should not be 404
> 
> * I searched for this on google, and in MDN as well.
> * here is the link : https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
> * I found that 404 status code is thrown if user is not found.
> * Can you tell me, what according to you should be the status code ?

The HTTP 404 Not Found response status code indicates that the server cannot find the requested resource. Links that lead to a 404 page are often called broken or dead links and can be subject to [link rot](https://en.wikipedia.org/wiki/Link_rot).

iamitprakash (dismissed):
Need to write test case of -ve case

controllers/users.js
```diff
@@ -142,9 +142,11 @@ const getUsers = async (req, res) => {
         try {
           result = await dataAccess.retrieveUsers({ discordId });
           user = result.user;
-
           if (!result.userExists) {
-            return res.boom.notFound("User doesn't exist");
+            return res.json({
+              message: "No user found",
```
iamitprakash, line 147: change to bit meanningful message like "User not found with Id {<whatever id is>}"

test/integration/users.test.js
```diff
@@ -679,12 +679,12 @@ describe("Users", function () {
...
 
-    it("Should return an error when passing an invalid Discord ID", async function () {
+    it("Should return an empty object when passing an invalid Discord ID", async function () {
       const invalidDiscordId = "50485556209423";
       const res = await chai.request(app).get(`/users?dev=true&discordId=${invalidDiscordId}`);
-      expect(res).to.have.status(404);
+      expect(res).to.have.status(200);
       expect(res.body).to.be.a("object");
-      expect(res.body.message).to.equal("User doesn't exist");
+      expect(res.body.message).to.equal("No user found");
```
iamitprakash, line 687: same here

iamitprakash (changes requested):

controllers/users.js
vinit717, line 147:
```suggestion
              message: "user not found",
```

j24m: We are fetching users on the basis of a query param discordId=12345 which is self explanatory in my opinion. So if user not found then we already know through query param what discordId it was.

test/integration/users.test.js
```diff
@@ -679,12 +679,12 @@ describe("Users", function () {
       expect(res.body.message).to.equal("Route not found");
     });
 
-    it("Should return an error when passing an invalid Discord ID", async function () {
+    it("Should return an empty object when passing an invalid Discord ID", async function () {
       const invalidDiscordId = "50485556209423";
       const res = await chai.request(app).get(`/users?dev=true&discordId=${invalidDiscordId}`);
-      expect(res).to.have.status(404);
+      expect(res).to.have.status(200);
```
vinit717, line 685: why there is a success status of 200 if a discord ID is wrong?

models/users.js
```diff
@@ -260,7 +260,7 @@ const fetchUser = async ({ userId = null, username = null, githubUsername = null
         userData = doc.data();
       });
     } else if (discordId) {
-      const user = await userModel.where("discordId", "==", discordId).limit(1).get();
+      const user = await userModel.where("discordId", "==", discordId).where("roles.archived", "==", false).get();
```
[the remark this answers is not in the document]
j24m: image

test/unit/models/users.test.js
```diff
@@ -128,9 +128,9 @@ describe("users", function () {
     afterEach(async function () {
       await cleanDb();
     });
-    it("It should have discordId field", async function () {
+    it("It should fetch users who have archived:false role", async function () {
```
vinit717, line 131: why change the test name to this?

j24m: Now we are checking on the basis of role archived:false hence the change.

controllers/users.js
j24m: **No user found** v/s **user not found** are same things.

test/integration/users.test.js
j24m: This was changed as per discussion with @iamitprakash  and @ankushdharkar.

controllers/users.js
vinit717: but user not found is more appropriate from my point of 
ask @iamitprakash for confirmation

test/integration/users.test.js
vinit717: But here you are testing for the wrong ID then it should be 404

test/unit/models/users.test.js
vinit717: Then make the test name more generic like it should fetch users based on discord id

models/users.js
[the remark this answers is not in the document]
j24m: Already done.

DashDeipayan (approved):

controllers/users.js
satyam73: +1

kotesh-arya (approved):

kotesh-arya (approved):

vinit717 (approved):

Achintya-Chatterjee (approved):

Issue #1449, opened 2023-08-25
Update Query Parameter Format in API Endpoint


sahsisunny:
**Description:**
The current format of the query parameter in our API endpoint `users` is as follows:
`/users?query=filterBy:UNMERGED_PRS+days:<days>`

We would like to update the format to make it more consistent with industry standards and to improve usability. The new format will be:

`/users?q=filterBy:UNMERGED_PRS+days:<days>`

This change will not only make the API endpoint more intuitive for users but also align it with common query parameter naming conventions.

**Acceptance Criteria:**

- [ ] Update the API endpoint `/users` to accept the query parameter in the new format `/users?q=filterBy:UNMERGED_PRS+days:<days>`.
- [ ] Ensure that the old query parameter format `/users?query=filterBy:UNMERGED_PRS+days:<days>` is not deprecated you have to add the new query behind the feature flag.
- [ ] Update relevant documentation and API references to reflect the new query parameter format.
- [ ] Perform thorough testing to ensure that the updated format does not introduce any regressions or issues in functionality.
- [ ] Notify internal teams and external API consumers about this change and provide them with updated documentation.
- [ ] Update any automated tests or scripts that rely on the old query parameter format.

skt17:
ETA 10 September

PR #1603, opened 2023-10-13
 update query paramter format from query to q for users endpoint

suryxks:
Issue
-----

-   [Update Query Parameter Format in API Endpoint #1449](https://github.com/Real-Dev-Squad/website-backend/issues/1449)

Description
-----------

-   Update Query Parameter Format in API Endpoint.

Backend changes

-   [x]  Yes
-   [ ]  No

Frontend Changes

-   [ ]  Yes
-   [x]  No

Database changes

-   [ ]  Yes
-   [x]  No

-   [ ]  Yes
-   [x]  No

Is Under Feature Flag 

-   [x]  Yes
-   [ ]  No

### Deployment notes

None

Test stats
----------

backend-test-1

-   Before :\
backend-before

-   After :\
backend-after

Working Proof
----------------

backend-result-0
backend-result-1
backend-result-2
Route-not-found

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -80,10 +80,11 @@ const getUserById = async (req, res) => {
 const getUsers = async (req, res) => {
   try {
     // getting user details by id if present.
-    const query = req.query?.query ?? "";
-    const transformedQuery = parseSearchQuery(query);
-    const qualifiers = getQualifiers(query);
-
+    const { q, dev: devParam, query } = req.query;
+    const dev = devParam === "true";
+    const queryString = (dev ? q : query) || "";
+    const transformedQuery = parseSearchQuery(queryString);
+    const qualifiers = getQualifiers(queryString);
     // getting user details by id if present.
     if (req.query.id) {
       const id = req.query.id;
@@ -135,7 +136,6 @@ const getUsers = async (req, res) => {
     // getting user details by discord id if present.
     const discordId = req.query.discordId;
 
-    const dev = req.query.dev === "true";
     if (req.query.discordId) {
       if (dev) {
         let result, user;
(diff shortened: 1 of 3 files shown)


iamitprakash (approved):

controllers/users.js
```diff
@@ -80,11 +80,17 @@ const getUserById = async (req, res) => {
...
-
+    const { q, dev: devParam, query } = req.query;
+    const dev = devParam === "true";
+    const queryString = (dev ? q : query) || "";
+    const transformedQuery = parseSearchQuery(queryString);
+    const qualifiers = getQualifiers(queryString);
+    // Should throw an error if the new query parameter is without feature flag
+    if (q && !dev) {
+      return res.boom.notFound("Route not found");
+    }
```
shubhamsigdar1, line 91: Perhaps we could optimize the code flow by considering feature flag checks earlier in the process. Currently, we run all the functionality, including database access and user retrieval, before verifying the presence of a feature flag. This approach might have implications for performance and efficiency. It might be worth exploring ways to introduce feature flag checks at an earlier stage to improve the overall system performance and response times.

so please check this condition before.


Correct me if i written something wrong above

suryxks: The steps that are done before checking the presence of feature flag is just parsing the query string and transforming it to an object and no database access and user retrieval is done here.I think  parsing the query string and converting it to object is not going to make much difference in performance here. Please correct me if I am wrong.

shubhamsigdar1 (approved):

shubhamsigdar1 (approved):

Achintya-Chatterjee (approved):

PR #1630, opened 2023-10-27
Improve response time of overdue task  user API

sahsisunny:
**Description:** 
In this PR for improving the performance of /users?query=filterBy:overdue_tasks API endpoint, I have made following changes:

1. Fetching all users data at once instead of fetching one by one.
2. Fetching only those users data which have archived false and not in_discord roles.
3. Add early return if no overdue tasks found.

**Test Report**
image
image


vinit717:
Does your Discord command work now?

controllers/users.js
```diff
@@ -174,33 +174,32 @@ const getUsers = async (req, res) => {
     if (transformedQuery?.filterBy === OVERDUE_TASKS) {
       try {
         const tasksData = await getOverdueTasks(days);
+        if (!tasksData.length) {
+          return res.json({
+            message: "No users found",
+            users: [],
```
Pratiyushkumar, line 180: NIT - can we not return the message saying `No overdue task found` ?

sahsisunny:
> Does your Discord command work now?

it should work

Pratiyushkumar (approved):
everything looks good to me expect one comment...you can  choose to fix it or leave it

controllers/users.js
sahsisunny: No, we cannot do that since the API response contains information about returning users.

test/unit/models/users.test.js
```diff
@@ -475,4 +475,18 @@ describe("users", function () {
...
+    it("It should fetch users based on userids", async function () {
+      const userArr = userData();
+      const userIds = [];
+      for (let i = 0; i < userArr.length; i++) {
+        const userId = await addUser(userArr[i]);
+        userIds.push(userId);
+      }
+      const result = await users.fetchUser({ userIds });
+      expect(result).to.haveOwnProperty("users");
+      expect(result.users).to.have.length(15);
```
gauravsinhaweb, line 489:
```suggestion
    it("It should fetch users based on userids", async function () {
      const userArr = userData();
      const userIds = [];
      for (const user of userArr) {
       const userId = await addUser(user);
        userIds.push(userId);
      }
      const result = await users.fetchUser({ userIds });
      expect(result).to.haveOwnProperty("users");
      expect(result.users).to.have.length(15);
```

test/unit/models/users.test.js
```diff
@@ -475,4 +475,18 @@ describe("users", function () {
...
   });
+
+  describe("fetchUser", function () {
+    it("It should fetch users based on userids", async function () {
+      const userArr = userData();
+      const userIds = [];
+      for (let i = 0; i < userArr.length; i++) {
+        const userId = await addUser(userArr[i]);
+        userIds.push(userId);
+      }
```
gauravsinhaweb, line 486:
```suggestion
      for (const user of userArr) {
      const userId = await addUser(user);
      userIds.push(userId);
      }
```

gauravsinhaweb (approved):
NiT:

vinit717:
> > Does your Discord command work now?
> 
> it should work

Have you checked locally?

gauravsinhaweb (approved):

sahsisunny:
> > > Does your Discord command work now?
> > 
> > 
> > it should work
> 
> Have you checked locally?

How?

Pratiyushkumar (dismissed):

models/users.js
```diff
@@ -249,6 +255,23 @@ const fetchUser = async ({ userId = null, username = null, githubUsername = null
         id = doc.id;
         userData = doc.data();
       });
+    } else if (userIds) {
+      const users = await userModel
+        .where(admin.firestore.FieldPath.documentId(), "in", userIds)
+        .where("roles.archived", "==", false)
```
prakashchoudhary07, line 261: firestore IN has a limit of 30 values, I think 
So what will happen if there are 30+ userIds here?

prakashchoudhary07 (changes requested):

models/users.js
sahsisunny: You are right, Firestore has a limit of 30 values for the 'in' operator. I have fixed this by fetching the data in smaller chunks now. Thanks for pointing that out.

gauravsinhaweb (dismissed):

Pratiyushkumar (approved):
LGTM

test/integration/users.test.js
```diff
@@ -761,9 +766,14 @@ describe("Users", function () {
...
-          expect(res.body.users).to.be.a("array");
-          return done();
+          if (res.body.users.length > 0) {
+            expect(res.body.message).to.equal("Users returned successfully!");
+            expect(res.body.users).to.be.a("array");
+            return done();
+          } else {
+            expect(res.body.message).to.equal("No users found");
+            return done();
+          }
```
prakashchoudhary07, line 776: The test shouldn't have any conditions in it,
It should test both cases separately

models/discordactions.js
```diff
@@ -246,7 +246,7 @@ const enrichGroupDataWithMembershipInfo = async (discordId, groups = []) => {
     const subscribedGroupIds = findSubscribedGroupIds(discordId, groupsToUserMappings);
 
     return groups.map((group) => {
-      const groupCreator = groupCreatorsDetails[group.createdBy];
+      const groupCreator = groupCreatorsDetails.find((user) => user.id === group.createdBy);
```
prakashchoudhary07, line 249: Why is this being changed here?

shubhamsigdar1:
> * Fetching only those users data which have archived false and not in_discord roles.


I have confusion in above point - What do you mean by not in_discord roles??

models/discordactions.js
sahsisunny: Because of [this](https://github.com/Real-Dev-Squad/website-backend/pull/1630/files#diff-9a5062203df8184c1d64d655f2dcd16eebeea484b834b0f461771b463037ceefL668-R668) . I changed the user response format from object to array.

sahsisunny:
> in_discord

You don't know about `in_discord` role?

shubhamsigdar1:
> > in_discord
> 
> You don't know about `in_discord` role?

You said user not have in_discord role
but every user have in_discord role

Correct me if I understand wrong

shubhamsigdar1 (approved):
LGTM

Pratiyushkumar (approved):

prakashchoudhary07 (approved):

PR #1633, opened 2023-10-28
Discard user to udpate their roles

vinit717:
### Purpose
To prevent a user from updating their roles as it may cause a security concern and allow only those user who want to join as maven and have not joined Discord yet

### Backend Changes
- [x] Yes
- [ ] No

### Is Under Feature Flag
- [ ] Yes
- [x] No

### Database changes: 
- [ ] Yes
- [x] No

### Breaking changes (If your feature is breaking/missing something please mention pending tickets): 
- [ ] Yes
- [x] No

### Deployment notes:
NA

### Tested in local
- [x] Yes
- [ ] No

**When in_discord is false**
image

**When in_discord is true** 
image

**Tests Stats:**
image

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -363,6 +363,8 @@ const updateSelf = async (req, res) => {
   try {
     const { id: userId } = req.userData;
     const { user } = await dataAccess.retrieveUsers({ id: userId });
+    // eslint-disable-next-line no-console
+    console.log("usrt", req.userdata);
 
     if (req.body.username) {
       if (!user.incompleteUserDetails) {
@@ -372,7 +374,7 @@ const updateSelf = async (req, res) => {
     }
 
     if (req.body.roles) {
-      if (user && user.roles.in_discord) {
+      if (user && user.roles.in_discord && !user.roles.developer) {
         return res.boom.forbidden("Cannot update roles");
       }
     }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -363,8 +363,6 @@ const updateSelf = async (req, res) => {
   try {
     const { id: userId } = req.userData;
     const { user } = await dataAccess.retrieveUsers({ id: userId });
-    // eslint-disable-next-line no-console
-    console.log("usrt", req.userdata);
 
     if (req.body.username) {
       if (!user.incompleteUserDetails) {
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -372,7 +372,7 @@ const updateSelf = async (req, res) => {
     }
 
     if (req.body.roles) {
-      if (user && user.roles.in_discord && user.roles.developer) {
+      if ((user && user.roles.in_discord) || (user && user.roles.in_discord)) {
         return res.boom.forbidden("Cannot update roles");
       }
     }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -372,7 +372,7 @@ const updateSelf = async (req, res) => {
     }
 
     if (req.body.roles) {
-      if ((user && user.roles.in_discord) || (user && user.roles.in_discord)) {
+      if (user && (user.roles.in_discord || user.roles.developer)) {
         return res.boom.forbidden("Cannot update roles");
       }
     }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -372,7 +372,7 @@ const updateSelf = async (req, res) => {
     }
 
     if (req.body.roles) {
-      if (user && user.roles.in_discord && !user.roles.developer) {
+      if (user && user.roles.in_discord && user.roles.developer) {
         return res.boom.forbidden("Cannot update roles");
       }
     }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -379,7 +379,7 @@ const updateSelf = async (req, res) => {
 
     const updatedUser = await userQuery.addOrUpdate(req.body, userId);
 
-    if (updatedUser.isNewUser) {
+    if (!updatedUser.isNewUser) {
       // Success criteria, user finished the sign-up process.
       userQuery.initializeUser(userId);
       return res.status(204).send();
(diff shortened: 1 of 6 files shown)


middlewares/validators/user.js
```diff
@@ -52,9 +52,6 @@ const updateUser = async (req, res, next) => {
         .optional(),
       discordId: joi.string().optional(),
       roles: joi.object().keys({
-        archived: joi.boolean().required(),
-        in_discord: joi.boolean().required(),
-        developer: joi.boolean().optional(),
```
heyrandhir, line 57: this validator function is being used at multiple places. Are you sure removing here will not affect other places ?

controllers/users.js
```diff
@@ -362,18 +362,25 @@ const getSelfDetails = async (req, res) => {
...
         return res.boom.forbidden("Cannot update username again");
       }
       await userQuery.setIncompleteUserDetails(userId);
     }
 
-    const user = await userQuery.addOrUpdate(req.body, userId);
+    if (req.body.roles) {
+      if (user && user.roles.in_discord) {
+        return res.boom.forbidden("Cannot update roles");
+      }
```
heyrandhir, line 377: you have removed the `in_discord` in the validator file then it will not be there in the body Right ? The validator will throw an error of unknown key i.e Bad Request. And the code will not reach here. Correct me If I am wrong ?

test/integration/users.test.js
```diff
@@ -121,24 +121,22 @@ describe("Users", function () {
...
+            maven: true,
           },
         })
         .end((err, res) => {
           if (err) {
             return done(err);
           }
 
-          expect(res).to.have.status(204);
+          expect(res).to.have.status(403);
```
heyrandhir, line 139: why are we changing the status code from 204 to 403 ?

middlewares/validators/user.js
vinit717: It is not being used in any user route apart from this

controllers/users.js
vinit717: Here this case is only arise when user sign-in with GitHub and not joined discord yet so we can allow user who may join as maven, designer, product manager

test/integration/users.test.js
vinit717: because we are not allowing user to update their role

test/integration/users.test.js
```diff
@@ -121,24 +122,45 @@ describe("Users", function () {
         });
     });
 
-    it("Should update the user roles", function (done) {
+    // eslint-disable-next-line mocha/no-exclusive-tests
+    it.only("Should allow updating user role when in_discord is not present", function (done) {
```
vinayak-trivedi, line 126: why is this pushed?

controllers/users.js
```diff
@@ -362,18 +362,25 @@ const getSelfDetails = async (req, res) => {
...
-      const { user } = await dataAccess.retrieveUsers({ id: userId });
       if (!user.incompleteUserDetails) {
         return res.boom.forbidden("Cannot update username again");
       }
       await userQuery.setIncompleteUserDetails(userId);
     }
 
-    const user = await userQuery.addOrUpdate(req.body, userId);
+    if (req.body.roles) {
+      if (user && user.roles.in_discord) {
```
vinayak-trivedi, line 375: should we not add developers role as well here?

vinayak-trivedi: why? please explain, so let's suppose I get kicked out of the group, I know that maven, product manager and developers tag allows me to directly generate link, I hit that API with those roles, and get a link for it, and then I change it

vinit717: Added
thanks for pointing this out

test/integration/users.test.js
vinit717: removed

controllers/users.js
```diff
@@ -362,18 +362,27 @@ const getSelfDetails = async (req, res) => {
 const updateSelf = async (req, res) => {
   try {
     const { id: userId } = req.userData;
+    const { user } = await dataAccess.retrieveUsers({ id: userId });
+    // eslint-disable-next-line no-console
+    console.log("usrt", req.userdata);
```
satyam73, line 367: remove this log please

test/integration/users.test.js
```diff
@@ -121,24 +122,44 @@ describe("Users", function () {
...
+            if (err) {
+              return done(err);
+            }
+
+            expect(res).to.have.status(204);
+            return done();
+          });
+      });
+    });
+    it("Should not update the user roles", function (done) {
```
satyam73, line 147: "Should not update the user roles" but when? what are the condition?
also cookie name `jwt` doesn't give context about for which the token is made for whether it's `superUserToken`, `memberToken` or `normalUserToken`

controllers/users.js
```diff
@@ -362,18 +362,27 @@ const getSelfDetails = async (req, res) => {
...
+    }
+
+    const updatedUser = await userQuery.addOrUpdate(req.body, userId);
 
-    if (!user.isNewUser) {
-      // Success criteria, user finished the sign up process.
+    if (!updatedUser.isNewUser) {
+      // Success criteria, user finished the sign-up process.
       userQuery.initializeUser(userId);
       return res.status(204).send();
```
satyam73, line 387: Aren't we sending any message in the response?

satyam73 (changes requested):

controllers/users.js
vinit717: That's not under this PR I have only changed the variable name after you look at this function I think you'll be more clear on what I have done

controllers/users.js
vinit717: done

test/integration/users.test.js
vinit717: Updated and `jwt` is basically a normal RDS user

satyam73 (dismissed):
LGTM

heyrandhir (approved):

satyam73 (approved):

PR #1653, opened 2023-11-01
Fix Issue: Update Specific Role Without Removing Others

vinit717:
### Developer Name: vinit 

Issue: #1654

### Purpose
By adding new role data with the existing roles in the Firestore document then the previous roles got removed. This change will ensure that the update only affects the specified role without affecting other roles that may already be present.

### Backend Changes
- [x] Yes
- [ ] No

### Is Under Feature Flag
- [ ] Yes
- [x] No

### Database changes: 
- [ ] Yes
- [x] No

### Breaking changes (If your feature is breaking/missing something please mention pending tickets): 
- [ ] Yes
- [x] No

### Deployment notes:
NA

### Tested in local
- [ ] Yes
- [x] No

**Tests Stats**
image

### DIFF
--- a/test/integration/users.test.js
+++ b/test/integration/users.test.js
@@ -146,11 +146,7 @@ describe("Users", function () {
     });
 
     it("Should not remove old roles when updating user roles", async function () {
-      let newUserId;
-      await addUser(newUser).then((userId) => {
-        newUserId = userId;
-      });
-
+      const newUserId = await addUser(newUser);
       const newUserJwt = authService.generateAuthToken({ userId: newUserId });
 
       const getUserResponseBeforeUpdate = await chai
--- a/test/integration/users.test.js
+++ b/test/integration/users.test.js
@@ -144,6 +144,47 @@ describe("Users", function () {
           });
       });
     });
+
+    it("Should not remove old roles when updating user roles", async function () {
+      let newUserId;
+      await addUser(newUser).then((userId) => {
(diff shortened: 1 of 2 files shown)


prakashchoudhary07 (dismissed):

sahsisunny (dismissed):

heyrandhir (approved):

heyrandhir (commented):
Are tests not needed for this. Please confirm

vinit717:
> Are tests not needed for this. Please confirm

Added some tests Please check

test/integration/users.test.js
```diff
@@ -144,6 +144,47 @@ describe("Users", function () {
           });
       });
     });
+
+    it("Should not remove old roles when updating user roles", async function () {
+      let newUserId;
+      await addUser(newUser).then((userId) => {
+        newUserId = userId;
+      });
```
heyrandhir, line 152: Nit : If we are using async await, it is better to avoid chaining .then() promises, as this can lead to code that is difficult to read and maintain.

heyrandhir (approved):
LGTM. but I have a suggestion for improvement.

test/integration/users.test.js
vinit717: Thanks for pointing out I have removed

heyrandhir (approved):

shreya-mishra (approved):
lgtm

PR #1660, opened 2023-11-02
Added "-" in validation for generating username

vinit717:
Developer Name: @vinit717 

Issue Ticket Number:- #1661 

**Description**
Added "-" in regex to validate the username and write its test case

Backend changes
- [x] Yes
- [ ] No

Frontend Changes
- [ ] Yes
- [x] No

Is Under the Feature Flag 
- [ ] Yes
- [x] No

Database changes
- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Deployment notes
Any special instructions for the deployment (none)

Tested in staging
- [ ] Yes
- [x] No

**Before Changes:**
image

**After Changes:**
image

**Test Stats:**
image

### DIFF
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -22,7 +22,7 @@ const updateUser = async (req, res, next) => {
         .optional()
         .min(4)
         .max(20)
-        .regex(/^[a-zA-Z0-9]+$/)
+        .regex(/^[a-zA-Z0-9-]+$/)
         .message("Username must be between 4 and 20 characters long and contain only letters or numbers."),
       first_name: joi.string().optional(),
       last_name: joi.string().optional(),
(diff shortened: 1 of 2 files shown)


samyakshah3008 (approved):
LGTM 🚀

middlewares/validators/user.js
```diff
@@ -22,7 +22,7 @@ const updateUser = async (req, res, next) => {
         .optional()
         .min(4)
         .max(20)
-        .regex(/^[a-zA-Z0-9]+$/)
+        .regex(/^[a-zA-Z0-9-]+$/)
```
gauravsinhaweb, line 25: Can we define this regex in a variable and use it, it would be descriptive

shubhamsigdar1 (approved):

gauravsinhaweb (approved):

middlewares/validators/user.js
vinit717: I think no need to add in variable as it is not use anywhere and it is self explanatory

PR #1405, opened 2023-08-12
API for applications

vinayak-trivedi:
closes #1282 

Backend changes
- [x] Yes
- [ ] No

Frontend Changes
- [ ] Yes
- [x] No

Is Under Feature Flag
- [ ] Yes
- [x] No

Database changes
- [ ] Yes
- [x] No

Breaking changes (If your feature is breaking/missing something please mention pending tickets)
- [ ] Yes
- [x] No

Tested in staging
- [ ] Yes
- [x] No

Controllers file
image

Model files
image

Router file
image

Authorize own middleware
image

Validator
image

PR #1689, opened 2023-11-17
Revert "API for applications"

prakashchoudhary07:
Reverts Real-Dev-Squad/website-backend#1405

Reason: deployment fails
image


prakashchoudhary07:
Reverting this as its breaking deployment

PR #1691, opened 2023-11-17
Batch operation to Migrate status COMPLETED to DONE in tasks model

shubhamsigdar1:
Date: 18-11-2023

Developer Name: @shubhamsigdar1 

----

## Description: 

- ### **Please refer this pr also while reviewing**: https://github.com/Real-Dev-Squad/website-backend/pull/1692
- Added service file which is used to perform batch operation and update all the task status COMPLETED to DONE
- Design Doc: https://docs.google.com/document/d/14reXnCwvnaQcjwKgfZweSCXdvLxjjdAA7Tr_DZOqJ84/edit?usp=sharing
- Written Test

Is Under Feature Flag 
- [ ] Yes
- [X] No

Database changes
- [X] Yes
- [ ] No

Breaking changes (If your feature is breaking/missing something please mention pending tickets)
- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

### Test Coverage:

**services/tasks.js**

image

### After Change video

https://github.com/Real-Dev-Squad/website-backend/assets/107163260/2300aa8b-b8c1-4b5c-90ca-f3447e02cf2f

PR #1692, opened 2023-11-17
Script to Migrate status COMPLETED to DONE in tasks model

shubhamsigdar1:
Date: 18-11-2023

Developer Name: @shubhamsigdar1 

## Description: 

- ### Please refer this pr also while reviewing: https://github.com/Real-Dev-Squad/website-backend/pull/1691
- Design Doc: https://docs.google.com/document/d/14reXnCwvnaQcjwKgfZweSCXdvLxjjdAA7Tr_DZOqJ84/edit?usp=sharing
- Added route /migration
- Added controller and model function 
- model function fetch all the tasks of status: "COMPLETED" and add into array + chunks of 500 element
- Written test 

Is Under Feature Flag 
- [ ] Yes
- [X] No

Database changes
- [X] Yes
- [ ] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

### Test Coverage

**Controllers/tasks.js**

image

**Models/tasks.js**

image

### After Changes video:

https://github.com/Real-Dev-Squad/website-backend/assets/107163260/fcac7503-feea-40b1-a81b-44df0146cc6e


routes/tasks.js
```diff
@@ -40,4 +40,6 @@ router.patch(
 );
 router.patch("/assign/self", authenticate, invalidateCache({ invalidationKeys: [ALL_TASKS] }), tasks.assignTask);
 
+router.post("/migration", authenticate, authorizeRoles([SUPERUSER]), tasks.updateStatus);
+
```
vinit717, line 44: This endpoint is not clear to me look for better conventions

vinit717 (commented):
Please mention the test stats

routes/tasks.js
shubhamsigdar1: Thank you for your feedback regarding the /migration endpoint. The term "migration" has been intentionally selected to align with the semantic meaning of a one-time data transformation or update.
If you have any further concerns or suggestions for improvement, I am open to discussion.

models/tasks.js
```diff
@@ -575,6 +579,51 @@ const getOverdueTasks = async (days = 0) => {
...
+
+    snapshot.forEach((task) => {
+      const id = task.id;
+      const taskData = task.data();
+      tasksStatusCompleted.push({ ...taskData, id });
+    });
+    const taskStatusCompletedChunks = chunks(tasksStatusCompleted, DOCUMENT_WRITE_SIZE);
+
+    for (const tasks of taskStatusCompletedChunks) {
+      const res = await updateTaskStatusToDone(tasks);
```
ardourApeX, line 606: The DOCUMENT_WRITE_SIZE is set to 500. Ultimately, it will loop over each document, update it, and only then proceed to the next document. Wouldn't it be better if we reduced it to 100 and updated 100 * 5 concurrently?

<pre>

const res = []
for (const tasks of taskStatusCompletedChunks) {
      res.push(updateTaskStatusToDone(tasks))
}
const batchResult = await Promise.all(res)
//You can update summary through each index

</pre>

The current implementation may cause server timeout.

shubhamsigdar1: A DOCUMENT_WRITE_SIZE of 500 ensures a single network call to Firestore, offering better efficiency.

However, I understand your concern about potential server timeouts. To address this, you've proposed dividing the batch size to 100 and updating 100 * 5 documents concurrently. While this approach increases the number of network calls, Which is inefficient. As the number of network calls increases, the overall time to complete the operation grows and number of error also increase (In my case I think I have to update more than 3000 document )

It's crucial to note that in the batch operation, if any individual operation fails, it won't trigger a rollback, leaving the already updated documents unchanged. 

But your concern is also non avoidable and server timeout totally depend on firestore execution timeout

ardourApeX (approved):

vinit717 (dismissed):

sahsisunny (dismissed):

vinit717 (dismissed):

prakashchoudhary07 (approved):

PR #1755, opened 2023-12-07
Fix api failure when a single user discord details were not present

Ajeyakrishna-k:
Date: 7 Dec 2023

Developer Name: @Ajeyakrishna-k  @bharati-21 

----

## Description: 
Api failed when a single user's details were missing while update their discord nickname. This fix handles those errors gracefully and updates remaining users discord names.

Is Under Feature Flag 
- [ ] Yes
- [x] No

Database changes
- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

### Add relevant Screenshot below ( e.g test coverage etc. )
Screenshot 2023-12-07 at 8 01 02 PM
[
Screenshot 2023-12-07 at 8 00 38 PM
](url)

### DIFF
--- a/models/discordactions.js
+++ b/models/discordactions.js
@@ -470,7 +470,7 @@ const updateUsersNicknameStatus = async (lastNicknameUpdate) => {
 
     const today = new Date().getTime();
 
-    const nicknameUpdatePromises = [];
+    let successfulUpdates = 0;
     const nicknameUpdateBatches = [];
     const totalUsersStatus = usersStatusDocs.length;
 
@@ -505,13 +505,18 @@ const updateUsersNicknameStatus = async (lastNicknameUpdate) => {
         }
       });
 
-      const settledPromises = await Promise.all(promises);
-      nicknameUpdatePromises.push(...settledPromises);
+      const settledPromises = await Promise.allSettled(promises);
+
+      settledPromises.forEach((result) => {
+        if (result.status === "fulfilled" && !!result.value) {
+          successfulUpdates++;
+        } else {
+          logger.error(`Error while updating nickname: ${result.reason}`);
+        }
+      });
 
       await new Promise((resolve) => setTimeout(resolve, 5000));
     }
-
-    const successfulUpdates = nicknameUpdatePromises.length;
 
     const res = {
       totalUsersStatus,
--- a/test/integration/discordactions.test.js
+++ b/test/integration/discordactions.test.js
@@ -40,7 +40,7 @@ const { updateUserStatus } = require("../../models/userStatus");
 const { generateUserStatusData } = require("../fixtures/userStatus/userStatus");
 const { getDiscordMembers } = require("../fixtures/discordResponse/discord-response");
 const { getOnboarding31DPlusMembers } = require("../fixtures/discordResponse/discord-response");
-
+const discordRolesModel = require("../../models/discordactions");
 chai.use(chaiHttp);
 const { userStatusDataForOooState } = require("../fixtures/userStatus/userStatus");
 const { generateCronJobToken } = require("../utils/generateBotToken");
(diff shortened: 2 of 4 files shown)


models/discordactions.js
```diff
@@ -505,14 +505,19 @@ const updateUsersNicknameStatus = async (lastNicknameUpdate) => {
...
-      nicknameUpdatePromises.push(...settledPromises);
+      const settledPromises = await Promise.allSettled(promises);
+
+      settledPromises.forEach((result) => {
+        if (result.status === "fulfilled" && !!result.value) {
+          successfulUpdates++;
+        } else {
+          logger.error(`Error while updating nickname: ${result.reason}`);
+        }
+      });
```
bhtibrewal, line 516: why are we not doing any try, catch-error handling here?
how are we conveying this error to the user? is it somewhere else that I am missing ?

Ajeyakrishna-k: > why are we not doing any try, catch-error handling here?
Its present in all the functions

We are okay if 1 out of 100 user nickname update fails. We don't necessarily have to return status 500 for that. We are conveying that using the `successfulUpdates` count in response.

bharati-21: But the problem with this approach is that let's say I am on OOO and my update in the status document was done at time 0ms. Say the cron job ran and it tried to update all the available status whose update was done >= 0ms. Now in case my nickname update failed and we allSettle the promise, the nickname update will be successful and the cron job's lastUpdate timestamp will update say to 1ms. Now next time cron job runs, it won't pick up my status. And my OOO will not be marked.

test/integration/discordactions.test.js
```diff
@@ -460,11 +475,11 @@ describe("Discord actions", function () {
             return done(err);
           }
 
-          expect(res).to.have.status(500);
+          // expect(res).to.have.status(500);
```
bharati-21, line 478: Pls remove comment

models/discordactions.js
Ajeyakrishna-k: - In this case should the cron job handle when it updates the time? 
- In terms of our API, should we send status 500 stating server issues when in reality only one operation failed, out of 100s of mutually exclusive ones?

bharati-21: > In this case should the cron job handle when it updates the time?
Can you elaborate as to what you are suggesting?

> In terms of our API, should we send status 500 stating server issues when in reality only one operation failed, out of 100s of mutually exclusive ones?
Ideally, such operations should be a transaction. But maybe we can discuss this more in depth and come to a conclusion.

bharati-21: As discussed, we will not be failing the entire call if any of the update for the nickname fails (i.e. this will no longer be a transaction). We will only update the number of calls failed and updated. 
If there were 0 failed updates then in the cron job we will update the `lastNicknameUpdated` timestamp. Else, the timestamp will remain the same.

bharati-21 (approved):

test/integration/discordactions.test.js
```diff
@@ -464,7 +479,7 @@ describe("Discord actions", function () {
           expect(res.body.message).to.equal("An internal server error occurred");
           return done();
         });
-    });
+    }).timeout(10000);
```
prakashchoudhary07, line 482: Why are we waiting for 10s here to complete the test?

prakashchoudhary07 (changes requested):

test/integration/discordactions.test.js
Ajeyakrishna-k: This extends timeout for the test. The test doesn't necessarily have to take 10s.

This has been done in a lot of other tests for this method to avoid the hard coded delay. Will solve this in a separate PR.

prakashchoudhary07: Please create an issues an add the link here,
Approving it for now

Ajeyakrishna-k: https://github.com/Real-Dev-Squad/website-backend/pull/1772

prakashchoudhary07 (approved):

PR #1773, opened 2023-12-12
Application update script

vinayak-trivedi:
Date: 27/12/2023

----

## Description: 
This PR adds an endpoint for running script which adds createdAt null to the existing applications in our backend, so as per our new requirement we are ordering the application with this field which is not there in the older applications doc. So they will be ignored when ordering.

TLDR: We are trying to create an `index` on `createdAt`, some older fields does't have them, so an API to backfill them with default(NULL) values.

Is Under Feature Flag 
- [x] Yes
- [ ] No

Database changes
- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [x] Yes
- [ ] No

Test coverage:

controllers:
image

models:
image

routes
image

Local working proof:
image


models/applications.ts
```diff
@@ -1,6 +1,56 @@
...
+const batchUpdateApplications = async () => {
+  try {
+    const operationStats = {
+      failedApplicationUpdateIds: [],
+      totalFailedApplicationUpdates: 0,
+      totalApplicationUpdates: 0,
+    };
+
+    const updatedApplications = [];
+    const applications = await ApplicationsModel.get();
```
vikhyat187, line 16: Hey Vinayak, can you share some context about this application model, what it does?

models/applications.ts
```diff
@@ -1,6 +1,56 @@
...
+    const multipleApplicationUpdateBatch = [];
+    const chunkedApplication = chunks(updatedApplications, FIRESTORE_BATCH_OPERATIONS_LIMIT);
+
+    chunkedApplication.forEach(async (applications) => {
+      const batch = firestore.batch();
+      applications.forEach(({ id, data }) => {
+        batch.update(ApplicationsModel.doc(id), data);
+      });
+      try {
+        await batch.commit();
```
vikhyat187, line 39: Just asking, do we have the limit on the number of record being saved at a time?

models/applications.ts
```diff
@@ -1,6 +1,56 @@
...
+        multipleApplicationUpdateBatch.push(batch);
+      } catch (error) {
+        operationStats.totalFailedApplicationUpdates += applications.length;
```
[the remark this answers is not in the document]
vikhyat187: So here we are trying to say that all the applications failed from being updated, so we are storing the count of failed in the variable.

models/applications.ts
```diff
@@ -1,6 +1,56 @@
...
+      const taskData = application.data();
+      taskData.createdAt = null;
+      updatedApplications.push({ id: application.id, data: taskData });
+    });
+
+    const multipleApplicationUpdateBatch = [];
+    const chunkedApplication = chunks(updatedApplications, FIRESTORE_BATCH_OPERATIONS_LIMIT);
+
+    chunkedApplication.forEach(async (applications) => {
+      const batch = firestore.batch();
```
vikhyat187, line 34: what is the batch size here?

vikhyat187:
Hi @vinayak-trivedi can you share the bigger picture, where is this going to be used?

models/applications.ts
vinayak-trivedi: max it will be 500

models/applications.ts
vinayak-trivedi: so any one who wants to join in RDS needs to fill in a form, which gets stored in applicants collection. and the super user can see all those request, accept or reject it.

models/applications.ts
```diff
@@ -1,6 +1,56 @@
 import { application } from "../types/application";
 const firestore = require("../utils/firestore");
 const ApplicationsModel = firestore.collection("applicants");
+const { DOCUMENT_WRITE_SIZE: FIRESTORE_BATCH_OPERATIONS_LIMIT } = require("../constants/constants");
```
vikhyat187, line 4: Hi @vinayak-trivedi can we use a similar name here `DOCUMENT_WRITE_SIZE`, something related to `FIRESTORE_BATCH_OPERATIONS_LIMIT`  like `FIRESTORE_BATCH_WRITE_LIMIT`

models/applications.ts
vikhyat187: Are we getting all the entries? shouldn't they be on some condition like `where createdAt = null`?

models/applications.ts
vikhyat187: do we have to mention it explicitly?

models/applications.ts
vinayak-trivedi: firebase doesn't allow that, so if I filter the applications with where statement containing createdAt, it will ignore those docs.

models/applications.ts
```diff
@@ -1,6 +1,56 @@
...
+      try {
+        await batch.commit();
+        multipleApplicationUpdateBatch.push(batch);
+      } catch (error) {
+        operationStats.totalFailedApplicationUpdates += applications.length;
+        applications.forEach(({ id }) => operationStats.failedApplicationUpdateIds.push(id));
+      }
+    });
+
+    await Promise.allSettled(multipleApplicationUpdateBatch);
```
vikhyat187, line 47: didn't get this one, we are catching promise in catch block, can you explain this

routes/applications.ts
```diff
@@ -24,5 +24,6 @@ router.patch(
   applicationValidator.validateApplicationUpdateData,
   applications.updateApplication
 );
+router.patch("/batch/update", authenticate, authorizeRoles([SUPERUSER]), applications.batchUpdateApplications);
```
vikhyat187, line 27: Here update is a verb, can we have it like `/batch` its a PATCH operation -> it updates something

models/applications.ts
vinayak-trivedi: it is not inside catch block

routes/applications.ts
vinayak-trivedi: so if I do only batch, it will conflicts with /application_id route, and consider it to be a id

models/applications.ts
vikhyat187: Does that filtering requires an existing index on that columns?

models/applications.ts
vikhyat187: Correct, my bad

routes/applications.ts
vikhyat187: understood

vikhyat187 (approved):

anishpawaskar (approved):

Issue #1707, opened 2023-11-22
Organising Groups functionality 


skv93-coder:
# Organising Groups 
## Overview

In Real Dev Squad, we want to be able to manage groups that users can join. we want to present groups in orderly way Recently used, Recently created and Most popular.

## User stories

- As a user, I should be able to see groups categorised Recently used, Recently created and Most popular and join this groups.

## Acceptance Criteria
- As a user, i should be able to go to group discord page and see which group are recently used and join that group.
- As a user, i should be able to go to group discord page and see which group are recently created and join that group.
- As a user, i should be able to go to group discord page and see which group are Most popular among dev and join that group.

PR #1715, opened 2023-11-23
I1712 migration script to add created at updated at in task doc

prakashchoudhary07:
Date: 23-11-23

Developer Name: @prakashchoudhary07 

----

## Issue Ticket Number:- 
- Closes #1712 

## Description: 
Migration script for adding createdAt and updatedAt fields in task documents

API CALL: 

```
POST: /tasks/migration
body: {
    "action": "ADD",
    "field": "CREATED_AT+UPDATED_AT"
}

```

Is Under the Feature Flag 
- [ ] Yes
- [x] No

Database changes
- [x] Yes
- [ ] No

Breaking changes (If your feature is breaking/missing something please mention pending tickets)
- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [ ] Yes
- [ ] No

### Add relevant Screenshot below ( e.g test coverage etc. )
image

Test coverage
image

PR #1799, opened 2023-12-19
Added last used in field in post group API

skv93-coder:
Date: 19-12-23

Developer Name: @skv93-coder 

----

## Issue Ticket Number:- 
- #1707 

## Description: 
Added lastUsedOn field.

Is Under Feature Flag 
- [ ] Yes
- [x] No

Database changes
- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [ ] Yes
- [x] No

### Add relevant Screenshot below ( e.g test coverage etc. )

image
image


iamitprakash (approved):

ajoykumardas12 (approved):

ajoykumardas12:
Some group data won't have `lastUsedOn` field until used once after this change, right?
If yes, this can have problems when trying to access this property during sorting groups.

skv93-coder:
> Some group data won't have `lastUsedOn` field until used once after this change, right? If yes, this can have problems when trying to access this property during sorting groups.

Yes, you are right that will be handled in the frontend side

ajoykumardas12:
> Yes, you are right that will be handled in the frontend side

You could also run migration script to add this field to existing group data. For example #1715

skv93-coder:
> > Yes, you are right that will be handled in the frontend side
> 
> You could also run migration script to add this field to existing group data. For example #1715

Yes but null check will be needed even then. So adding migration will not solve it. Eg. Newly created groups will not have last joined date. So, migration will not solve null issue.

PR #1827, opened 2024-01-04
Fixes types errors in task requests

Ajeyakrishna-k:
Date: 4 January 2024

Developer Name: @Ajeyakrishna-k 

----

## Description: 
- Fixes errors when running on typescript strict mode

Is Under Feature Flag 
- [x] Yes
- [ ] No

Database changes
- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Is Development Tested?

- [x] Yes
- [ ] No

Tested in staging?

- [ ] Yes
- [x] No

### DIFF
--- a/utils/task-requests.ts
+++ b/utils/task-requests.ts
@@ -42,7 +42,7 @@ const transformTaskRequests = async (taskRequestsList: TaskRequestType[]) => {
   const usersMap = new Map<string, User>(userList.map((data) => [data.id, data]));
   taskRequestsList.forEach((data) => {
     data.users = data.users.map((userData) => {
-      const { username, first_name: firstName, last_name: lastName, picture }: User = usersMap.get(userData.userId);
+      const { username, first_name: firstName, last_name: lastName, picture }: any = usersMap.get(userData.userId);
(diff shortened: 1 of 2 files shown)


utils/task-requests.ts
```diff
@@ -42,7 +42,7 @@ const transformTaskRequests = async (taskRequestsList: TaskRequestType[]) => {
   const usersMap = new Map<string, User>(userList.map((data) => [data.id, data]));
   taskRequestsList.forEach((data) => {
     data.users = data.users.map((userData) => {
-      const { username, first_name: firstName, last_name: lastName, picture }: User = usersMap.get(userData.userId);
+      const { username, first_name: firstName, last_name: lastName, picture }: any = usersMap.get(userData.userId);
```
ardourApeX, line 45: Can't we add type here as well instead of specifying `any`?

sahsisunny: +1

Ajeyakrishna-k: Will have change to `User | undefined`, I think it will be the same as any.  And anyways we are destructuring and collecting only string fields and returning them right away.
Let me know what you think.

ardourApeX (approved):

PR #1852, opened 2024-01-17
FEAT: New Discord invite API  for external service

sahsisunny:
**Description:**
 The purpose of this PR is to facilitate the generation of Discord invite links for external use. The need for such functionality arose from recent scenarios, such as Ankush taking a test to join the React Dev Squad, where a dedicated API endpoint (/invite) was required to dynamically generate and provide Discord invite links.

Test Coverage:
image

**Issue Ticket Number:**
**Parent Issue ticket:** https://github.com/Real-Dev-Squad/website-backend/issues/1853

**Backend changes**
- [x] Yes
- [ ] No

**Frontend Changes**
- [ ] Yes
- [x] No

**Is Under Feature Flag**
- [ ] Yes
- [x] No

**Database changes**
- [x] Yes
- [ ] No

**Breaking changes**
- [ ] Yes
- [x] No

**Deployment notes:**
<Add Deployment Notes Here>

**Tested in development**
- [x] Yes
- [ ] No

**Screenshot**
image
image
image
image


vinit717 (commented):
We already have discord invite API so why did you change everything 
can't we just modify that API only?

sahsisunny:
> We already have discord invite API so why did you change everything can't we just modify that API only?

Not changed i add new API with Service to service authentication its a different

Ajeyakrishna-k:
I had some questions: 
- Who can call this api? 
- Who should the `userId` belong to? 
- More info on invite model. 

Can you please update the documentations for these and also use the appropriate PR description format.

test/unit/utils/discord-actions.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  let fetchStub;
+
+  beforeEach(function () {
+    fetchStub = sinon.stub(global, "fetch");
+  });
+  afterEach(function () {
+    fetchStub.restore();
+  });
+
+  it.skip("should return the invite link", async () => {
```
Ajeyakrishna-k, line 16: Why skip?

sahsisunny: I will fix it

sahsisunny:
> I had some questions:
> 
> * Who can call this api?
>
This API used by external service like in this case Ankush use this API in own backend.
> * Who should the `userId` belong to?
>
userId is encrypted or hash of like twitter id,Facebook id.
> * More info on invite model.
>
Some suggestions please 
> 
> Can you please update the documentations for these and also use the appropriate PR description format.
>
Ok

models/invites.ts
```diff
@@ -0,0 +1,47 @@
+import firestore from "../utils/firestore";
+import { InviteBody, Invite } from "../types/invites";
+const inviteModel = firestore.collection("invite");
+
```
prakashchoudhary07, line 4: Let's not create a new collection for it, We should use existing collection for discord invites

controllers/invites.ts
```diff
@@ -0,0 +1,74 @@
+import { createInviteLinkModel, getInviteLinkModel } from "../models/invites";
+import { InviteResponse, InviteBodyRequest } from "../types/invites";
+import { addLog } from "../models/logs";
+import { generateDiscordInviteLink } from "../utils/discord-actions";
+import { verifyAuthToken } from "../utils/verifyAuthToken";
+
+export const createInviteLink = async (req: InviteBodyRequest, res: InviteResponse) => {
+  try {
+    const { userId, purpose } = req.body;
+    const authHeader = req.headers.authorization;
```
iamitprakash, line 10: can we check null here

controllers/invites.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+  } catch (error) {
+    logger.error("Internal server error", error);
+    return res.boom.internal(error);
+  }
+};
+
+export const getInviteLink = async (req: InviteBodyRequest, res: InviteResponse) => {
+  try {
+    const { userId } = req.params;
```
iamitprakash, line 51: can we check null here

iamitprakash (dismissed):
Conditionally Approved, check with Ankush for Prakash comments

controllers/invites.ts
sahsisunny: Done

models/invites.ts
sahsisunny: Fix

controllers/invites.ts
```diff
@@ -0,0 +1,95 @@
...
+import { verifyAuthToken } from "../utils/verifyAuthToken";
+
+export const createInviteLink = async (req: InviteBodyRequest, res: CustomResponse) => {
+  try {
+    const { userId, purpose } = req.body;
+    const authHeader = req.headers?.authorization;
+    if (!authHeader) {
+      return res.boom.unauthorized();
+    }
+    const token = authHeader.split(" ")[1];
```
Ajeyakrishna-k, line 15: - What should be `authHeader.split(" ")[1];`?
- What if I don't pass space in header eg: `authorization: sdafklhjasdjkfh`;

utils/discord-actions.js
```diff
@@ -26,7 +29,39 @@ const generateDiscordProfileImageUrl = async (discordId) => {
...
+const generateDiscordInviteLink = async () => {
+  try {
+    const channelId = config.get("discordNewComersChannelId");
+    const authToken = jwt.sign({}, RDS_SERVERLESS_PRIVATE_KEY, {
+      algorithm: "RS256",
+      expiresIn: RDS_SERVERLESS_TTL,
+    });
+
+    const inviteOptions = {
+      channelId: channelId,
```
Ajeyakrishna-k, line 41: Just a question:  Isn't there a way where we can specify the times where we can use this link? Do we follow that here? 
Eg: if someone was kicked from the server can they use the same link to join again?

controllers/invites.ts
```diff
@@ -0,0 +1,95 @@
...
+    };
+    const invite = await addInviteToInviteModel(inviteData);
+    if (!invite) {
+      return res.boom.badRequest("Error while adding invite link to database");
+    }
+    const inviteLog = {
+      type: "invite",
+      meta: {
+        action: "create",
+        createdBy: userId,
```
Ajeyakrishna-k, line 44: isn't `createdBy` misleading? 
I can call this api with the appropriate token and pass anyone's userId right?

controllers/invites.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export const getInviteLink = async (req: InviteBodyRequest, res: CustomResponse) => {
+  try {
+    const { userId } = req.params;
+    const authHeader = req.headers?.authorization;
+    if (!authHeader) {
+      return res.boom.unauthorized("Unauthorised");
+    }
+    const token = authHeader.split(" ")[1];
```
Ajeyakrishna-k, line 74: Re: https://github.com/Real-Dev-Squad/website-backend/pull/1852/files#r1458583624

controllers/invites.ts
sahsisunny: If you don't include a space after "Bearer," it will blend with the token, resulting in an invalid token and triggering an "Unauthorized" error.

sahsisunny: I also added this steps in implementation doc

utils/discord-actions.js
sahsisunny: This invitation link is designed for a single user to join, and once it has been used, it becomes invalid.

controllers/invites.ts
sahsisunny: > isn't `createdBy` misleading?
>
Then what? Can you suggest me another field name?

> I can call this api with the appropriate token and pass anyone's userId right?
>
Yes you can call this API with valid token any userId

controllers/invites.ts
Ajeyakrishna-k: Where are we verifying "Bearer"?

controllers/invites.ts
Ajeyakrishna-k: Maybe not have it all as the invite is not created by that userid. Instead its created for that userId.

controllers/invites.ts
sahsisunny: "Bearer" it should be any string not exactly

controllers/invites.ts
sahsisunny: Adding createdBy for logs only

sahsisunny: The userId initiates a call to an external service API, which in turn invokes the RDS API to generate an invitation link. The link is created by the user, and the external service requires user involvement for the call; it does not occur automatically.

controllers/invites.ts
Ajeyakrishna-k: Didn't get you.

controllers/invites.ts
sahsisunny: Then what?

controllers/invites.ts
sahsisunny: What you want?

PR #1865, opened 2024-01-29
feat/issue-1088: isPendingExtensionRequest field added for pending extension requests

palakgupta2712:
Date: 29/01/2024

Developer Name: Palak Gupta (palakgupta2712)

---

## Issue Ticket Number
resolves [#1088](https://github.com/Real-Dev-Squad/website-status/issues/1088)

## Description
Conditionally show the icon that will redirect user to extension-requests/taskId={taskId}, only if there is any pending extension request for that particular task.

### Documentation Updated?

- [x] Yes
- [ ] No
As per current documentation format, so changes were made.

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>
<summary>Scenario 1</summary>
- isPendingExtensionRequest is `false` for that task, 

image

image

</details>

<details>
<summary>Scenario 2</summary>
- isPendingExtensionRequest is `true` for that task,
- On clicking the icon it will redirect to `/extension-requests?order=asc&q=taskId%{taskId}`

image

image</details>

## Test Coverage
<details>

</details>

### DIFF
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -254,13 +254,13 @@ const getSelfTasks = async (req, res) => {
 const getTask = async (req, res) => {
   try {
     const taskId = req.params.id;
-    const { taskData, dependencyDocReference, isPendingExtensionRequest } = await tasks.fetchTask(taskId);
+    const { taskData, dependencyDocReference, isExtensionRequestPending } = await tasks.fetchTask(taskId);
     if (!taskData) {
       return res.boom.notFound("Task not found");
     }
     return res.json({
       message: "task returned successfully",
-      taskData: { ...taskData, dependsOn: dependencyDocReference, isPendingExtensionRequest },
+      taskData: { ...taskData, dependsOn: dependencyDocReference, isExtensionRequestPending },
     });
   } catch (err) {
     return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
--- a/controllers/tasks.js
+++ b/controllers/tasks.js
@@ -254,13 +254,13 @@ const getSelfTasks = async (req, res) => {
 const getTask = async (req, res) => {
   try {
     const taskId = req.params.id;
-    const { taskData, dependencyDocReference } = await tasks.fetchTask(taskId);
+    const { taskData, dependencyDocReference, isPendingExtensionRequest } = await tasks.fetchTask(taskId);
     if (!taskData) {
       return res.boom.notFound("Task not found");
     }
     return res.json({
       message: "task returned successfully",
-      taskData: { ...taskData, dependsOn: dependencyDocReference },
+      taskData: { ...taskData, dependsOn: dependencyDocReference, isPendingExtensionRequest },
     });
   } catch (err) {
     return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
--- a/models/tasks.js
+++ b/models/tasks.js
@@ -283,15 +283,13 @@ const fetchActiveTaskMembers = async () => {
 const fetchTask = async (taskId) => {
   try {
     const task = await tasksModel.doc(taskId).get();
-    const extensionRequestSnapshot = await extensionRequestsModel.where("taskId", "==", taskId).get();
-    const extensionRequestsData = extensionRequestSnapshot.docs.map((doc) => {
-      const extensionRequestId = doc.id;
-      const status = doc.get("status");
-      return { extensionRequestId, status };
-    });
-    const isExtensionRequestPending = Boolean(
-      extensionRequestsData.filter((doc) => doc.status === EXTENSION_REQUEST_STATUS.PENDING).length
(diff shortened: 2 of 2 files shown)


controllers/tasks.js
```diff
@@ -254,13 +254,13 @@ const getSelfTasks = async (req, res) => {
 const getTask = async (req, res) => {
   try {
     const taskId = req.params.id;
-    const { taskData, dependencyDocReference } = await tasks.fetchTask(taskId);
+    const { taskData, dependencyDocReference, isPendingExtensionRequest } = await tasks.fetchTask(taskId);
```
yesyash, line 257: - I think a more suitable term here is `isExtensionRequestPending` as the pending status is a part of the extension request not a by product.

models/tasks.js
```diff
@@ -281,6 +283,14 @@ const fetchActiveTaskMembers = async () => {
 const fetchTask = async (taskId) => {
   try {
     const task = await tasksModel.doc(taskId).get();
+    const extensionRequestSnapshot = await extensionRequestsModel.where("taskId", "==", taskId).get();
+    const extensionRequestsData = extensionRequestSnapshot.docs.map((doc) => {
+      const extensionRequestId = doc.id;
+      return { extensionRequestId, status: doc.get("status") };
```
yesyash, line 289:
```suggestion
      const status = doc.get("status");
      return { extensionRequestId, status };
```

yesyash (changes requested):

Achintya-Chatterjee (changes requested):
@palakgupta2712 can you please add the tests stats

Achintya-Chatterjee (changes requested):
@palakgupta2712 backend test is failing. Kindly look into it.

models/tasks.js
```diff
@@ -281,6 +283,15 @@ const fetchActiveTaskMembers = async () => {
...
     const task = await tasksModel.doc(taskId).get();
+    const extensionRequestSnapshot = await extensionRequestsModel.where("taskId", "==", taskId).get();
+    const extensionRequestsData = extensionRequestSnapshot.docs.map((doc) => {
+      const extensionRequestId = doc.id;
+      const status = doc.get("status");
+      return { extensionRequestId, status};
+    });
+    const isExtensionRequestPending = Boolean(
+      extensionRequestsData.filter((doc) => doc.status === EXTENSION_REQUEST_STATUS.PENDING).length
+    );
```
Ajeyakrishna-k, line 294: Not sure of this approach, can't we just use` /extension-requests` route get the pending extension request? 

In this approach say `/task` is being 5 different places. We will be fetching and sending extension request details even thought its not required.

Ajeyakrishna-k (changes requested):
Please update the documents(API docs) if you plan to change the request/response schema

models/tasks.js
palakgupta2712: @Ajeyakrishna-k Didn't understand your concern here. 
Do you want me to call  /extension-requests when tasks/:id route is hit on ui, then get this value?

Ajeyakrishna-k: > Do you want me to call /extension-requests when tasks/:id route is hit on ui, then get this value?
I would go with this approach. I have mentioned my reasons for the same, feel free to ask others opinion.

If you want to stick to your current approach, you can add filters to query only `PENDING` extension requests and also fetch only the count.
https://firebase.google.com/docs/firestore/query-data/aggregation-queries#use_the_count_aggregation

models/tasks.js
iamitprakash: what if "status" is not available..?

iamitprakash (changes requested):

models/tasks.js
palakgupta2712: done. @Ajeyakrishna-k Can you please review once?

yesyash:
- @palakgupta2712 @Ajeyakrishna-k and had a conversation on the requirement for this change, is there any conclusion on it?

models/tasks.js
palakgupta2712: @iamitprakash 
```const extensionRequestSnapshot = await extensionRequestsModel
      .where("taskId", "==", taskId)
      .where("status", "==", EXTENSION_REQUEST_STATUS.PENDING)
      .count()
      .get();
    const extensionRequestsData = extensionRequestSnapshot.data().count;```
    
    after the suggestion of @Ajeyakrishna-k have updated the logic for getting the extensionRequest. I guess that won;t cause issue now?

palakgupta2712:
not relevant anymore. 
https://github.com/Real-Dev-Squad/website-status/issues/1088 done with this PR.

PR #1843, opened 2024-01-10
Added new route for create ooo request

sahsisunny:
**Description:**
This is the last PR for creating an OOO  status request. in this pR added a new generic rote named `/requests` and used OOO requests middleware, model, and controller to create OOO requests

**Test Coverage:**
| File           | Coverage | Statements | Branches | Functions | Lines |
| -------------- | -------- | ---------- | -------- | --------- | ----- |
| requests.ts    |   88.88% |         50 |      100 |      87.5 |    14 |

**Screenshots:**
image
image
image

PR #1870, opened 2024-01-31
Added controller for request and ooo requests API

sahsisunny:
**Description:**
In this pull requests added controller function for OOO requests with requests API.

**TEST:** Integration test is added in #https://github.com/Real-Dev-Squad/website-backend/pull/1843 PR


controllers/oooRequests.ts
```diff
@@ -0,0 +1,52 @@
+import { createOooRequest } from "../models/oooRequests";
+import { OooStatusRequestResponse, OooStatusRequestRequest } from "../types/oooRequest";
+import { addLog } from "../models/logs";
+import { CREATE, ERRORS, ERROR_WHILE_CREATING_OOO_REQUEST, OOO_LOG_TYPE, OOO_STATUS_REQUEST_CREATED_SUCCESSFULLY } from "../constants/oooRequest";
+
+export const createOooRequestController = async (req: OooStatusRequestRequest, res: OooStatusRequestResponse) => {
+    const oooRequestBody = req.body;
+    const userId = req.userData.id;
```
iamitprakash, line 8: what if req is null or userData is null

controllers/oooRequests.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    try {
+        const oooRequestDoc = await createOooRequest({userId, ...oooRequestBody});
```
iamitprakash, line 11: oooRequestDoc please use better name

controllers/oooRequests.ts
```diff
@@ -0,0 +1,52 @@
...
+        if ('error' in oooRequestDoc) {
+            const oooRequestLog = {
+                type: OOO_LOG_TYPE.OOO_REQUEST_BLOCKED,
+                meta: {
+                    action: ERRORS,
+                    createdBy: userId,
+                    createdAt: Date.now(),
```
iamitprakash, line 18: don't we need to convert date into specific timezone

controllers/requests.ts
```diff
@@ -0,0 +1,16 @@
+import { OooStatusRequestResponse, OooStatusRequestRequest } from "../types/oooRequest";
+import { REQUEST_TYPE } from "../constants/request";
+import { createOooRequestController } from "./oooRequests";
+
+export const createRequestController = async (
+  req: OooStatusRequestRequest,
```
iamitprakash, line 6: please change this

iamitprakash (changes requested):

controllers/oooRequests.ts
sahsisunny: Done , added null check condition

controllers/oooRequests.ts
sahsisunny: Done

controllers/requests.ts
sahsisunny: Done

iamitprakash (approved):

Issue #1929, opened 2024-02-27
Fix `updatedAt` in user documents in firestore


MehulKChaudhari:
### Issue Description
Now we don't update `updatedAt` field after updating the user details in DB.

### Expected Behavior
`updatedAt` should be updated with latest timestamp every time user details are updated in DB. 

### Current Behavior
`updatedAt` is not getting updated after user details are updated in DB. both the `createdAt` and `updatedAt` fields have same value. 

### Screenshots
<img width="422" alt="image" src="https://github.com/Real-Dev-Squad/website-backend/assets/55375534/3d8d259f-162a-47b5-9583-757724d53e2a">

### Reproducibility

 - [] This issue is reproducible
 - [ ] This issue is not reproducible

### Steps to Reproduce
1. hit API `https://api.realdevsquad.com/users` and you can see the timestamp not being updated after user details are updated. 

### Severity/Priority
 - [] Critical
 - [ ] High
 - [ ] Medium
 - [ ] Low

### Checklist

- [ ] I have read and followed the project's code of conduct.
- [ ] I have searched for similar issues before creating this one.
- [ ] I have provided all the necessary information to understand and reproduce the issue.
- [ ] I am willing to contribute to the resolution of this issue.

PR #1980, opened 2024-03-23
update updated_at in user documents whenever user data is updated

MehulKChaudhari:
Date: 23/03/2024

Developer Name: Mehul Chaudhari

---

## Issue Ticket Number

#1929 
## Description

Some APIs which update the user documents now don't update the `updated_at` field in the user document which says when the document was updated latest. 

Not adding or updating tests, because we are just updating the `updated_at` field while updating the `user` document, and most of the API's do not return the `updated_at` field in the response.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [x] Yes
- [ ] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [ ] Yes
- [x] No

## Test Coverage
image

</details>

### DIFF
--- a/controllers/auth.js
+++ b/controllers/auth.js
@@ -80,7 +80,7 @@ const githubAuthCallback = (req, res, next) => {
         github_created_at: Number(new Date(user._json.created_at).getTime()),
         github_user_id: user.id,
         created_at: Date.now(),
-        updated_at: Date.now(),
+        updated_at: null,
       };
 
       const { userId, incompleteUserDetails, role } = await users.addOrUpdate(userData);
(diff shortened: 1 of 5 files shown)


lakshayman (approved):

shreya-mishra (approved):

prakashchoudhary07:
Ideally, we should have a single service that only updates the user data, and in that service updated_at is added

prakashchoudhary07:
Please create a following issues to refactor that

MehulKChaudhari:
> Ideally, we should have a single service that only updates the user data, and in that service updated_at is added

Yes Right, But the old code was such that, I had to do it this way. we have a model function called addOrUpdate but developers have not used it properly.

PR #2080, opened 2024-08-19
Fix/allow users to view their own application status data

Achintya-Chatterjee:
Date: `19th August, 2024`

Developer Name: @Achintya-Chatterjee 

---

## Description
This PR updates the backend authorization logic for accessing user intro data
-  Users can now view their own application status without requiring superuser privileges.
- Superusers can still access the intro data of any user.
- Proper handling of unauthorized access with a 403 status code.
- Maintains the existing 404 response if no data is found.

### Documentation Updated?

- [ ] Yes
- [ ] No

### Under Feature Flag

- [ ] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [ ] No

### Breaking Changes

- [ ] Yes
- [ ] No

### Development Tested?

- [ ] Yes
- [ ] No

## Screenshots
<details>

</details>

## Test Coverage
<details>

</details>

### DIFF
--- a/routes/users.js
+++ b/routes/users.js
@@ -33,7 +33,7 @@ router.patch(
   users.updateDiscordUserNickname
 );
 router.get("/:username", users.getUser);
-router.get("/:userId/intro", authenticate, authorizeRoles([SUPERUSER]), users.getUserIntro);
+router.get("/:userId/intro", authenticate, users.getUserIntro);
 router.put("/self/intro", authenticate, userValidator.validateJoinData, users.addUserIntro);
 router.get("/:id/skills", users.getUserSkills);
 router.get("/:id/badges", getUserBadges);
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -720,7 +720,7 @@ const addUserIntro = async (req, res) => {
   }
 };
 
-const getUserIntro = async (req, res) => {
+/* const getUserIntro = async (req, res) => {
   try {
     const data = await userQuery.getJoinData(req.params.userId);
     if (data.length) {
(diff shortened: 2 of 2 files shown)


routes/users.js
```diff
@@ -33,7 +33,7 @@ router.patch(
   users.updateDiscordUserNickname
 );
 router.get("/:username", users.getUser);
-router.get("/:userId/intro", authenticate, authorizeRoles([SUPERUSER]), users.getUserIntro);
```
vinit717, line 36: Why this? 

As we already have  application route why you can't fetch with it

vinit717 (changes requested):

routes/users.js
Achintya-Chatterjee: Have to change many code in website-www, to use this application route. Also user's should see their application status in the `/intro` route only.

vinit717: But this is not the good solution 
And ask with Ankush or Tejas is your PR is going to merge or my

vinit717: And also in /intro route we are using /application API

Achintya-Chatterjee: Only superusers would be able to accept new user request, no one else, also if you're not the applicants you will always get `you are not authorised `

vinit717: But there is already route present for view the application then what is the need of this new route?

Achintya-Chatterjee: I don't build this from the ground up, it was there, just did some modifications, also previously told you that if I wanted to use the application route , will have to change many code.

Issue #2110, opened 2024-08-30
[feat] Allow super user to toggle their privileges


shubhdevelop:
### Issue Description
Super user should be able to toggle between their super_user privilege, and the effect should be seen on the UI as well the routes he/she can hit with their current level of privilege

### Expected Behavior

Toggling between super_user privilege from the frontend

### Current Behavior

Nothing 

### Reproducibility

 - [ ] This issue is reproducible
 - [x] This issue is not reproducible

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [ ] Medium
 - [x] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

PR #2111, opened 2024-08-30
[feat] Super User Privilege Toggle

shubhdevelop:
Date: 30 Aug 2024

Developer Name: Shubham Raj

---

## Issue Ticket Number
 
#2110

## Description

This Solution involves the mutating the userData Object returned from the database before adding it  to the response object, which then is used in authenticateRoles middleware for authentication of roles.

Added two routes, one that will help us get the current value of `SuperUserAccess` as the jwt token in user browser are encrypted

2nd route's purpose would be to send an updated jwt token, with the `SuperUserAccess` property in addition with the id of the user.

In the authenticateUser middleware if the value of the `SuperUserAccess === false`, we’ll remove the super_user property or either set it to false, from the userData Object this should make the authorizeRoles middleware return an unauthorised user response back to the user.

If the cookie’s value is either true or undefined(not present in the request), we shouldn’t  be tampering with the user_object and add the returned `userData` object as it is to the Response object, this should make the authorizeRoles middleware to authorise if the roles exist and call to `next()`.

Depending up the value of `SuperUserAccess` we also modify the user object in the `/users/self` and the `super_user` value accordingly before sending it to the user.

### Documentation Updated?

- [ ] Yes
- [x] No

Once this PR gets merged I'll update the the api-contract documentation as well

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screens Capture
<details>
<summary>Development Testing</summary>

`/users/:userId/intro is an routes accessible only when super_user role is present on the user

[hitting restricted routes](https://github.com/user-attachments/assets/0ba15f6d-707d-4d22-a12c-807ff769801a)

[condtional rendering](https://github.com/user-attachments/assets/0ae98120-e452-4a62-8adf-2c90c1854357)

</details>

## Test Coverage
<details>

Test coverage for the changes in /users/self route
image

Test coverage for 2 routes that were added  
image

</details>

## Additional Notes

For more Detail one can refer to this design doc
https://docs.google.com/document/d/17khsviClkNVNYH4Xy5Pkk98Xvrid1jH1jL9QtfGp7aw/edit?usp=sharing

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -414,7 +414,7 @@ const updateSelf = async (req, res) => {
       }
     }
 
-    if (userRoles.in_discord) {
+    if (userRoles.in_discord && !user.incompleteUserDetails) {
       const membersInDiscord = await getDiscordMembers();
       const discordMember = membersInDiscord.find((member) => member.user.id === discordId);
       if (discordMember) {
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 - [About the Project](#about-the-project)
 - [Running the Project](#running-the-project)
 - [Prerequisites](#prerequisites)
-- [API Documentation](#api-documentation)
+- [API Documentation](https://github.com/Real-Dev-Squad/website-api-contracts/)
 - [CONTRIBUTING](CONTRIBUTING.md)
 
 ## About the Project
--- a/controllers/applications.ts
+++ b/controllers/applications.ts
@@ -7,16 +7,6 @@ const { API_RESPONSE_MESSAGES } = require("../constants/application");
 const { getUserApplicationObject } = require("../utils/application");
 const admin = require("firebase-admin");
 
-const batchUpdateApplications = async (req: CustomRequest, res: CustomResponse): Promise<any> => {
-  try {
-    const updateStats = await ApplicationModel.batchUpdateApplications();
-    return res.json(updateStats);
-  } catch (err) {
-    logger.error(`Error in batch updating application: ${err}`);
-    return res.boom.badImplementation(INTERNAL_SERVER_ERROR);
-  }
-};
-
 const getAllOrUserApplication = async (req: CustomRequest, res: CustomResponse): Promise<any> => {
   try {
     const { userId, status, next, size } = req.query;
@@ -154,5 +144,4 @@ module.exports = {
   addApplication,
   updateApplication,
   getApplicationById,
-  batchUpdateApplications,
 };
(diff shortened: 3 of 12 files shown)


middlewares/authenticate.js
```diff
@@ -51,12 +50,24 @@ module.exports = async (req, res, next) => {
...
+      userData.superUserAccess = false;
+    } else if (superUserAccess === true || superUserAccess === undefined) {
+      userData = userDoc.user;
+      if (superUserAccess === true) {
+        userData.superUserAccess = true;
+      } else {
+        userData.superUserAccess = undefined;
+      }
+    }
+    req.userData = userData;
```
vinit717, line 70: Middleware tests are missing

routes/users.js
```diff
@@ -19,6 +19,8 @@ router.get("/userId/:userId", users.getUserById);
 router.patch("/self", authenticate, userValidator.updateUser, users.updateSelf);
 router.get("/", userValidator.getUsers, users.getUsers);
 router.get("/self", authenticate, users.getSelfDetails);
+router.get("/set-super-user-access", authenticate, users.setSuperUserAccessLimiter);
```
vinit717, line 22: Can you change the route name to be more descriptive? 
as this is a GET request and route name is kind on weird

controllers/users.js
```diff
@@ -827,6 +831,53 @@ const setInDiscordScript = async (req, res) => {
...
+      return res.json({ message: "Super User Privilege Access Set!", currentAccess: value });
+    } else {
+      return res.boom.badRequest("Wrong value in query param, value can be either true or false");
+    }
+  } catch (error) {
+    logger.error(`Error while Setting Super Privilege Access Limiter: ${error}`);
+    return res.boom.badImplementation({ message: INTERNAL_SERVER_ERROR });
+  }
+};
+
```
vinit717, line 880: This function become complex and not readable so fix this

vinit717 (changes requested):

prakashchoudhary07:
We need to discuss the approach to it. I am not in favor of this approach. Let's discuss about it with the team before processing please

prakashchoudhary07:
Also, let's not restrict ourselves to superuser privileges. Lets take a step back and discuss about it

shubhdevelop:
> We need to discuss the approach to it. I am not in favor of this approach. Let's discuss about it with the team before processing please

Sure, whenever you free to discuss

PR #2132, opened 2024-09-09
[Feat] Extends Toggle privilege feature allowing other roles to be Toggled off

shubhdevelop:
Date: 09-09-2024

Developer Name: Shubham Raj

---

## Issue Ticket Number
#2110

## Description
The PR add a new property `disabled_roles` in the users object with an array of string value; these string value will an enum value for now  but the list can be extended to more roles as well. 
For now these value are these, which are basically the the property name of the roles
`[ “super_user”, “members”];`

An Empty Array `[ ]` : means that no roles have been toggled off / privileges are downgraded.

`[‘super_user’]`: means that the `super_user` role has been restricted.

`[ “super_user”, “members”]`: means that both `super_user` as well as `member` roles have been restricted.

In the Data Access Layer we mutate the user Object, to be very specific about the function, in the `fetchUser` function, this way all the instances where we would be calling/fetching users we will get the same user object with the updated roles.

If the property name is found in the `disabled_roles` array, we set the value of those properties in the userData to false.

If the `disabled_roles` array is empty, we do nothing, users can use the services without any restriction on the privileges.
 
For adding this property I'm using the `PATCH \users\self`

If we go with this solution, the only place we need to put this function of setting user roles to false, based on the `disabled_roles` property, inside the data Access Layer services only. 

We won’t need to modify any other routes.

### Documentation Updated?

- [x] Yes
- [] No

Update PR raised

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [x] Yes
- [ ] No

Added an another attribute in the user Object in database
### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>
<summary>Screen Cast</summary>

[Screencast from 10-09-24 12:31:35 AM IST.webm](https://github.com/user-attachments/assets/d0ff1906-42c9-4ffe-89e5-10a1534a74bc)

</details>

## Test Coverage
<details>
<summary>Unit Test Coverage</summary>

image

</details>
<details>
<summary>Integration Test Coverage</summary>

image

</details>

## Additional Notes
For additional detailed explanation refer this [**design doc**](https://docs.google.com/document/d/17khsviClkNVNYH4Xy5Pkk98Xvrid1jH1jL9QtfGp7aw/edit?usp=sharing) 

flow of the updateSelf()
image

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -411,20 +411,25 @@ const updateSelf = async (req, res) => {
     }
 
     if (req.body.disabled_roles) {
-      let roles = [...req.body.disabled_roles];
+      let rolesToDisable;
+      const data = req.body.disabled_roles;
+
       if (user.disabled_roles !== undefined) {
-        roles = [...user.disabled_roles];
-        req.body.disabled_roles.forEach((role) => {
-          if (user.disabled_roles.includes(role)) {
-            const index = roles.indexOf(role);
-            roles.splice(index, 1);
+        rolesToDisable = user.disabled_roles;
+
+        data.forEach((role) => {
+          const roleIndex = rolesToDisable.indexOf(role);
+          if (roleIndex !== -1) {
+            rolesToDisable.splice(roleIndex, 1);
           } else {
-            roles.push(role);
+            rolesToDisable.push(role);
           }
         });
+      } else {
+        rolesToDisable = data;
       }
 
-      const updatedUser = await userQuery.addOrUpdate({ disabled_roles: roles }, userId);
+      const updatedUser = await userQuery.addOrUpdate({ disabled_roles: rolesToDisable }, userId);
       if (updatedUser) {
         return res.status(200).send({ message: "Privilege modified successfully!" });
       }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -410,10 +410,10 @@ const updateSelf = async (req, res) => {
       }
     }
 
-    if (req.body.revoked_roles) {
+    if (req.body.disabled_roles) {
       // filter roles which are 'true'
-      const revokableRoles = req.body.revoked_roles.filter((role) => user.roles[`${role}`] === true);
-      const updatedUser = await userQuery.addOrUpdate({ revoked_roles: revokableRoles }, userId);
+      const toggleableRoles = req.body.disabled_roles.filter((role) => user.roles[`${role}`] === true);
+      const updatedUser = await userQuery.addOrUpdate({ disabled_roles: toggleableRoles }, userId);
       if (updatedUser) {
         return res.status(200).send({ message: "Privilege modified successfully!" });
       }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -415,7 +415,7 @@ const updateSelf = async (req, res) => {
       const revokableRoles = req.body.revoked_roles.filter((role) => user.roles[`${role}`] === true);
       const updatedUser = await userQuery.addOrUpdate({ revoked_roles: revokableRoles }, userId);
       if (updatedUser) {
-        return res.status(200).send({ message: "privilege modified successfully!", revoked_roles: revokableRoles });
+        return res.status(200).send({ message: "Privilege modified successfully!" });
       }
     }
 
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -411,9 +411,20 @@ const updateSelf = async (req, res) => {
     }
 
     if (req.body.disabled_roles) {
-      // filter roles which are 'true'
-      const toggleableRoles = req.body.disabled_roles.filter((role) => user.roles[`${role}`] === true);
-      const updatedUser = await userQuery.addOrUpdate({ disabled_roles: toggleableRoles }, userId);
+      let roles = [...req.body.disabled_roles];
+      if (user.disabled_roles !== undefined) {
+        roles = [...user.disabled_roles];
+        req.body.disabled_roles.forEach((role) => {
+          if (user.disabled_roles.includes(role)) {
+            const index = roles.indexOf(role);
+            roles.splice(index, 1);
+          } else {
+            roles.push(role);
+          }
+        });
+      }
+
+      const updatedUser = await userQuery.addOrUpdate({ disabled_roles: roles }, userId);
       if (updatedUser) {
         return res.status(200).send({ message: "Privilege modified successfully!" });
       }
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -396,6 +396,7 @@ const updateSelf = async (req, res) => {
   try {
     const { id: userId, roles: userRoles, discordId } = req.userData;
     const { user } = await dataAccess.retrieveUsers({ id: userId });
+    let rolesToDisable = [];
 
     if (req.body.username) {
       if (!user.incompleteUserDetails) {
@@ -410,10 +411,8 @@ const updateSelf = async (req, res) => {
       }
     }
 
-    if (req.body.disabled_roles) {
-      let rolesToDisable;
-      const data = req.body.disabled_roles;
-
+    if (req.body.disabledRoles) {
+      const data = req.body.disabledRoles;
       if (user.disabled_roles !== undefined) {
         rolesToDisable = user.disabled_roles;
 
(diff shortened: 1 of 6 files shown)


controllers/users.js
```diff
@@ -410,6 +410,26 @@ const updateSelf = async (req, res) => {
...
+          }
+        });
+      }
+
+      const updatedUser = await userQuery.addOrUpdate({ disabled_roles: roles }, userId);
+      if (updatedUser) {
+        return res.status(200).send({ message: "Privilege modified successfully!" });
+      }
+    }
+
```
vinit717, line 432: Can you please refactor this code

test/unit/models/users.test.js
```diff
@@ -488,4 +488,53 @@ describe("users", function () {
...
+      const memberUser = { ...userData()[6], disabled_roles: ["member"] };
+      const userId = await addUser(memberUser);
+
+      const userDoc = await users.fetchUser({ userId });
+      expect(userDoc.user.disabled_roles.length).to.be.equal(1);
+      expect(userDoc.user.roles.member).to.be.equal(false);
+    });
+
+    it("should fetch users with modified roles : super_user & member", async function () {
+      // super_user & member
```
vinit717, line 525: Do we need this comment as test message is already declarative

shubhdevelop: sure i'll fix this

controllers/users.js
shubhdevelop: sure

controllers/users.js
```diff
@@ -410,6 +410,31 @@ const updateSelf = async (req, res) => {
...
+          }
+        });
+      } else {
+        rolesToDisable = data;
+      }
+
+      const updatedUser = await userQuery.addOrUpdate({ disabled_roles: rolesToDisable }, userId);
+      if (updatedUser) {
+        return res.status(200).send({ message: "Privilege modified successfully!" });
+      }
```
prakashchoudhary07, line 435: Why is this being done here?

prakashchoudhary07, line 435: Also does this API specifically modifies privilege?

shubhdevelop: No, it's meant for other purposes as well, As suggested I tried to reuse the `Patch /users/self` route

shubhdevelop: Can you please clarify a bit more

prakashchoudhary07: What was this piece of code previously doing? and what is the new feature that you are adding?

prakashchoudhary07: So how is the role being updated like and user data being updated?
Will the roles be only edited in single call? or can roles be edited with other user data?

Are you not considering disabled_roles as part of user data. 
will you update only disabled_roles ?

test/unit/middlewares/user-validator.test.js
```diff
@@ -308,6 +308,56 @@ describe("Middleware | Validators | User", function () {
     });
   });
 
+  describe("Test the update Self Validator for disabled_roles property", function () {
+    it("Allows the request to pass with disabled_roles property", async function () {
+      const req = {
+        body: {
+          disabled_roles: [],
+        },
```
prakashchoudhary07, line 316: What happens if some other data is being sent with disabled_roles in backend?

controllers/users.js
```diff
@@ -410,6 +410,31 @@ const updateSelf = async (req, res) => {
       }
     }
 
+    if (req.body.disabled_roles) {
+      let rolesToDisable;
+      const data = req.body.disabled_roles;
```
prakashchoudhary07, line 415: Please follow camelCasing for naming in backend?

the DB field can have snake case

controllers/users.js
shubhdevelop: So this code was basically used to data to user Object they have been registered with github and have an id associated with their object. moreover this controller was also used to update the user profile as well but now we're not allow user to update that, for that we now require users to setup profile services

In the new feature  we are adding function which set up a value in the userObject `disabled_roles`, ["super_user", "member"], both or just one, we also handle the removal and addition of these roles in the array in this code block only

test/unit/middlewares/user-validator.test.js
shubhdevelop: other data is being ignored as inside the conditional
```
if(req.body.disabled_roles){
......
}
```
we only giving `{disbaled_roles:req.body.disabled_roles}` in the updateFunction()

I have tested it even if we sent the other properties with it ignores those property, and if the properties are 
- username,
image

- roles
image
not just the super_user role but all the roles are not allowed to be updated

which are handle in the above if block, it also return error of 400 series

test/integration/users.test.js
```diff
@@ -487,6 +487,100 @@ describe("Users", function () {
...
+
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.an("object");
+          expect(res.body).to.eql({
+            message: "Privilege modified successfully!",
+          });
+
+          return done();
+        });
+    });
```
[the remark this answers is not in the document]
shubhdevelop: I'll make sure to add necessary assertion

controllers/users.js
iamitprakash, line 415: what if req is null, empty or undefined

iamitprakash (changes requested):

controllers/users.js
shubhdevelop: > So how is the role being updated like and user data being updated?

we have a `query.addOrupdate()` takes in userId and properties to add in our case `{disabled_roles: [] }`

> will you update only disabled_roles ?

Yes, the effect of toggle of the super_user privileges happens somewhere else, in the `fetchUser()` function

> Are you not considering disabled_roles as part of user data.

No, I'm considering disabled_roles as the part of userData. addOrUpdate adds this disable_roles property to the userObject

controllers/users.js
shubhdevelop: req object won't be empty or null, if that the case validators will catch that

models/users.js
```diff
@@ -317,6 +317,16 @@ const fetchUser = async ({ userId = null, username = null, githubUsername = null
         userData = doc.data();
       });
     }
+
+    if (userData && userData.disabled_roles !== undefined) {
+      const updatedRoles = { ...userData.roles };
```
prakashchoudhary07, line 322: So here you are spreading a object inside an object, is it not negating it self?

will it not be the same as `updatedRoles = userData.roles`?

prakashchoudhary07:
@shubhdevelop Please update the API contract  and data models for your change please
Ref: https://github.com/Real-Dev-Squad/website-data-models https://github.com/Real-Dev-Squad/website-api-contracts/

test/unit/middlewares/user-validator.test.js
```diff
@@ -308,6 +308,56 @@ describe("Middleware | Validators | User", function () {
...
+    it("Allows the request to pass with disabled_roles property with roles `super_user` & `member'", async function () {
+      const req = {
+        body: {
+          disabledRoles: ["super_user", "member"],
+        },
+      };
+      const res = {};
+      const nextSpy = sinon.spy();
+      await updateUser(req, res, nextSpy);
+      expect(nextSpy.notCalled).to.be.equal(false);
```
prakashchoudhary07, line 333: Please update it to the callCount, because here expecting a negative test case is confusing

controllers/users.js
```diff
@@ -410,13 +411,41 @@ const updateSelf = async (req, res) => {
...
+            const updatedUser = await userQuery.addOrUpdate({ disabled_roles: rolesToDisable }, userId);
+            if (updatedUser) {
+              return res
+                .status(200)
+                .send({ message: "Privilege modified successfully!", disabled_roles: rolesToDisable });
+            }
+          }
+          return res.boom.forbidden(
+            "Developers can only update disabled_roles. Use profile service for updating other attributes."
+          );
```
prakashchoudhary07, line 448: Please create a new Isseus to fix the update flow for users API, and paste the link here

models/users.js
shubhdevelop: Wanted to have a deep copy of the `userData.roles` object and mutate it independently without mending with the roles, but yes we can do simple assignment as well.

controllers/users.js
shubhdevelop: https://github.com/Real-Dev-Squad/website-backend/issues/2155

shubhdevelop:
> @shubhdevelop Please update the API contract and data models for your change please. Update the documentation for it Ref: https://github.com/Real-Dev-Squad/website-data-models https://github.com/Real-Dev-Squad/website-api-contracts/

#### Documentation update
https://github.com/Real-Dev-Squad/website-api-contracts/pull/187

prakashchoudhary07 (approved):

iamitprakash (approved):

PR #2134, opened 2024-09-10
fix: username generation for new user and modify format format func for migration

### LINKED ISSUE #2133: Fix Username Generation for New Users During Signup Based on Migration Logic
### Description:
The current username generation process during user signup does not comply with the new migration logic introduced for existing users. This issue aims to update the username generation for new users to ensure consistency with the migration logic, which enforces a specific format, length constraint, and conflict resolution for duplicate usernames.

### Acceptance Criteria:
**Username Format:**

The username should follow the format: {firstname}-{lastname}-{n}.
First name: The first part of the first name (ignore middle names).
Last name: Truncated as needed to fit within the 32-character limit.
Suffix: A numeric suffix starting from -1 for the first occurrence of the name and incremented for subsequent users with the same name combination.

**Character Limit:**

The total username length should not exceed 32 characters, including the first name, last name, hyphens, and suffix.
If the total length exceeds the limit, only the last name should be truncated to fit within the character limit.

**Edge Case Handling:**

Middle Names: If a user’s first name includes a middle name, only the first part (e.g., Vinit from Vinit Kumar) should be used.
Whitespace: Ensure that leading and trailing whitespaces are removed from both the first name and last name during processing.
Conflict Resolution: When a username conflict arises (i.e., if a username already exists), append a numeric suffix starting with -1 and increment for each subsequent user with the same name combination.
Example: If the first user with the name John Doe has the username john-doe-1, the next user with the same name should be assigned john-doe-2.

vinit717:
## Issue Ticket Number
- closes #2133

## Description

In this PR username generation functionality is modified and make format func is generic so It can be used in migration
 https://docs.google.com/document/d/1qO9QqBR6cgmFzC4UEnehSVugAOThtY0u90BsHFqrcUw/edit?usp=sharing

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
image

## Test Coverage
image
image
image

### DIFF
--- a/middlewares/validators/user.js
+++ b/middlewares/validators/user.js
@@ -315,8 +315,18 @@ const validateGenerateUsernameQuery = async (req, res, next) => {
     .object()
     .strict()
     .keys({
-      firstname: joi.string().min(1).required(),
-      lastname: joi.string().min(1).required(),
+      firstname: joi
+        .string()
+        .trim()
+        .pattern(/^[a-zA-Z]+$/)
+        .min(1)
+        .required(),
+      lastname: joi
+        .string()
+        .trim()
+        .pattern(/^[a-zA-Z]+$/)
+        .min(1)
+        .required(),
       dev: joi.string().valid("true").optional(),
     });
 
@@ -328,6 +338,7 @@ const validateGenerateUsernameQuery = async (req, res, next) => {
     res.boom.badRequest("Invalid Query Parameters Passed");
   }
 };
+
 const migrationsValidator = async (req, res, next) => {
   const { action, page, size } = req.query;
   const schema = joi
--- a/models/users.js
+++ b/models/users.js
@@ -793,12 +793,15 @@ const getUsersByRole = async (role) => {
 
 const generateUniqueUsername = async (firstName, lastName) => {
   try {
-    const snapshot = await userModel.where("first_name", "==", firstName).where("last_name", "==", lastName).get();
+    const snapshot = await userModel
+      .where("first_name", "==", firstName)
+      .where("last_name", "==", lastName)
+      .count()
+      .get();
 
-    const existingUserCount = snapshot.size;
+    const existingUserCount = snapshot.data().count || 0;
 
     const suffix = existingUserCount + 1;
-
     const finalUsername = formatUsername(firstName, lastName, suffix);
 
     return finalUsername;
--- a/models/users.js
+++ b/models/users.js
@@ -793,23 +793,16 @@ const getUsersByRole = async (role) => {
 
 const generateUniqueUsername = async (firstName, lastName) => {
   try {
-    let suffix = 1;
-    let finalUsername;
-    let isUnique = false;
+    const snapshot = await userModel
+      .where("first_name", "==", firstName)
+      .where("last_name", "==", lastName)
+      .count()
+      .get();
 
-    while (!isUnique) {
-      finalUsername = formatUsername(firstName, lastName, suffix);
+    const existingUserCount = snapshot.data().count || 0;
 
-      const snapshot = await userModel.where("username", "==", finalUsername).count().get();
-
-      const existingCount = snapshot.data().count || 0;
-
-      if (existingCount === 0) {
-        isUnique = true;
-      } else {
-        suffix++;
-      }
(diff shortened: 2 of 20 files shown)


models/users.js
```diff
@@ -790,21 +791,17 @@ const getUsersByRole = async (role) => {
...
-    const initialUsername = `${firstname}-${lastname}`;
-    if (existingUserCount === 0) {
-      return initialUsername;
-    } else {
-      const uniqueUsername = `${initialUsername}-${existingUserCount + 1}`;
-      return uniqueUsername;
-    }
+    const snapshot = await userModel.where("first_name", "==", firstName).where("last_name", "==", lastName).get();
+
+    const existingUserCount = snapshot.size;
```
prakashchoudhary07, line 798: I think there is something in firestore that directly returns count have a look once

vinit717:
#2147

test/unit/models/users.test.js
```diff
@@ -425,6 +425,26 @@ describe("users", function () {
...
+
+    it("should generate a unique username when no existing users are present", async function () {
+      const firstName = "john";
+      const lastName = "doe";
+
+      const whereStub = sinon.stub(userModel, "where").returnsThis();
+      const countStub = sinon.stub().returnsThis();
+      const getStub = sinon.stub().resolves({
+        data: () => ({ count: 0 }),
+      });
```
prakashchoudhary07, line 437: Why are we mocking this?
We are using firestore emulator right

utils/username.ts
```diff
@@ -0,0 +1,29 @@
...
+    const validFirstName = /^[a-zA-Z]+$/.test(firstNamePart)
+        ? firstNamePart.toLowerCase()
+        : "null";
+
+    const lastNameParts = trimmedLastName.split(/[\s-]/);
+    const lastNamePart = lastNameParts[lastNameParts.length - 1];
+    const validLastName = /^[a-zA-Z]+$/.test(lastNamePart)
+        ? lastNamePart.toLowerCase()
+        : "null";
+
```
prakashchoudhary07, line 17: What are the regex for?

vinit717: To validate the characters
for more, you can look format username tests

test/unit/models/users.test.js
vinit717: fixed

models/users.js
vinit717: fixed

models/users.js
```diff
@@ -790,21 +791,20 @@ const getUsersByRole = async (role) => {
   }
 };
 
-const generateUniqueUsername = async (firstname, lastname) => {
+const generateUniqueUsername = async (firstName, lastName) => {
```
iamitprakash, line 794:
```
const generateUniqueUsername = async (firstName, lastName) => {
  try {
    let suffix = 1;
    let finalUsername;
    let isUnique = false;

    while (!isUnique) {
      finalUsername = formatUsername(firstName, lastName, suffix);
      
      const snapshot = await userModel
        .where("username", "==", finalUsername)
        .count()
        .get();
      
      const existingCount = snapshot.data().count || 0;
      
      if (existingCount === 0) {
        isUnique = true;
      } else {
        suffix++;
      }
    }

    return finalUsername;
  } catch (err) {
    logger.error(`Error while generating unique username: ${err.message}`);
    throw err;
  }
};
```

prakashchoudhary07, line 794: Please move this out of model

vinit717: Done

middlewares/validators/user.js
```diff
@@ -328,6 +338,7 @@ const validateGenerateUsernameQuery = async (req, res, next) => {
     res.boom.badRequest("Invalid Query Parameters Passed");
   }
 };
```
iamitprakash, line 340: please remove white space

prakashchoudhary07 (approved):
LGTM

iamitprakash (approved):

Issue #2195, opened 2024-10-04
Allow Users to Edit Extension Requests Before Approval


ahsrah7:
### Issue Description

- Currently, once a user raises an extension request, they are unable to edit it, regardless of its status. This issue creates problems when users need to update or correct information before it is  approved.
### Expected Behavior

- Users should be able to edit extension requests if they have not yet been approved.
- Once an extension request is approved, it should become locked for editing.
### Current Behavior

- Users are unable to modify their request details, even if the request has not yet been approved.
### Reproducibility

 - [x] This issue is reproducible
 - [ ] This issue is not reproducible

### Steps to Reproduce
1. Raise an extension request as a user.
2. Try to edit the request before it is approved.
3. Observe that editing is allowed on UI but once save is hit the content is still not updated even though the request is still pending approval.

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [ ] Medium
 - [x] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

Mangesh1511:
if extension requests are made by the logged in user, for the user can we store the extension requests by the user as pending extension requests , and user can edit the pending extension requests and once approved we will update the extension requests.
But I didn't got what are the extension request you are referring to , I assumed this is any feature of the website(actually new, so didn't saw the actual feature), and just proposed a solution,

PR #2215, opened 2024-10-17
Allow user to edit Extension Requests before approval

ahsrah7:
Date: 17-10-2024

Developer Name: Harshavardhan

---

## Issue Ticket Number

- closes https://github.com/Real-Dev-Squad/website-backend/issues/2195
## Description

- Allows user to edit an extension request if  it is not yet approved.
### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

</details>

https://github.com/user-attachments/assets/93e6636a-a572-4cbb-a50b-722e3ed52fb8

## Test Coverage
<details>

image
image
image
image
image
image

</details>

### DIFF
--- a/controllers/extensionRequests.js
+++ b/controllers/extensionRequests.js
@@ -202,10 +202,20 @@ const getSelfExtensionRequests = async (req, res) => {
  * @param res {Object} - Express response object
  */
 const updateExtensionRequest = async (req, res) => {
+  const { dev } = req.query;
+  const isDev = dev ? Boolean(dev) : false;
   try {
     const extensionRequest = await extensionRequestsQuery.fetchExtensionRequest(req.params.id);
     if (!extensionRequest.extensionRequestData) {
       return res.boom.notFound("Extension Request not found");
+    }
+
+    if (
+      isDev &&
+      !req.userData?.roles.super_user &&
+      extensionRequest.extensionRequestData.status === EXTENSION_REQUEST_STATUS.APPROVED
+    ) {
+      return res.boom.badRequest("Extension Request cannot be updated");
     }
 
     if (req.body.assignee) {
--- a/middlewares/skipAuthorizeRolesWrapper.js
+++ b/middlewares/skipAuthorizeRolesWrapper.js
@@ -1,0 +1,13 @@
+const skipIfDev = (authorizeMiddleware) => {
+  return (req, res, next) => {
+    const { dev } = req.query;
+    const isDev = dev ? Boolean(dev) : false;
+    if (isDev) {
+      next();
+    } else {
+      authorizeMiddleware(req, res, next);
+    }
+  };
+};
+
+export default skipIfDev;
--- a/routes/extensionRequests.js
+++ b/routes/extensionRequests.js
@@ -10,6 +10,7 @@ const {
   updateExtensionRequestStatus,
   getExtensionRequestsValidator,
 } = require("../middlewares/validators/extensionRequests");
+const { default: skipIfDev } = require("../middlewares/skipAuthorizeRolesWrapper");
 
 router.post("/", authenticate, createExtensionRequest, extensionRequests.createTaskExtensionRequest);
 router.get("/", authenticate, getExtensionRequestsValidator, extensionRequests.fetchExtensionRequests);
@@ -18,7 +19,7 @@ router.get("/:id", authenticate, authorizeRoles([SUPERUSER, APPOWNER]), extensio
 router.patch(
   "/:id",
   authenticate,
-  authorizeRoles([SUPERUSER, APPOWNER]),
+  skipIfDev(authorizeRoles([SUPERUSER, APPOWNER])),
   updateExtensionRequest,
   extensionRequests.updateExtensionRequest
 );
--- a/test/integration/extensionRequests.test.js
+++ b/test/integration/extensionRequests.test.js
@@ -921,6 +921,57 @@ describe("Extension Requests", function () {
         });
     });
 
+    it("User should be able to update the extensionRequest for the given extensionRequestId", function (done) {
+      chai
+        .request(app)
+        .patch(`/extension-requests/${extensionRequestId4}?dev=true`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send({
+          title: "new-title",
+        })
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(204);
+          return done();
+        });
+    });
+
+    it("User should not be able to update the extensionRequest if already approved", function (done) {
+      chai
+        .request(app)
+        .patch(`/extension-requests/${extensionRequestId1}?dev=true`)
+        .set("cookie", `${cookieName}=${jwt}`)
+        .send({
+          title: "new-title",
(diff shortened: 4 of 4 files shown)


routes/extensionRequests.js
```diff
@@ -18,7 +19,7 @@ router.get("/:id", authenticate, authorizeRoles([SUPERUSER, APPOWNER]), extensio
 router.patch(
   "/:id",
   authenticate,
-  authorizeRoles([SUPERUSER, APPOWNER]),
+  skipIfDev(authorizeRoles([SUPERUSER, APPOWNER])),
```
pankajjs, line 22: 1. skipIfDev does not describe what it does by its name.

pankajjs:
There are no tests for disabled feature flag.

ahsrah7:
> 1. There are no tests for disabled feature flag.
> 2. Could you add the test coverage for you changes?

- will the tests for the feature flag not become the tests once feature flag is removed?
- I have added the relevant test cases but  if you have any specific cases please mention

routes/extensionRequests.js
ahsrah7: updated the name. Please have a look

pankajjs:
> > 1. There are no tests for disabled feature flag.
> > 2. Could you add the test coverage for you changes?
> 
> * will the tests for the feature flag not become the tests once feature flag is removed?
> * I have added the relevant test cases but  if you have any specific cases please mention

I don't see 

> > 1. There are no tests for disabled feature flag.
> > 2. Could you add the test coverage for you changes?
> 
> * will the tests for the feature flag not become the tests once feature flag is removed?
> * I have added the relevant test cases but  if you have any specific cases please mention

@ahsrah7 I don't see any test when dev=false.

ahsrah7:
> > > 1. There are no tests for disabled feature flag.
> > > 2. Could you add the test coverage for you changes?
> > 
> > 
> > 
> > * will the tests for the feature flag not become the tests once feature flag is removed?
> > * I have added the relevant test cases but  if you have any specific cases please mention
> 
> @ahsrah7 I don't see any test when dev=false.

@pankajjs  please let me know if I am getting this correctly. The existing flow where only super user  can edit the ERs already has test cases. For the changes I have added where a  normal user can also edit the ER until it is approved is  under FF so have added tests accordingly. For the case which you are mentioning dev=false should be the test cases for the already existing flow which are already present. If I am missing anything please let me know.

ahsrah7:
> As I see in updateExtensionRequest, you have introduced dev flag. When dev is false, you are returning `res.boom.badRequest("Extension Request cannot be updated")` this. But you have not written test for this.

Yes I have introduced the dev flag but when dev flag is false the control does not reach the updateExtensionRequest at all. But thanks for raising this I notice that string type dev flag is always being type casted to true  irrespective of the value provided in the query params.

controllers/extensionRequests.js
```diff
@@ -202,12 +202,23 @@ const getSelfExtensionRequests = async (req, res) => {
...
     }
 
+    if (
+      isDev &&
+      !req.userData?.roles.super_user &&
+      (extensionRequest.extensionRequestData.status === EXTENSION_REQUEST_STATUS.APPROVED ||
+        extensionRequest.extensionRequestData.status === EXTENSION_REQUEST_STATUS.DENIED)
+    ) {
+      return res.boom.badRequest("Extension Request cannot be updated");
+    }
```
Achintya-Chatterjee, line 220: `Extension Request cannot be updated` is a little bit generic. It would be beneficial to provide more context on why the request cannot be updated

ahsrah7: Got it. The message is updated to `Only pending extension request can be updated`. Please let me know if this is better.

middlewares/skipAuthorizeRolesWrapper.js
vinit717, line 1: Please write the test for the middleware

vinit717 (changes requested):

middlewares/skipAuthorizeRolesWrapper.js
ahsrah7: Added the test cases for middleware. Please review

controllers/extensionRequests.js
```diff
@@ -202,12 +202,23 @@ const getSelfExtensionRequests = async (req, res) => {
...
     const extensionRequest = await extensionRequestsQuery.fetchExtensionRequest(req.params.id);
     if (!extensionRequest.extensionRequestData) {
       return res.boom.notFound("Extension Request not found");
     }
 
+    if (
+      isDev &&
+      !req.userData?.roles.super_user &&
+      (extensionRequest.extensionRequestData.status === EXTENSION_REQUEST_STATUS.APPROVED ||
+        extensionRequest.extensionRequestData.status === EXTENSION_REQUEST_STATUS.DENIED)
```
yesyash, line 217: - instead of checking for status not being approved (or) denied can we check for pending directly? so that if any new status is added in the future it does not slip through this check.

middlewares/skipAuthorizeRolesWrapper.js
yesyash, line 1: - why are we doing this?

routes/extensionRequests.js
```diff
@@ -18,7 +19,7 @@ router.get("/:id", authenticate, authorizeRoles([SUPERUSER, APPOWNER]), extensio
 router.patch(
   "/:id",
   authenticate,
-  authorizeRoles([SUPERUSER, APPOWNER]),
+  skipAuthorizeRolesUnderFF(authorizeRoles([SUPERUSER, APPOWNER])),
```
yesyash, line 22: - why're we doing this?

yesyash (changes requested):

middlewares/skipAuthorizeRolesWrapper.js
ahsrah7: Added this middleware to skip the roles check so so that normal user other than super user or app owner can edit a extension request after an extension request is created. As this has to be under feature flag a new middleware `skipAutorizaeRolesWrapper` is which skips the roles check when feature flag is true.

routes/extensionRequests.js
ahsrah7: here the `skipAuthorizeRolesUnderFF` is invoked to skip the `authorizeRoles` middleware so that the end point is accessible to normal user for editing the extension request.

controllers/extensionRequests.js
ahsrah7: Changed the condition to check for pending directly.

routes/extensionRequests.js
yesyash: - do we plan to remove the `skipAuthorizeRolesUnderFF` & `authorizeRoles` middleware when removing the feature flag?

ahsrah7: Yes that is the plan. I have created an issues and added in the PR description under additional notes.

yesyash (approved):

vinit717 (approved):

PR #2192, opened 2024-10-01
remove unverified role on /verify

shobhan-sundar-goutam:
Date: 01/10/24

Developer Name: Shobhan Sundar Goutam

---

## Issue Ticket Number
- Closes <https://github.com/Real-Dev-Squad/discord-slash-commands/issues/234>

## Description
This PR fixes the first issue which is removing unverified role from discord when user runs /verify and connects their RDS Discord

### Documentation Updated?

- [ ] Yes
- [X] No

### Under Feature Flag

- [ ] Yes
- [X] No

### Database Changes

- [ ] Yes
- [X] No

### Breaking Changes

- [ ] Yes
- [X] No

### Development Tested?

- [X] Yes
- [ ] No

## Screenshots
<details>
<summary>Screenshot 1</summary>

https://github.com/user-attachments/assets/98d0a3e3-9a34-429a-81d6-3dd79625eb0a

</details>

## Test Coverage
<details>
<summary>Screenshot 1</summary>

image
image

image
image

</details>

PR #2224, opened 2024-10-23
Revert "remove unverified role on /verify"

vinit717:
Reverts Real-Dev-Squad/website-backend#2192


iamitprakash:
@shobhan-sundar-goutam  reverting this PR because of 1, issues is not working 2. FF is missing as well

PR #2236, opened 2024-10-28
fix: staging issue for remove unverified role on /verify

shobhan-sundar-goutam:
Date: 01/10/24

Developer Name: Shobhan Sundar Goutam

---

## Issue Ticket Number
- Closes <https://github.com/Real-Dev-Squad/discord-slash-commands/issues/234>

## Description
This PR fixes the staging issue for <https://github.com/Real-Dev-Squad/website-backend/pull/2192>

### Documentation Updated?

- [ ] Yes
- [X] No

### Under Feature Flag

- [ ] Yes
- [X] No

### Database Changes

- [ ] Yes
- [X] No

### Breaking Changes

- [ ] Yes
- [X] No

### Development Tested?

- [X] Yes
- [ ] No

## Screenshots
<details>

https://github.com/user-attachments/assets/a3e786bc-7ee4-4d88-9f64-6b57b4930ad1

</details>

## Test Coverage
<details>

image

image

</details>

### DIFF
--- a/controllers/external-accounts.js
+++ b/controllers/external-accounts.js
@@ -4,6 +4,8 @@ const { getDiscordMembers } = require("../services/discordService");
 const { addOrUpdate, getUsersByRole, updateUsersInBatch } = require("../models/users");
 const { retrieveDiscordUsers, fetchUsersForKeyValues } = require("../services/dataAccessLayer");
 const { EXTERNAL_ACCOUNTS_POST_ACTIONS } = require("../constants/external-accounts");
+const discordServices = require("../services/discordService");
+const config = require("config");
 const logger = require("../utils/logger");
 const { markUnDoneTasksOfArchivedUsersBacklog } = require("../models/tasks");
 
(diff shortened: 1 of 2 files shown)


shubhdevelop:
Can you please mention what was the issue happening in the staging ?? doesn't give me any context

shobhan-sundar-goutam:
> Can you please mention what was the issue happening in the staging ?? doesn't give me any context

Issue was roles not getting removed from firestore db as it shouldn't be present there which I found out that for a new user joining RDS discord, unverified role only gets added in the discord and not in the db. So no point of removing from db. This was a mistake from my end, should've checked it earlier.

shubhdevelop:
Okay Just Add it in the PR description would be handy for other reviewer

controllers/external-accounts.js
```diff
@@ -69,6 +71,14 @@ const linkExternalAccount = async (req, res) => {
...
     );
 
+    try {
+      const unverifiedRoleId = config.get("discordUnverifiedRoleId");
+      await discordServices.removeRoleFromUser(unverifiedRoleId, attributes.discordId, req.userData);
+    } catch (error) {
+      logger.error(`Error getting external account data: ${error}`);
+      return res.status(500).json({ message: `Role Deletion failed. Please contact admin.` });
+    }
+
```
prakashchoudhary07, line 81: So since there a 500 respone code, below. Why do we need to create it at multiple places?
What handling is done here or how is it different from the below serviceUnavailable one?

prakashchoudhary07 (changes requested):

controllers/external-accounts.js
prakashchoudhary07: Okay cool. 
But please use res.boom instead

shobhan-sundar-goutam: I've made the changes. Can you please check it?

prakashchoudhary07 (approved):

PR #2257, opened 2024-11-18
Add unverified role to db and remove on /verify

### LINKED ISSUE #2245: Store "Unverified" Role in Database on User Join and Remove upon Verification
### Issue Description

When a new user joins Discord, they are assigned an "Unverified" role/tag. However, this role is currently not stored in the database for tracking. The goal is to update the system so that this "Unverified" role is stored in the database as well. When a user completes the /verify command and authorizes themselves, the "Unverified" role should be removed from both the Discord and the database.

### Expected Behavior

1. When a new user joins Discord:

    - "Unverified" role is assigned to the user in Discord.
    - "Unverified" role is recorded in the database against the user.

3. When a user executes /verify and successfully authorizes:

    - "Unverified" role is removed from both the Discord tag and the database record.

### Current Behavior

- The "Unverified" role is applied to the user in Discord but is not saved in the database.
- Upon verification, there is no removal of the "Unverified" role from the database, as it is not stored there currently.

### Reproducibility

 - [ ] This issue is reproducible
 - [X] This issue is not reproducible

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [X] Medium
 - [ ] Low

### Checklist

- [X] I have read and followed the project's code of conduct.
- [X] I have searched for similar issues before creating this one.
- [X] I have provided all the necessary information to understand and reproduce the issue.
- [X] I am willing to contribute to the resolution of this issue.

shobhan-sundar-goutam:
Date: 19-11-2024

Developer Name: Shobhan Sundar Goutam 

---

## Issue Ticket Number

- Closes #2245 
## Description

When a new user joins Discord, they are assigned an "Unverified" role/tag. However, this role is currently not stored in the database for tracking. The goal is to update the system so that this "Unverified" role is stored in the database as well. When a user completes the /verify command and authorizes themselves, the "Unverified" role should be removed from both the Discord and the database. Also respective logs are stored during this process.
### Documentation Updated?

- [ ] Yes
- [X] No

### Under Feature Flag

- [ ] Yes
- [X] No

### Database Changes

- [ ] Yes
- [X] No

### Breaking Changes

- [ ] Yes
- [X] No

### Development Tested?

- [X] Yes
- [ ] No

## Screenshots
<details>

https://github.com/user-attachments/assets/6f029880-1e92-4063-9344-0103ea83b61b

</details>

## Test Coverage
<details>

image
image

image
image

</details>


controllers/external-accounts.js
```diff
@@ -71,14 +71,20 @@ const linkExternalAccount = async (req, res) => {
...
+    const unverifiedRoleId = config.get("discordUnverifiedRoleId");
+    const unverifiedRoleRemovalResponse = await removeDiscordRoleUtils.removeDiscordRoleFromUser(
+      req.userData,
+      attributes.discordId,
+      unverifiedRoleId
+    );
+
+    if (!unverifiedRoleRemovalResponse.success) {
+      return res.boom.internal(
+        `User details updated but ${unverifiedRoleRemovalResponse.message}. Please contact admin`,
```
pankajjs, line 83: This messages is repeated twice.

pankajjs, line 83: Why this is not in try catch block?

controllers/external-accounts.js
```diff
@@ -71,14 +71,20 @@ const linkExternalAccount = async (req, res) => {
...
+        {
+          message: `User details updated but ${unverifiedRoleRemovalResponse.message}. Please contact admin`,
+        }
+      );
     }
 
     return res.status(204).json({ message: "Your discord profile has been linked successfully" });
```
pankajjs, line 90: For 204 status, why are you sending body?

utils/removeDiscordRoleFromUser.js
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @param {Object} userData - User data.
+ * @param {string} discordId - User's Discord ID.
+ * @param {string} roleid - Discord Role ID.
+ *
+ * @returns {Promise<Object>} - Result with success status and message.
+ */
+export const removeDiscordRoleFromUser = async (userData, discordId, roleid) => {
+  try {
+    const role = await discordRolesModel.isGroupRoleExists({ roleid });
```
pankajjs, line 17: de-structure it on the fly

controllers/external-accounts.js
pankajjs, line 83: Why this condition?
Can't we just throw error from removeDiscordRoleFromUser which will be go to catch block and from there we can send 500 response?

utils/removeDiscordRoleFromUser.js
```diff
@@ -0,0 +1,44 @@
...
+      logType.REMOVE_UNVERIFIED_ROLE,
+      { roleid, userid: discordId },
+      { message: "Unverified role removed successfully" }
+    );
+
+    return { success: true, message: "Role deleted successfully" };
+  } catch (error) {
+    logger.error(`Error removing role ${roleid} for user ${discordId}: ${error.message}`);
+
+    return { success: false, message: error.message };
```
pankajjs, line 42: throw error from here

utils/removeDiscordRoleFromUser.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+export const removeDiscordRoleFromUser = async (userData, discordId, roleid) => {
+  try {
+    const role = await discordRolesModel.isGroupRoleExists({ roleid });
+
+    if (!role.roleExists) {
+      throw new Error("Role doesn't exist");
+    }
+
+    const roleData = role.existingRoles.docs[0].data();
```
pankajjs, line 23: why are we doing this? We already have access to roleId.

utils/removeDiscordRoleFromUser.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Removes a Discord role from a user using Discord Id.
+ *
+ * @param {Object} userData - User data.
+ * @param {string} discordId - User's Discord ID.
+ * @param {string} roleid - Discord Role ID.
+ *
+ * @returns {Promise<Object>} - Result with success status and message.
+ */
+export const removeDiscordRoleFromUser = async (userData, discordId, roleid) => {
```
pankajjs, line 15: Can we use this function to remove other role?

pankajjs:
Left few comments @shobhan-sundar-goutam

controllers/external-accounts.js
shobhan-sundar-goutam: boom.internal is not throwing the custom message if I don't user it twice like this. It is throwing its default message in that case. That's why I'm sending "message" property in the body with that message so that if it fails to take the first one it will definitely take the second one
image
image

controllers/external-accounts.js
shobhan-sundar-goutam: I've not touched that part of the code, it was already there. Also if I remove the body, I don't know how the frontend will behave. So it's beyond my scope.

controllers/external-accounts.js
shobhan-sundar-goutam: Because within the function I've already specified the return statements. So it will always get a success or failure with a message.

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: This is a common function. Why should we assume that anyone using this function will be passing correct roleid through params?

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: Yes exactly. Made this a common function so that it can be used to remove other roles as well from user.

controllers/external-accounts.js
shobhan-sundar-goutam: As this is a common function, by this way it will be much more robust as whoever using this will have full control for success(true/false) and message and they can use in all the ways possible. This is a much more flexible way I guess.

utils/removeDiscordRoleFromUser.js
```diff
@@ -0,0 +1,44 @@
...
+
+    const { wasSuccess } = await discordRolesModel.removeMemberGroup(roleData.roleid, discordId);
+    if (!wasSuccess) {
+      throw new Error("Role deletion failed");
+    }
+
+    await addLog(
+      logType.REMOVE_UNVERIFIED_ROLE,
+      { roleid, userid: discordId },
+      { message: "Unverified role removed successfully" }
```
pankajjs, line 35: If this is a generic function. Then why this message?

utils/removeDiscordRoleFromUser.js
pankajjs: I pointed to line 23.

controllers/external-accounts.js
pankajjs: There is another method for 500. Have you checked?

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: Made the changes. Can you check it please?

controllers/external-accounts.js
shobhan-sundar-goutam: Are you asking about using normal res.status(500) ? If that, I did use that in my previous task. Prakash asked me to use boom

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: Oh, yes you're right. we already have access to roleid but doing this to make sure that we are using the correct roleid only returned directly from db and not relying on external params. If we didn't have that we have surely used the one coming from params. Now if we are getting the id generated through process. So I thought it would be better to use that.

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: Not throwing because not planning to use try catch where the function is used

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: { message: "Unverified role removed successfully" } - this is the body that is getting stored in the logs collection. Seen in few other places that this type of example is being used. That's why

pankajjs: I am saying if it is a generic function then why the message is `Unverified role removed successfully`?

utils/removeDiscordRoleFromUser.js
pankajjs: You already have a condition for checking role.

controllers/external-accounts.js
pankajjs: res.boom.serverUnavailable

shobhan-sundar-goutam: Please have a look at this https://github.com/Real-Dev-Squad/website-backend/pull/2236#discussion_r1826583987

pankajjs: What is the first and second argument of internal function then?

shobhan-sundar-goutam: message and data
This is how it is coming in this case
image

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: Okay made the necessary changes that you asked. Please have a look.

utils/removeDiscordRoleFromUser.js
shobhan-sundar-goutam: Okay got it. This was a bad mistake from my end. Please have a look to the updated change

controllers/external-accounts.js
pankajjs: ```unverifiedRoleRemovalResponse.message```
what will be the value of this?

shobhan-sundar-goutam: For failure cases - Role deleion failed or Role doesn't exist
For success case - Role deleted successfully.

The function already has these. You can see there

pankajjs: In line 23 of utils function can throw any error which will be different from these two case. So are we going to return that error to client?

shobhan-sundar-goutam: I've updated the code. Can you please check it?

pankajjs, line 83: Can you store this message in a variable and use it?

utils/removeDiscordRoleFromUser.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @returns {Promise<Object>} - Result with success status and message.
+ */
+export const removeDiscordRoleFromUser = async (userData, discordId, roleId) => {
+  try {
+    const { roleExists } = await discordRolesModel.isGroupRoleExists({ roleid: roleId });
+
+    if (!roleExists) {
+      const message = "Role doesn't exist";
+      await addLog(logType.REMOVE_ROLE_FROM_USER_FAILED, { roleId }, { message, userid: discordId, userData });
```
pankajjs, line 21: Are we logging errors in db for others api?

shobhan-sundar-goutam: Maybe no. But I think it is a good practice to store log errors in db for mutation queries.

controllers/users.js
```diff
@@ -612,7 +614,19 @@ const markUnverified = async (req, res) => {
...
     usersToApplyUnverifiedRole.forEach((id) => {
-      addRolePromises.push(addRoleToUser(id, unverifiedRoleId));
+      addRolePromises.push(
+        addRoleToUser(id, unverifiedRoleId),
+        addGroupRoleToMember({
+          roleid: unverifiedRoleId,
+          userid: id,
+          date: admin.firestore.Timestamp.fromDate(new Date()),
+        }),
+        addLog(
```
Achintya-Chatterjee, line 624: @shobhan-sundar-goutam Combining multiple database writes into a batch operation can reduce latency and improve performance

test/unit/utils/remvoeDiscordRoleFromUser.test.js
```diff
@@ -0,0 +1,82 @@
...
+  it("should throw an error if role deletion from database failed", async function () {
+    fetchStub.rejects(new Error("Role deletion from database failed"));
+
+    try {
+      await removeDiscordRoleFromUser(userData[0], discordId, roleid);
+    } catch (error) {
+      expect(error.message).to.equal("Role deletion from database failed");
+    }
+  });
+});
```
Achintya-Chatterjee, line 82: 1. Add tests for logging behaviour to ensure success/failure logs are generated as expected.
2. Test for scenarios where role removal partially succeeds (e.g., removed from Discord but not from the database)

controllers/users.js
shobhan-sundar-goutam: Batch insertion is done. Can you check it please

test/unit/utils/remvoeDiscordRoleFromUser.test.js
shobhan-sundar-goutam: I've updated the changes. Can you please check it?

utils/removeDiscordRoleFromUser.ts
```diff
@@ -0,0 +1,57 @@
...
+      await addLog(logType.REMOVE_ROLE_FROM_USER_FAILED, { roleId, userid: discordId, userData }, { message });
+      throw new Error(message);
+    }
+
+    const deleteResponse = await discordActions.removeMemberGroup(roleId, discordId);
+    if (deleteResponse && !deleteResponse.wasSuccess) {
+      const message = "Role deletion from database failed";
+      await addLog(logType.REMOVE_ROLE_FROM_USER_FAILED, { roleId, userid: discordId }, { message, userData });
+      throw new Error(message);
+    }
```
pankajjs, line 43: What should happen if the role is removed from discord but the db operation failed? Shouldn't it role back the previous operation?

shobhan-sundar-goutam: Role back means? Again adding the role to the user in the discord?

pankajjs: The mentioned situation occur right? How would you handle that?

shobhan-sundar-goutam: I think rolling back won't be a good strategy because the primary job of the function is to remove the role from discord. For the rare cases (maybe data inconsistency) for users whose role is not getting removed from db there needs manual intervention to look out for the actual root cause for which logs are being maintained. 
Also for this feature rolling back would make it problematic because the user won't be able to authorize again and remove the unverified role because the modal won't open as in_discord: true will already been done through addorupdate function in controller.

pankajjs: What will happen if we reverse the operation?
Let's first delete it from db and then discord.

controllers/users.js
```diff
@@ -611,11 +616,40 @@ const markUnverified = async (req, res) => {
       }
     });
 
+    const batchSize = 500;
```
pankajjs, line 619: How we came up with this number?

shobhan-sundar-goutam: Firestore limit for batch insertion

utils/removeDiscordRoleFromUser.ts
shobhan-sundar-goutam: What will be the purpose of reversing the operation?

shobhan-sundar-goutam: Wouldn't the whole feature gets blocked if the db deletion process fails?
From user perspective getting the role removed from discord is the first priority, shouldn't it be independent of the db deletion process?
Don't you think getting the role deleted from db(which is a low priority operation) should happen after role getting removed from discord?

pankajjs: Have we agreed on the priority of operation? If yes, then it make sense.

controllers/users.js
```diff
@@ -611,11 +616,40 @@ const markUnverified = async (req, res) => {
...
+
+      batchPromises.push(firestoreBatch.commit());
+    });
+
     usersToApplyUnverifiedRole.forEach((id) => {
       addRolePromises.push(addRoleToUser(id, unverifiedRoleId));
     });
 
-    await Promise.all(addRolePromises);
+    await Promise.all([...addRolePromises, ...batchPromises]);
```
pankajjs, line 652: Can it happen like some number of operation fails in a batch?

shobhan-sundar-goutam: If some operations within a batch insert to Firestore fail, the entire batch will fail, meaning none of the operations in the batch will be applied to the database

Achintya-Chatterjee (approved):

pankajjs (approved):

Achintya-Chatterjee (approved):

pankajjs (approved):

PR #2278, opened 2024-12-02
Feature: Implement Google Oauth login

### LINKED ISSUE #2216: Feature: Users can login via Google.
### Issue Description

Enable users to login via Google OAuth via the main RDS site

### Expected Behavior

If the person is a developer they have to login via GitHub but a non developer should be able to login by their Google Account.

### Current Behavior

Currently all the people whether developers, designers or product managers have the option to only login via GitHub

### Screenshots

N/A

### Reproducibility

 - [ ] This issue is reproducible
 - [x] This issue is not reproducible

### Steps to Reproduce
1. 
2. 

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

listiclehub1:
Date: 02/12/2024

Developer Name: Utkarsh Bansal

---

## Issue Ticket Number

- Closes #2216 

## Description

Adds Google OAuth authentication, enabling users to log in or sign up with their Google account.

### Documentation Updated?

- [x] Yes
- [ ] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

https://github.com/user-attachments/assets/50253c75-4e41-4aa7-aa78-16ef4096d3a6

</details>

## Test Coverage
<details>
Screenshot 2024-12-02 at 3 28 25 PM

</details>

## Additional Notes

- Need to generate oauth credentials for staging as well as prod from google cloud console.
- [Design doc](https://docs.google.com/document/d/18y9qPM1Kdp8A-WqOcsxJrkb4VZL2d_KmnTmc0gSOtHE/edit?tab=t.0#heading=h.fgev8pvsrjqi)
- [API contract PR](https://github.com/Real-Dev-Squad/website-api-contracts/pull/202)


controllers/auth.js
```diff
@@ -9,6 +9,81 @@ const {
...
+        secure: true,
+        sameSite: "lax",
+      };
+
+      res.cookie(config.get("userToken.cookieName"), token, cookieOptions);
+
+      if (incompleteUserDetails) authRedirectionUrl = "https://my.realdevsquad.com/new-signup";
+
+      return res.redirect(authRedirectionUrl);
+    })(req, res, next);
```
yesyash, line 80: - let's move the logic to a different function instead of adding it inside the callback function of the `passport.authenticate` method.

controllers/auth.js
```diff
@@ -77,12 +151,27 @@ const githubAuthCallback = (req, res, next) => {
...
         github_display_name: user.displayName,
+        email: user._json.email,
         github_created_at: Number(new Date(user._json.created_at).getTime()),
         github_user_id: user.id,
         created_at: Date.now(),
         updated_at: null,
       };
 
+      if (userData.email === null) {
+        const res = await fetch("https://api.github.com/user/emails", {
```
yesyash, line 162: - let's move the base path (https://api.github.com) inside an env variable, to allow it to be configurable for different environments if the situation arises.

controllers/auth.js
```diff
@@ -77,12 +151,27 @@ const githubAuthCallback = (req, res, next) => {
...
         github_id: user.username,
         github_display_name: user.displayName,
+        email: user._json.email,
         github_created_at: Number(new Date(user._json.created_at).getTime()),
         github_user_id: user.id,
         created_at: Date.now(),
         updated_at: null,
       };
 
+      if (userData.email === null) {
```
yesyash, line 161: - this will also check for undefined along with null. 
```suggestion
      if (!userData.email) {
```

models/users.js
```diff
@@ -117,15 +117,23 @@ const addOrUpdate = async (userData, userId = null) => {
...
     }
+
+    if (userData.email && (!user || (user && user.empty))) {
+      user = await userModel.where("email", "==", userData.email).limit(1).get();
+    }
+
     if (user && !user.empty && user.docs !== null) {
-      await userModel.doc(user.docs[0].id).set({ ...userData, updated_at: Date.now() }, { merge: true });
+      const { created_at: createdAt, ...updatedUserData } = userData;
+      await userModel.doc(user.docs[0].id).set({ ...updatedUserData, updated_at: Date.now() }, { merge: true });
```
yesyash, line 136: - can `user.docs` be empty? if yes are we handling that case?

models/users.js
```diff
@@ -117,15 +117,23 @@ const addOrUpdate = async (userData, userId = null) => {
...
     if (userData.github_user_id) {
       user = await userModel.where("github_user_id", "==", userData.github_user_id).limit(1).get();
     }
-    if (!user || (user && user.empty)) {
+
+    if (userData.github_id && (!user || (user && user.empty))) {
       user = await userModel.where("github_id", "==", userData.github_id).limit(1).get();
     }
+
+    if (userData.email && (!user || (user && user.empty))) {
```
yesyash, line 130: - if we're already doing a `!user` check do we need to do `(user && user.empty))` can we just do `!user || user.empty` ?

yesyash (changes requested):

controllers/auth.js
```diff
@@ -9,6 +9,81 @@ const {
...
 } = require("../constants/errorMessages");
 
+const googleAuthLogin = (req, res, next) => {
+  const { redirectURL, dev } = req.query;
+  if (dev === "true") {
+    return passport.authenticate("google", {
+      scope: ["email"],
+      state: redirectURL,
+    })(req, res, next);
+  } else {
```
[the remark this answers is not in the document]
listiclehub1: if we remove the else block without handling the case where the dev query parameter is missing, the request will be left hanging as this function is directly used here
`router.get("/google/login", auth.googleAuthLogin);`

controllers/auth.js
```diff
@@ -9,6 +9,81 @@ const {
...
+
+      if (`.${redirectUrl.hostname}`.endsWith(`.${rdsUiUrl.hostname}`)) {
+        // Matching *.realdevsquad.com
+        authRedirectionUrl = redirectUrl;
+      } else {
+        logger.error(`Malicious redirect URL provided URL: ${redirectUrl}, Will redirect to RDS`);
+      }
+    } catch (error) {
+      logger.error("Invalid redirect URL provided", error);
+    }
```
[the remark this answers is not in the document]
listiclehub1: Will remove it.

controllers/auth.js
listiclehub1: Will move to a different function.

models/users.js
listiclehub1: `user.docs` can be empty. If `user.docs` is empty, `user.empty` will be true, which implies that no documents were found.

models/users.js
listiclehub1: I'll update the condition to !user || user.empty to simplify it.

controllers/auth.js
listiclehub1: I will update it.

controllers/auth.js
listiclehub1: I will update the api to env.

controllers/auth.js
```diff
@@ -9,6 +9,81 @@ const {
...
+      const userData = {
+        email: user.emails[0].value,
+        created_at: Date.now(),
+        updated_at: null,
+      };
+
+      const userDataFromDB = await users.fetchUser({ email: userData.email });
+
+      if (userDataFromDB.userExists) {
+        if (userDataFromDB.user.roles.developer === true) {
```
VinuB-Dev, line 57: Let's add optional chaining here: `userDataFromDB.user.roles?.developer`, so it doesn't throw an error if the developer role is missing.
If we are certain that roles.developer is always a boolean, we can simplify the condition to:
`if (userDataFromDB.user.roles?.developer)`

listiclehub1: I will update the code to `if (userDataFromDB.user.roles?.developer)`

controllers/auth.js
```diff
@@ -9,6 +9,81 @@ const {
...
+    }
+  }
+  try {
+    return passport.authenticate("google", { session: false }, async (err, accessToken, user) => {
+      if (err) {
+        logger.error(err);
+        return res.boom.unauthorized("User cannot be authenticated");
+      }
+      const userData = {
+        email: user.emails[0].value,
```
VinuB-Dev, line 49: can we add a check to ensure emails is present and in the required format before accessing `user.emails[0].value`?

listiclehub1: Have added a check for this.

Achintya-Chatterjee (commented):
@listiclehub1 can you please add the design doc here ?

Achintya-Chatterjee (commented):
@listiclehub1 also you have to change the API contract

listiclehub1:
> @listiclehub1 can you please add the design doc here ?

@Achintya-Chatterjee I have added the design doc in the additional notes section.

>@listiclehub1 also you have to change the API contract

@Achintya-Chatterjee I didn’t find an explicit API contract for the existing GitHub OAuth flow in the codebase.
Could you clarify whether any updates or changes to the current GitHub OAuth flow are required as well.

controllers/auth.js
```diff
@@ -9,6 +9,91 @@ const {
...
+
+    if (incompleteUserDetails) {
+      authRedirectionUrl = "https://my.realdevsquad.com/new-signup";
+    }
+
+    return res.redirect(authRedirectionUrl);
+  } catch (err) {
+    logger.error(err);
+    return res.boom.unauthorized("User cannot be authenticated");
+  }
```
Achintya-Chatterjee, line 67: Users are being sent generic error messages. To improve debugging, distinct error messages should be sent for different error scenarios. Some errors might not be authentication failures but server errors, yet they're all returning `401`.

controllers/auth.js
```diff
@@ -9,6 +9,91 @@ const {
...
+async function handleGoogleLogin(req, res, user, authRedirectionUrl) {
+  const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
+  try {
+    if (!user.emails || user.emails.length === 0) {
+      throw new Error("Email not found in user data");
+    }
+    const userData = {
+      email: user.emails[0].value,
+      created_at: Date.now(),
+      updated_at: null,
```
Achintya-Chatterjee, line 33: While Google typically provides validated emails, as a best practice we should still validate the email format from our end. This provides an additional security layer. Let's say for some reason `user.emails[0]` is `undefined`, this could throw an error. Also, Google can return multiple emails, and we should ideally check for the primary email.

models/users.js
```diff
@@ -117,15 +117,23 @@ const addOrUpdate = async (userData, userId = null) => {
...
     }
+
+    if (userData.email && (!user || user.empty)) {
+      user = await userModel.where("email", "==", userData.email).limit(1).get();
+    }
+
     if (user && !user.empty && user.docs !== null) {
-      await userModel.doc(user.docs[0].id).set({ ...userData, updated_at: Date.now() }, { merge: true });
+      const { created_at: createdAt, ...updatedUserData } = userData;
+      await userModel.doc(user.docs[0].id).set({ ...updatedUserData, updated_at: Date.now() }, { merge: true });
```
Achintya-Chatterjee, line 136: Currently, we're performing sequential Firestore queries. so this would lead to slower response times as each query waits for the previous one to complete. we should consider using `Promise.all()` to run these queries in parallel.

controllers/auth.js
```diff
@@ -9,6 +9,91 @@ const {
...
+    const cookieOptions = {
+      domain: rdsUiUrl.hostname,
+      expires: new Date(Date.now() + config.get("userToken.ttl") * 1000),
+      httpOnly: true,
+      secure: true,
+      sameSite: "lax",
+    };
+
+    res.cookie(config.get("userToken.cookieName"), token, cookieOptions);
+
```
Achintya-Chatterjee, line 58: At Real Dev Squad we operate across multiple subdomains (api.realdevsquad.com, www.realdevsquad.com, members.realdevsquad.com, status.realdevsquad.com etc.), I think we should review the `sameSite: "lax"` cookie setting. we should consider making this configurable through the config files to allow different settings per environment, as our cross-origin requirements might vary. Also, the domain setting might need adjustment for proper subdomain handling.

test/integration/auth.test.js
```diff
@@ -254,4 +256,191 @@ describe("auth", function () {
...
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle" })
+      .redirects(0)
+      .end((err, res) => {
+        if (err) {
+          return done(err);
+        }
+
+        expect(res).to.have.status(302);
+        // rds-session=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VySWQiOiJ1c2VySWQiLCJpYXQiOjE1OTkzOTEzODcsImV4cCI6MTYwMTk4MzM4N30.AljtAmXpZUmErubhSBbA0fQtG9DwH4ci6iroa9z5MBjIPFfQ5FSbaOqU0CQlmgOe-U7XDVPuGBp7GzBzA4yCH7_3PSS9JrHwEVZQQBScTUC-WHDradit5nD1ryKPqJE2WlRO6q0uLOKEukMj-7iPXQ-ykdYwtlokhyJbLVS1S3E; Domain=realdevsquad.com; Path=/; Expires=Tue, 06 Oct 2020 11:23:07 GMT; HttpOnly; Secure
```
Achintya-Chatterjee, line 415: is this comment necessary?

test/integration/auth.test.js
```diff
@@ -254,4 +256,191 @@ describe("auth", function () {
...
+    const res = await chai
+      .request(app)
+      .get(`/auth/google/callback`)
+      .query({ code: "codeReturnedByGoogle", state: invalidRedirectUrl })
+      .redirects(0);
+    expect(res).to.have.status(302);
+    expect(res.headers.location).to.equal(rdsUiUrl);
+  });
+
+  it("should send a cookie with JWT in the response for google user", function (done) {
```
Achintya-Chatterjee, line 396: No tests for invalid email formats

test/integration/auth.test.js
```diff
@@ -254,4 +256,191 @@ describe("auth", function () {
...
+    const res = await chai
+      .request(app)
+      .get(`/auth/google/callback`)
+      .query({ code: "codeReturnedByGoogle", state: invalidRedirectUrl })
+      .redirects(0);
+    expect(res).to.have.status(302);
+    expect(res.headers.location).to.equal(rdsUiUrl);
+  });
+
+  it("should redirect the google user to realdevsquad.com if invalid redirect URL provided", async function () {
```
Achintya-Chatterjee, line 378: No tests for developer trying to use Google login

test/integration/auth.test.js
```diff
@@ -254,4 +256,191 @@ describe("auth", function () {
...
+    const res = await chai
+      .request(app)
+      .get(`/auth/google/callback`)
+      .query({ code: "codeReturnedByGoogle", state: rdsUrl })
+      .redirects(0);
+    expect(res).to.have.status(302);
+    expect(res.headers.location).to.equal(rdsUrl);
+  });
+
+  it("should redirect the google user to realdevsquad.com if non RDS URL provided, any url that is other than *.realdevsquad.com is invalid", async function () {
```
Achintya-Chatterjee, line 360: No tests for duplicate email scenarios

test/integration/auth.test.js
```diff
@@ -254,4 +256,191 @@ describe("auth", function () {
...
+    const res = await chai
+      .request(app)
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle", state: rdsUiUrl })
+      .redirects(0);
+    expect(res).to.have.status(302);
+    expect(res.headers.location).to.equal(rdsUiUrl);
+  });
+
+  it("should redirect the google user to the redirect URL provided on successful login, if user has incomplete user details false", async function () {
```
Achintya-Chatterjee, line 343: No tests for invalid redirect URLs

controllers/auth.js
```diff
@@ -9,6 +9,91 @@ const {
...
+    };
+
+    const userDataFromDB = await users.fetchUser({ email: userData.email });
+
+    if (userDataFromDB.userExists) {
+      if (userDataFromDB.user.roles?.developer) {
+        const errorMessage = encodeURIComponent("Google login is restricted for developer role.");
+        return res.redirect(`${authRedirectionUrl}?error=${errorMessage}`);
+      }
+    }
```
Achintya-Chatterjee, line 43: Role validation is looking too simplistic. There should be a proper role management system rather than just checking for a boolean flag.

We can do something like this: these improvements would make the code more robust, secure, and maintainable.

`const REQUIRED_AUTH_PROVIDERS = {
  developer: 'github',
  designer: ['google', 'github'],
  product_manager: ['google', 'github']
};

const validateAuthProvider = (userRole, provider) => {
  const allowedProviders = REQUIRED_AUTH_PROVIDERS[userRole];
  return allowedProviders.includes(provider);
};`

Achintya-Chatterjee (changes requested):

test/integration/auth.test.js
listiclehub1: I have added the tests for invalid URL [here](https://github.com/Real-Dev-Squad/website-backend/pull/2278/files/0205c95872f984d471484ed33a233581a28af9d4#diff-aaf650b060bb12ba527fcfeb22c2c2d990137d3e1b4587cc76c1b5f3c5776465R378) Line 378

controllers/auth.js
listiclehub1: I have used 401 Unauthorized in the error handling as most of the errors are related to authentication or issues with the user's Google data (such as missing or unverified emails). Since these are primarily authentication-related failures, I believe a 401 is the most appropriate response. Additionally, the error details are logged for debugging purposes.
Let me know if you would like me to adjust the error handling further or add more specific error messages.

test/integration/auth.test.js
listiclehub1: could you kindly clarify what type of test you are looking for regarding the "duplicate email scenarios"?

controllers/auth.js
listiclehub1: Currently, we are not assigning specific roles to users, as we have been assuming the roles will be available in the future, based on a discussion with Ankush. 
We plan to implement a proper role collection feature as part of the new-signup process, and once that is in place, we will revisit and improve the role validation logic to reflect the more flexible, secure structure you're proposing.

test/integration/auth.test.js
listiclehub1: I have added this test now.

controllers/auth.js
listiclehub1: I’ve added a basic check for primary emails also now to ensure that the user’s email data exists and is properly verified. The code verifies that:
The user.emails array is not empty, and
The first verified email is selected.
While Google already verifies emails on their end, this additional check ensures that we're handling any unexpected cases (e.g., missing or unverified emails) gracefully. Typically, the first verified email should be the primary one, and this is the one we use.

test/integration/auth.test.js
listiclehub1: could you kindly clarify what type of test you are looking for regarding the "invalid email formats"?

controllers/auth.js
Achintya-Chatterjee: While logging helps with debugging, we should still provide more specific error messages to users. Even within authentication failures, there can be different scenarios (invalid email, role restrictions, server errors, etc) that should be communicated differently. This helps both users and front-end developers handle errors appropriately.

test/integration/auth.test.js
Achintya-Chatterjee: For invalid email formats, we should test:
- Malformed email addresses
- Empty email field
- Multiple email addresses where the primary is not first

test/integration/auth.test.js
Achintya-Chatterjee: For duplicate email scenarios, we should test:
- Same email trying to log in with different providers
- Existing email attempting new registration
- Email conflicts during user updates

models/users.js
Achintya-Chatterjee: I see your point about early termination when a match is found. However, in the worst case (no match found), we still end up running all queries sequentially. The actual performance impact would depend on our typical use cases. Let's monitor query performance in production and revisit this if we see latency issues.

controllers/auth.js
listiclehub1: I have added more specific error messages for users.

test/integration/auth.test.js
listiclehub1: Have added test for all the invalid email formats

test/integration/auth.test.js
listiclehub1: i have added test for this `Same email trying to log in with different providers`

Achintya-Chatterjee (approved):

vinit717:
Can you add a small video how this google login work

prakashchoudhary07:
Please update the description appropriately

routes/auth.ts
```diff
@@ -9,6 +9,10 @@ router.get("/github/login", auth.githubAuthLogin);
 
 router.get("/github/callback", auth.githubAuthCallback);
 
+router.get("/google/login", auth.googleAuthLogin);
+
+router.get("/google/callback", auth.googleAuthCallback);
+
```
prakashchoudhary07, line 15: There is a middleware present to check for dev flag at route level, see if that can be used

controllers/auth.js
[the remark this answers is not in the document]
prakashchoudhary07: How so? 
won't it be the same, if you add else or not?

controllers/auth.js
```diff
@@ -9,6 +9,99 @@ const {
...
+
+const googleAuthCallback = (req, res, next) => {
+  const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
+  let authRedirectionUrl = rdsUiUrl;
+
+  if ("state" in req.query) {
+    try {
+      const redirectUrl = new URL(req.query.state);
+
+      if (`.${redirectUrl.hostname}`.endsWith(`.${rdsUiUrl.hostname}`)) {
```
prakashchoudhary07, line 86: Can this be a separate util function, I think this is to be done in GitHub auth also.

controllers/auth.js
```diff
@@ -9,6 +9,99 @@ const {
...
+
+    const userDataFromDB = await users.fetchUser({ email: userData.email });
+
+    if (userDataFromDB.userExists) {
+      if (userDataFromDB.user.roles?.developer) {
+        const errorMessage = encodeURIComponent("Google login is restricted for developer role.");
+        const separator = authRedirectionUrl.search ? "&" : "?";
+        return res.redirect(`${authRedirectionUrl}${separator}error=${errorMessage}`);
+      }
+    }
```
prakashchoudhary07, line 51: Can you please explain this part?

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+    const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl")).href;
+    sinon.stub(passport, "authenticate").callsFake((strategy, options, callback) => {
+      callback(null, "accessToken", googleUserInfo[0]);
+      return (req, res, next) => {};
+    });
+
+    const res = await chai
+      .request(app)
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle", state: rdsUiUrl })
```
prakashchoudhary07, line 337: What if state is not passed? and code is not passed

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+    const res = await chai
+      .request(app)
+      .get(`/auth/google/callback`)
+      .query({ code: "codeReturnedByGoogle", state: invalidRedirectUrl })
+      .redirects(0);
+    expect(res).to.have.status(302);
+    expect(res.headers.location).to.equal(rdsUiUrl);
+  });
+
+  it("should send a cookie with JWT in the response for google user", function (done) {
```
prakashchoudhary07, line 396: Please use a more appropriate name, when does this happen

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+    });
+
+    const res = await chai
+      .request(app)
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle", state: rdsUiUrl })
+      .redirects(0);
+    expect(res).to.have.status(302);
+    const errorMessage = "Google login is restricted for developer role.";
+    const expectedUrl = `https://realdevsquad.com/?error=${encodeURIComponent(errorMessage)}`;
```
prakashchoudhary07, line 443: Why is the error message being passed as a query param?

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+        expect(res.headers["set-cookie"][0]).to.include("HttpOnly");
+        expect(res.headers["set-cookie"][0]).to.include("Secure");
+        expect(res.headers["set-cookie"][0]).to.include(`Domain=${rdsUiUrl.hostname}`);
+        expect(res.headers["set-cookie"][0]).to.include("SameSite=Lax");
+
+        return done();
+      });
+  });
+
+  it("should redirect the google user to login page if the user is a developer", async function () {
```
prakashchoudhary07, line 428: How do we know if its developer without users already being present in RDS

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle", state: rdsUiUrl })
+      .redirects(0);
+    expect(res).to.have.status(302);
+    const errorMessage = "Google login is restricted for developer role.";
+    const expectedUrl = `https://realdevsquad.com/?error=${encodeURIComponent(errorMessage)}`;
+    expect(res.headers.location).to.equal(expectedUrl);
+  });
+
+  it("should recognize existing user by email when logging in via different OAuth provider", async function () {
```
prakashchoudhary07, line 447: Appropriate name?

prakashchoudhary07: Where is the merging test for 2 accounts?

controllers/auth.js
```diff
@@ -77,12 +169,28 @@ const githubAuthCallback = (req, res, next) => {
...
+            Authorization: `token ${accessToken}`,
+          },
+        });
+        const emails = await res.json();
+        const primaryEmails = emails.filter((item) => item.primary);
+
+        if (primaryEmails.length > 0) {
+          userData.email = primaryEmails[0].email;
+        }
+      }
```
prakashchoudhary07, line 192: When does this happens? when does the above code userData.email is not present?

controllers/auth.js
```diff
@@ -77,12 +169,28 @@ const githubAuthCallback = (req, res, next) => {
...
       };
 
+      if (!userData.email) {
+        const githubBaseUrl = config.get("githubApi.baseUrl");
+        const res = await fetch(`${githubBaseUrl}/user/emails`, {
+          headers: {
+            Authorization: `token ${accessToken}`,
+          },
+        });
+        const emails = await res.json();
```
prakashchoudhary07, line 186: Will this always get the email, when is it that this will be called?

controllers/auth.js
```diff
@@ -9,6 +9,99 @@ const {
...
+      return res.boom.unauthorized("No verified email found in Google account");
+    }
+
+    const userData = {
+      email: primaryEmail.value,
+      created_at: Date.now(),
+      updated_at: null,
+    };
+
+    const userDataFromDB = await users.fetchUser({ email: userData.email });
```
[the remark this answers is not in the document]
listiclehub1: When a user logs in via Google OAuth, I'll first check if the email associated with their Google account already exists in our database (which was previously registered through GitHub). If an email match is found, I’ll log the user into their existing account.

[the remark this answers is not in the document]
listiclehub1: If the email doesn’t exist in our system (i.e., they are logging in with Google for the first time), I'll treat them as a new user and create a new account for them.

controllers/auth.js
listiclehub1: I’m checking if the user exists in the database, and if they do, I then verify whether they have a "developer" role. As Ankush told me, if the user has the developer role, we're preventing them from logging in via Google.

The user is then redirected back to the original URL with an error query parameter, which contains a message indicating that Google login is restricted for users with the "developer" role. This error parameter will be used on the frontend to display a toast notification to inform the user of the restriction.

controllers/auth.js
listiclehub1: done

controllers/auth.js
listiclehub1: When a user logs in via GitHub OAuth, we may not always get the user's email directly in the initial response, particularly if the user has set their email to private in their GitHub settings. To handle this, I’m fetching the email using the access token via GitHub’s /user/emails endpoint. This ensures we always retrieve the correct email, even if it's not included in the initial OAuth response.

controllers/auth.js
listiclehub1: When a user logs in via GitHub OAuth, the email field might not always be included in the initial OAuth response. This depends on whether the user has made their email private or not in their GitHub account settings.

routes/auth.ts
listiclehub1: done

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
     expect(res.headers["set-cookie"][1]).to.include(`Domain=${rdsUiUrl.hostname}`);
     expect(res.headers["set-cookie"][1]).to.include("SameSite=Lax");
   });
+
+  it("should redirect to the correct URL and update user email when GitHub API returns primary email", async function () {
```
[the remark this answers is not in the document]
listiclehub1: There are already existing tests in the codebase that cover GitHub-related functionality (All the lines above this is for testing GitHub Oauth). I wrote this additional test specifically to cover the new functionality I added, which involves fetching the user's email using the access token.

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+        _json: { email: null, created_at: "2022-01-01" },
+        id: 12345,
+      });
+      return (req, res, next) => {};
+    });
+
+    const res = await chai
+      .request(app)
+      .get(`/auth/github/callback`)
+      .query({ code: "codeReturnedByGithub", state: rdsUrl })
```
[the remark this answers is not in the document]
listiclehub1: The codeReturnedByGithub in the test is a placeholder for the actual authorization code that GitHub returns after a user successfully authenticates and authorizes the application. In a real scenario, this code would be provided by GitHub during the OAuth flow.

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+    expect(res.headers.location).to.equal(googleOauthURL);
+  });
+
+  it("should return google call back URL with redirectUrl", async function () {
+    const RDS_MEMBERS_SITE_URL = "https://members.realdevsquad.com";
+    const googleOauthURL = generateGoogleAuthRedirectUrl({ state: RDS_MEMBERS_SITE_URL });
+    const res = await chai
+      .request(app)
+      .get("/auth/google/login?dev=true")
+      .query({ redirectURL: RDS_MEMBERS_SITE_URL })
```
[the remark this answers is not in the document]
listiclehub1: added this test

test/integration/auth.test.js
listiclehub1: If the state is not passed, the user will be redirected to the RDS main site. If the code is missing or invalid, the request will return a 401 Unauthorized error because the code is necessary to exchange for an access token and complete the OAuth flow.

I’ve also written tests to verify both of these scenarios.

test/integration/auth.test.js
listiclehub1: Changed the name to `should Issue Jwt Cookie On using Google Login`.

test/integration/auth.test.js
listiclehub1: If the user is not already present in the RDS, we treat them as a new user and create an account for them. In this case, we don’t need to check for the "developer" role, since the user doesn't exist yet. The role (e.g., "developer") will be assigned during the new-signup application (in Future).

test/integration/auth.test.js
listiclehub1: The error message is being passed as a query param because this allows the frontend to easily retrieve and display it to the user as a toast notification.

test/integration/auth.test.js
listiclehub1: I’ve updated the test name to make it more appropriate. 
This test checks that an existing Google user with the same email can successfully log in via GitHub OAuth.

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+        ],
+      };
+      callback(null, "accessToken", googleUser);
+      return (req, res, next) => {};
+    });
+
+    const res = await chai
+      .request(app)
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle", state: rdsUiUrl })
```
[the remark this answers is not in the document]
listiclehub1: The codeReturnedByGoogle is a placeholder for the authorization code returned by Google during the OAuth flow.

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+      .redirects(0);
+    expect(res).to.have.status(302);
+    expect(res.headers.location).to.equal(rdsUiUrl);
+  });
+
+  it("should return 401 if google email does not exist", async function () {
```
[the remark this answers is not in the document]
listiclehub1: Typically, Google OAuth will always return the user's email, as long as the user has granted the appropriate permissions. I’ve added this test specifically to handle any edge cases where the email might not be returned.

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+      .request(app)
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle", state: rdsUiUrl })
+      .redirects(0);
+
+    expect(res).to.have.status(401);
+    expect(res.body.message).to.equal("No email found in Google account");
+  });
+
+  it("should return 401 if no verified email exists", async function () {
```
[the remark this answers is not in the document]
listiclehub1: Typically, Google OAuth will always return the user's verified email. I’ve added this test specifically to handle any unknown cases where there is no verified email.

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+    const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl")).href;
+    sinon.stub(passport, "authenticate").callsFake((strategy, options, callback) => {
+      const userInfoWithUnverifiedEmail = {
+        ...googleUserInfo[0],
+        emails: [{ value: "test@example.com", verified: false }],
+      };
+      callback(null, "accessToken", userInfoWithUnverifiedEmail);
+      return (req, res, next) => {};
+    });
```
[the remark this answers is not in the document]
listiclehub1: done

test/integration/auth.test.js
```diff
@@ -254,4 +256,296 @@ describe("auth", function () {
...
+      .request(app)
+      .get("/auth/google/callback")
+      .query({ code: "codeReturnedByGoogle", state: rdsUiUrl })
+      .redirects(0);
+
+    expect(res).to.have.status(401);
+    expect(res.body.message).to.equal("No verified email found in Google account");
+  });
+
+  it("should return 401 if google auth call fails", function (done) {
```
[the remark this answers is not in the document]
listiclehub1: The test is designed to handle cases where an invalid or expired authorization code is passed to the callback. In such cases, the appropriate response would be a 401 Unauthorized, as the issue is with the client

controllers/auth.js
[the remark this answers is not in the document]
listiclehub1: (Although now i have added devFlag middleware so this is not required but still wanted feedback)

@prakashchoudhary07  I’m not entirely clear on how it would behave same. My understanding is that if we remove the `else` block and don't explicitly handle the case where the `dev` query parameter is missing, the request could continue executing without sending a response, potentially leaving the user in a waiting state. This could lead to an unhandled request where the user doesn’t get any feedback or proper handling for the missing parameter.

controllers/auth.js
prakashchoudhary07: so basically you are passing an error message in the query param? is this the correct approach?

controllers/auth.js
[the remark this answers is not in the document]
prakashchoudhary07: But if they are already logged in from GitHub? then 2 accounts for the same user?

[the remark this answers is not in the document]
listiclehub1: So if the user doesn't have email and uses Google we will be creating a new account. 

But as discussed in the design doc, the initial plan was to ask users for an email when it wasn't available during OAuth login. However, after talking with Ankush, we agreed that without email verification in place, we risk storing incorrect or inconsistent email data. To mitigate this, Ankush suggested that we move forward with the feature, but behind a feature flag, until email verification is properly implemented.

controllers/auth.js
listiclehub1: To clarify, passing error messages via query parameters is consistent with how OAuth providers like Google handle errors. When authentication fails, Google (and others) typically pass error codes and descriptions through the URL, as explained [here](https://www.oauth.com/oauth2-servers/server-side-apps/possible-errors/).

Since our backend handles the OAuth flow and user verification, we need to send error feedback to the frontend. Given that the OAuth redirect directly returns to the backend, and the frontend can't interact with the OAuth process in real-time, using query parameters is the most efficient way to pass error messages to the frontend for display.

Please let me know if you believe this approach is not correct.

controllers/auth.js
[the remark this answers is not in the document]
prakashchoudhary07: So what is the plan ahead for getting the email process in place?

controllers/auth.js
```diff
@@ -9,6 +9,116 @@ const {
...
+    isV2FlagPresent,
+    devMode,
+  };
+}
+
+async function handleGoogleLogin(req, res, user, authRedirectionUrl) {
+  const rdsUiUrl = new URL(config.get("services.rdsUi.baseUrl"));
+  try {
+    if (!user.emails || user.emails.length === 0) {
+      logger.error("Google login failed: No emails found in user data");
```
prakashchoudhary07, line 61: What is the case for no emails?

controllers/auth.js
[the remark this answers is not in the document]
listiclehub1: Ankush suggested that this should be treated as a separate task in the future. This task can be taken up by me, or by anyone else who wants to work on the backend.

controllers/auth.js
listiclehub1: Typically, when using Google OAuth for authentication, Google provides the user's email address as part of the authentication response. This is the expected behavior, and in most cases, we can rely on this to identify users. However, as Achintya pointed out, adding a check to verify that an email is returned can add an extra layer of security for our application.

controllers/auth.js
```diff
@@ -77,12 +162,28 @@ const githubAuthCallback = (req, res, next) => {
...
         github_user_id: user.id,
         created_at: Date.now(),
         updated_at: null,
       };
 
+      if (!userData.email) {
+        const githubBaseUrl = config.get("githubApi.baseUrl");
+        const res = await fetch(`${githubBaseUrl}/user/emails`, {
+          headers: {
+            Authorization: `token ${accessToken}`,
```
prakashchoudhary07, line 176: Which accessToken is this?

listiclehub1: This is received from Github so that we can fetch user details from their servers after oauth.

controllers/auth.js
prakashchoudhary07: So GitHub gives email, by directly calling the above URL?
Can you please add the API reference

routes/auth.ts
```diff
@@ -4,11 +4,16 @@ import auth from "../controllers/auth";
...
 import qrCodeAuthValidator from "../middlewares/validators/qrCodeAuth";
+import { devFlagMiddleware } from "../middlewares/devFlag";
 
 router.get("/github/login", auth.githubAuthLogin);
 
 router.get("/github/callback", auth.githubAuthCallback);
 
+router.get("/google/login", devFlagMiddleware, auth.googleAuthLogin);
+
+router.get("/google/callback", auth.googleAuthCallback);
```
prakashchoudhary07, line 15: should this also be behind dev flag?

controllers/auth.js
listiclehub1: https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user

test/integration/auth.test.js
```diff
@@ -254,4 +256,288 @@ describe("auth", function () {
...
+    expect(res.headers.location).to.equal(googleOauthURL);
+  });
+
+  it("should redirect the google user to new sign up flow if they are have incomplete user details true", async function () {
+    const redirectURL = "https://my.realdevsquad.com/new-signup";
+    stubPassportAuthenticate(googleUserInfo[0]);
+
+    const res = await chai
+      .request(app)
+      .get("/auth/google/callback")
```
prakashchoudhary07, line 315: Shouldn't this also be behind a flag?

routes/auth.ts
listiclehub1: When Google redirects to the callback URL, it usually appends its own parameters (such as code, state, etc.) but will not include any query parameters like dev=true. and if a user directly visits this url he will not be able to authenticate and receive 401 error.

test/integration/auth.test.js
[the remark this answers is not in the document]
prakashchoudhary07: Like?

test/integration/auth.test.js
listiclehub1: RE: 
When Google redirects to the callback URL, it usually appends its own parameters (such as code, state, etc.) but will not include any query parameters like dev=true. and if a user directly visits this url he will not be able to authenticate and receive 401 error.

test/integration/auth.test.js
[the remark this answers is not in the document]
prakashchoudhary07: Can a google account exist without emal?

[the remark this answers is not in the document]
listiclehub1: No it cannot but I added this test as I added code for checking for emails from Google.

controllers/auth.js
```diff
@@ -9,6 +9,116 @@ const {
...
+
+    const token = authService.generateAuthToken({ userId });
+
+    const cookieOptions = {
+      domain: rdsUiUrl.hostname,
+      expires: new Date(Date.now() + config.get("userToken.ttl") * 1000),
+      httpOnly: true,
+      secure: true,
+      sameSite: "lax",
+    };
```
prakashchoudhary07, line 96: NIT: let cookie generation be at single, place. since the same cookie with the same values is set

controllers/auth.js
```diff
@@ -9,6 +9,116 @@ const {
...
+      updated_at: null,
+    };
+
+    const userDataFromDB = await users.fetchUser({ email: userData.email });
+
+    if (userDataFromDB.userExists) {
+      if (userDataFromDB.user.roles?.developer) {
+        const errorMessage = encodeURIComponent("Google login is restricted for developer role.");
+        const separator = authRedirectionUrl.search ? "&" : "?";
+        return res.redirect(`${authRedirectionUrl}${separator}error=${errorMessage}`);
```
prakashchoudhary07, line 82: Are we going to restrict developers to not log in from google?

test/integration/auth.test.js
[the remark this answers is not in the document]
listiclehub1: There might be a case where google doesn't return any verified email of the user and the emails returned are marked `verified: False;` so I am checking for that case.

controllers/auth.js
listiclehub1: Yes Ankush had suggested to do so.

controllers/auth.js
listiclehub1: Should i define the cookie options as a constant

listiclehub1: done

test/integration/auth.test.js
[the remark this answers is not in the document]
prakashchoudhary07: Didn't get it

test/integration/auth.test.js
prakashchoudhary07: is there no way to have a feature flag similar behavior to achieve here?

controllers/auth.js
prakashchoudhary07: So, basically, you are saying that during OAuth, if an email is private, it is not shared, and by making an API call, it is shared.
Can you share any reference which confirms this

Also what token will be used to get the emails?

has this been tested?

test/integration/auth.test.js
listiclehub1: Google and other oauth providers only allow state, redirect uri, etc parameters to be passed rest are ignored. [Mentioned here The Google authorization server supports the following query string parameters for web server applications:](https://developers.google.com/identity/protocols/oauth2/web-server#creatingclient).

controllers/auth.js
listiclehub1: I have tested it and it is working. [Stackover flow issue here explains this](https://stackoverflow.com/questions/35373995/github-user-email-is-null-despite-useremail-scope)

test/integration/auth.test.js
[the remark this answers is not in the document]
listiclehub1: So typically google oauth user will always have email. I just added this as a safety check as it was suggested by Achintya.

test/integration/auth.test.js
listiclehub1: Moreover this route cannot be accessed directly by any user if accesed directly it will always give 401.

prakashchoudhary07 (approved):

Achintya-Chatterjee (approved):

iamitprakash (approved):

PR #2372, opened 2025-01-26
Add support to send different error response from error handler middleware

### LINKED ISSUE #2371: Fix error handler middleware to serve different error response
### Issue Description
The current problem we have in our backend codebase is that we are not using error middleware to send error responses to the client. We send error responses directly from controller based on different error cases which keeps on adding redundant code for error handling.

### Expected Behavior
- We should use error middleware to send error response based on type of error.
- This will make error middleware a central place for handling error responses.
- The controller/service has to throw an error which will be caught in catch block and we should call next middleware with error object. This will call the error handler middleware and based on type of error client will see a error response.

### Current Behavior
 - We are not using error middleware to handle error response.
 - We do not have a structured way to handle error response.
 - Because of above reasons, we keep on adding redundant code to handle error responses.

### Example
 - Let just see an example with this function where when request is not found by Id, will throw a not found error with message.
Image

- Now the error will propagate to controller because the function is being called here.
Image

- Because of error, the control will directly jump to catch block. Here we can log the error and will call next middleware with error object.
Image

- As soon as, next middleware is called with an error object, this error middleware is called and based on type of error it will send error response.
Image

 - API Response when request is not found with current approach
 
Image

### Reproducibility

 - [x] This issue is reproducible
 - [ ] This issue is not reproducible

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low
### Additional Information
 - We are sending these many different responses with respective status code.
 - Each error response can be mapped to respective error ```http-errors``` class.
```
badImplementation - 500 - InternalServerError
notFound - 404 - NotFound
serverUnavailable - 503 - ServiceUnavailable
forbidden - 403 - Forbidden
unauthorized - 401 - Unauthorized
badRequest - 400 - BadRequest
conflict - 409 - Conflict
internal - 500 - InternalServerError
notImplemented - 501 - NotImplemented
unsupportedMediaType - 415 - UnsupportedMediaType
badData - 422 - UnprocessableEntity
entityTooLarge - 413 - PayloadTooLarge
``` 
### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

pankajjs:
Date: 26 Jan, 2025

Developer Name: @pankajjs 

---

## Issue Ticket Number

- Closes #2371 
## Description
This PR fixes the error handler middleware by adding support to send different error responses based on error types.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

</details>

## Test Coverage
<details>

</details>

### DIFF
--- a/app.js
+++ b/app.js
@@ -23,12 +23,13 @@ app.use(function (req, res, next) {
 });
 
 // error handler
-app.use(function (err, req, res, next) {
+app.use(function (err, req, res, _next) {
   if (isMulterError(err)) {
-    multerErrorHandling(err, req, res);
-  } else {
-    res.boom.notFound(err);
+    return multerErrorHandling(err, req, res);
   }
+  return res.boom.boomify(err, {
+    statusCode: err.statusCode,
+  });
 });
 
 module.exports = app;


app.js
```diff
@@ -23,12 +23,13 @@ app.use(function (req, res, next) {
 });
 
 // error handler
-app.use(function (err, req, res, next) {
+app.use(function (err, req, res, _next) {
```
Achintya-Chatterjee, line 26: `_next` why this change ?

pankajjs: It is a good practise to prefix the variable with underscore when it is unused in the function but required in function parameter. It also prevents warning.
Considering it, I have added _ before next middleware.

AnujChhikara (approved):
LGTM

mbramani (approved):

Prakhar-FF13 (approved):

iamitprakash (approved):

Issue #2338, opened 2025-01-06
Define Default Size for extension-requests API to Prevent Returning All Data


AnujChhikara:
### Issue Description

The `get extension-requests` API currently lacks a default page size. When the `size` query parameter is not provided, the API returns all available data. This can lead to performance issues, especially when the dataset is large. To improve the usability and efficiency of the API, a default page size should be implemented.
### Expected Behavior

- Define a default page size for the get extension-requests API (e.g., 20 records per page).
### Current Behavior

 - When the size query parameter is not provided , the API returns all available records instead of limiting the number of records.
### Reproducibility

 - [x] This issue is reproducible
 - [ ] This issue is not reproducible

### Steps to Reproduce
- Send a GET request to the extension-requests API endpoint without the size query parameter. [extension-requests](https://api.realdevsquad.com/extension-requests)
- See it is returning over 1k response if we not provide size query param

### Severity/Priority
 - [ ] Critical
 - [x] High
 - [ ] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

AnujChhikara:
Closing this issue as it is being resolved in [this PR](https://github.com/Real-Dev-Squad/website-backend/pull/2378).

PR #2378, opened 2025-02-07
fix: Set default size for GET /extension-requests API

AnujChhikara:
Date: 7 Feb 2025

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

- #2338 
## Description

- Added a default size=5 for the /extension-requests API when the size query parameter is not provided.
- Wrote a test to verify that the API returns exactly 5 extension requests by default.

### Documentation Updated?

- [x] Yes
- [ ] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

- In the video, you can see that if the size query parameter is not provided, the response returns only five objects.

image

https://github.com/user-attachments/assets/ff075540-2b66-40ea-ae4c-aefb0e23e0d0

https://github.com/user-attachments/assets/24d5d1f6-1bb7-4e26-b41a-c64568a715b5
</details>

## Test Coverage
<details>

image

</details>

### DIFF
--- a/controllers/extensionRequests.js
+++ b/controllers/extensionRequests.js
@@ -105,7 +105,7 @@ const createTaskExtensionRequest = async (req, res) => {
  */
 const fetchExtensionRequests = async (req, res) => {
   try {
-    const { cursor, size, order } = req.query;
+    const { cursor, size = 5, order } = req.query;
     const { status, taskId, assignee } = parseQueryParams(req._parsedUrl.search);
     const { transformedSize, transformedStatus } = transformQuery(size, status);
 
--- a/test/integration/extensionRequests.test.js
+++ b/test/integration/extensionRequests.test.js
@@ -281,6 +281,24 @@ describe("Extension Requests", function () {
         });
     });
 
+    it("should return 5 extension requests by default when size query is not provided", function (done) {
+      chai
+        .request(app)
+        .get(`/extension-requests`)
+        .set("cookie", `${cookieName}=${appOwnerjwt}`)
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(200);
+          expect(res.body).to.be.a("object");
+          expect(res.body.message).to.be.equal("Extension Requests returned successfully!");
+          expect(res.body.allExtensionRequests).to.be.a("array");
+          expect(res.body.allExtensionRequests).to.have.lengthOf(5);
+          return done();
+        });
+    });
+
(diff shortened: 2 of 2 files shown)


controllers/extensionRequests.js
```diff
@@ -105,7 +105,7 @@ const createTaskExtensionRequest = async (req, res) => {
  */
 const fetchExtensionRequests = async (req, res) => {
   try {
-    const { cursor, size, order } = req.query;
+    const { cursor, size = 5, order } = req.query;
```
vikasosmium, line 108: can we give any number apart from 5?

vikasosmium:
is this video a GIF in screenshots 1?

not able to pause it to see the response.

controllers/extensionRequests.js
AnujChhikara: Yes we can send any number in the size query from 1-100. This is just a fallback value if we don't send size query Param then by default we will send 5 extension request.

AnujChhikara:
> is this video a GIF in screenshots 1?
> 
> not able to pause it to see the response.

It's a gif, changing it with a video

vikasosmium (approved):

shobhan-sundar-goutam (changes requested):
1. Why are we only testing in postman? Why not in actual dashboard site?
2. Why this is not under feature flag?

AnujChhikara:
> 1. Why are we only testing in postman? Why not in actual dashboard site?
> 2. Why this is not under feature flag?

- The issue being discussed is specifically related to the API response, which is why we are testing it in Postman. However, to ensure complete verification, I have also included a video demonstration of the dashboard site. In the video, you can see that when no size query parameter is provided, the API returns only 5 requests by default. As we scroll down, additional requests are fetched dynamically.
- This change does not require a feature flag because size=5 is already the default behavior on the extension-requests page. You can refer to the relevant code here: [Constants.js](https://github.com/Real-Dev-Squad/website-dashboard/blob/c74e1f4230cf3fd4ea4b4552e2d2423a58d9860e/extension-requests/constants.js#L11). The only issue is that the page size is determined from the URL. If we remove size=5 from the URL, the page currently loads all requests at once, which is not the expected behavior. With these changes, even if the size=5 query parameter is removed from the URL, the default behavior remains intact, ensuring limited API requests.

iamitprakash (approved):

Issue #2397, opened 2025-03-29
Bug: Any user can modify extension requests instead of just super users and request owners


AnujChhikara:
### Issue Description

- Currently, any normal user is able to update any extension request.

- Ideally, only super users and the request owner should be able to update an extension request.

- Update the backend to enforce this restriction.

### Expected Behavior

- Only super users and the request owner should be able to update an extension request.

- Normal users should not be able to modify requests they do not own.

- The API should return an appropriate response if an unauthorized user attempts to update a request.

### Current Behavior

- Any normal user is able to update any extension request using ?dev=true.

- There are no restrictions preventing unauthorized users from modifying requests.
### Screenshots

![Image](https://github.com/user-attachments/assets/4696d0a1-3879-4575-a2b5-d965387ad6d4)

![Image](https://github.com/user-attachments/assets/52588859-2869-4fec-a570-abb0a53475e2)

### Reproducibility

 - [x] This issue is reproducible
 - [ ] This issue is not reproducible

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

PR #2398, opened 2025-03-29
Fix: Allow Only Super Users or Request Owners to Edit Requests

AnujChhikara:
Date: 29 March 2025

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

- #2397 
## Description
- Updated authorization logic to ensure that only superusers or the request assignee can edit extension requests.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

Screenshot from 2025-03-29 23-24-07

</details>

## Test Coverage
<details>

Screenshot from 2025-03-29 23-25-26
Screenshot from 2025-03-29 23-28-20

</details>

### DIFF
--- a/controllers/extensionRequests.js
+++ b/controllers/extensionRequests.js
@@ -218,6 +218,14 @@ const updateExtensionRequest = async (req, res) => {
       return res.boom.badRequest("Only pending extension request can be updated");
     }
 
+    if (
+      isDev &&
+      !req.userData?.roles.super_user &&
+      extensionRequest.extensionRequestData.assigneeId !== req.userData.id
+    ) {
+      return res.boom.forbidden("You don't have permission to update the extension request");
+    }
+
     if (req.body.assignee) {
       const { taskData: task } = await tasks.fetchTask(extensionRequest.extensionRequestData.taskId);
       if (task.assignee !== (await getUsername(req.body.assignee))) {
--- a/test/integration/extensionRequests.test.js
+++ b/test/integration/extensionRequests.test.js
@@ -19,10 +19,11 @@ const { LOGS_FETCHED_SUCCESSFULLY } = require("../../constants/logs");
 chai.use(chaiHttp);
 
 const user = userData[6];
+const user2 = userData[5];
 const appOwner = userData[3];
 const superUser = userData[4];
 
-let appOwnerjwt, superUserJwt, jwt, superUserId, extensionRequestId5;
+let appOwnerjwt, superUserJwt, jwt, jwt2, superUserId, extensionRequestId5;
 
 describe("Extension Requests", function () {
   let taskId0,
@@ -40,6 +41,7 @@ describe("Extension Requests", function () {
 
   before(async function () {
     const userId = await addUser(user);
+    const userId2 = await addUser(user2);
     user.id = userId;
     const appOwnerUserId = await addUser(appOwner);
     appOwner.id = appOwnerUserId;
@@ -47,6 +49,7 @@ describe("Extension Requests", function () {
     appOwnerjwt = authService.generateAuthToken({ userId: appOwnerUserId });
     superUserJwt = authService.generateAuthToken({ userId: superUserId });
     jwt = authService.generateAuthToken({ userId: userId });
+    jwt2 = authService.generateAuthToken({ userId: userId2 });
 
     const taskData = [
       {
@@ -1090,6 +1093,26 @@ describe("Extension Requests", function () {
             return done(err);
           }
           expect(res).to.have.status(204);
+          return done();
+        });
+    });
+
+    it("Should return a 403 error if a non-superuser and non-owner try to update the request with the dev flag enabled.", function (done) {
+      chai
+        .request(app)
+        .patch(`/extension-requests/${extensionRequestId4}?dev=true`)
+        .set("cookie", `${cookieName}=${jwt2}`)
+        .send({
+          title: "new-title",
+        })
+        .end((err, res) => {
+          if (err) {
+            return done(err);
+          }
+          expect(res).to.have.status(403);
+          expect(res.body)
+            .to.have.property("message")
+            .that.equals("You don't have permission to update the extension request");
           return done();
         });
     });


test/integration/extensionRequests.test.js
MayankBansal12, line 1: should we also add a test to see if a super user can update the request?

controllers/extensionRequests.js
```diff
@@ -218,6 +218,14 @@ const updateExtensionRequest = async (req, res) => {
       return res.boom.badRequest("Only pending extension request can be updated");
     }
 
+    if (
+      isDev &&
+      !req.userData?.roles.super_user &&
```
MayankBansal12, line 223: `!req.userData?.roles.super_user` is being used twice in the func, better to use a variable for it

test/integration/extensionRequests.test.js
AnujChhikara: Those tests already existsThose tests already exists [Here is link](https://github.com/Real-Dev-Squad/website-backend/blob/6146736bfac7b48139cc4d64cbf2aa0a781728af/test/integration/extensionRequests.test.js#L1031)

controllers/extensionRequests.js
AnujChhikara: fixed

controllers/extensionRequests.js
```diff
@@ -204,20 +204,21 @@ const getSelfExtensionRequests = async (req, res) => {
 const updateExtensionRequest = async (req, res) => {
   const { dev } = req.query;
   const isDev = dev === "true";
+  const isSuperUser = req.userData?.roles.super_user;
```
yesyash, line 207: - is this how we get user role throughout the application?

AnujChhikara: In most parts of the application, we use the authorizeRoles [middleware](https://github.com/AnujChhikara/realdev-website-backend/blob/develop/middlewares/authorizeRoles.js) to handle role-based access control. However, in some specific cases where we can't block the route entirely based on roles, we manually check user roles using req.userData?.roles. This approach is used in 5-6 places

MayankBansal12 (approved):

Suvidh-kaushik (approved):
Just a small suggestion but shouldn’t the PR title have Extension Request instead of just Requests,
Everything else LGTM.!

AnujChhikara:
> Just a small suggestion but shouldn’t the PR title have Extension Request instead of just Requests,
> Everything else LGTM.!

That's a good suggestion, fixed. Thanks

controllers/extensionRequests.js
```diff
@@ -204,20 +204,21 @@ const getSelfExtensionRequests = async (req, res) => {
...
     if (!extensionRequest.extensionRequestData) {
       return res.boom.notFound("Extension Request not found");
     }
 
-    if (
-      isDev &&
-      !req.userData?.roles.super_user &&
-      extensionRequest.extensionRequestData.status !== EXTENSION_REQUEST_STATUS.PENDING
-    ) {
+    if (isDev && !isSuperUser && extensionRequest.extensionRequestData.status !== EXTENSION_REQUEST_STATUS.PENDING) {
```
pankajjs, line 214: do we need to check for super user here?

pankajjs (changes requested):

controllers/extensionRequests.js
```diff
@@ -204,20 +204,21 @@ const getSelfExtensionRequests = async (req, res) => {
...
       return res.boom.badRequest("Only pending extension request can be updated");
     }
 
+    if (isDev && !isSuperUser && extensionRequest.extensionRequestData.assigneeId !== req.userData.id) {
```
pankajjs, line 218: Should we check if user is archived or not?

test/integration/extensionRequests.test.js
```diff
@@ -40,13 +41,15 @@ describe("Extension Requests", function () {
 
   before(async function () {
     const userId = await addUser(user);
+    const userId2 = await addUser(user2);
```
pankajjs, line 44:
```suggestion
    const userId2 = await addUser(userData[5]);
```

no need for variable if it is not being used in multiple places or required

test/integration/extensionRequests.test.js
```diff
@@ -1094,6 +1097,26 @@ describe("Extension Requests", function () {
         });
     });
 
+    it("Should return a 403 error if a non-superuser and non-owner try to update the request with the dev flag enabled.", function (done) {
```
pankajjs, line 1100:
```suggestion
    it("should return forbidden response if superuser or request owner does not update the request when dev is enabled", function (done) {
```

test/integration/extensionRequests.test.js
```diff
@@ -1094,6 +1097,26 @@ describe("Extension Requests", function () {
...
+      chai
+        .request(app)
+        .patch(`/extension-requests/${extensionRequestId4}?dev=true`)
```
pankajjs, line 1103: Are we not using constant for url?

pankajjs (changes requested):

controllers/extensionRequests.js
shobhan-sundar-goutam: But this is the block for checking if the extension request is in pending state or not right? Is the check for super user really needed here?

shobhan-sundar-goutam: But this is the block for checking if the extension request is in pending state or not right? Is the check for super user really needed here?

AnujChhikara: @shobhan-sundar-goutam @pankajjs  updated the reply please check

test/integration/extensionRequests.test.js
AnujChhikara: NO!

test/integration/extensionRequests.test.js
AnujChhikara: fixed

pankajjs:
I think, we should update the api doc for the forbidden response.

Achintya-Chatterjee (commented):
can you update the API contract and attach it here for reference

AnujChhikara:
@Achintya-Chatterjee @pankajjs  added API contract pr in additional notes

pankajjs (approved):
LGTM

Achintya-Chatterjee (approved):

shobhan-sundar-goutam (approved):

Achintya-Chatterjee (approved):

controllers/extensionRequests.js
```diff
@@ -204,20 +204,21 @@ const getSelfExtensionRequests = async (req, res) => {
...
+      return res.boom.forbidden("You don't have permission to update the extension request");
+    }
```
tejaskh3, line 220: Shouldn't the message be "You don't have a message to update this extension request"?

tejaskh3 (approved):
@AnujChhikara , Just a small change needed according to me rest all LGTM, let me know if you can change that, I'lll re-approve that PR then

controllers/extensionRequests.js
AnujChhikara: Thanks for the suggestion! I believe the original message is more appropriate here since it's about permissions, not missing content.

iamitprakash (approved):

PR #2498, opened 2025-11-06
bugfix:corrected memberCount to ensure only active members are counted

### LINKED ISSUE #2495: Group Page Revamp — Accurate Member Count
Date: 6/11/2025

Developer Name: @JAHANWEE 

## Description

 * The Groups page currently **inflates member counts** by including users who’ve left the Discord server.
* [PRD for group revamp 2025](https://docs.google.com/document/d/1-mhxnNIJBjQe7aGw8nLefbL5JFhxoJcWhbEj1BBkVJo/edit?usp=sharing)

 
## Current behaviour
 * Member count also includes inactive users
 
 ## Expected Behaviour
 * Member count should reflect **only active Discord members** in real time.

Screenshots

Image

In the backend group, there are 65 members, while the current RDS developer count is <20.

JAHANWEE:
Date: 7 Nov 2025

Developer Name: @JAHANWEE

---

## Issue Ticket Number

- Closes #2495 
- Closes #https://github.com/Real-Dev-Squad/website-dashboard/pull/1028
## Description

- Fixed `memberCount` calculation in `models/discordactions.js` to include only users with `in_discord: true`.
- Added filtering logic to fetch and count only active Discord members.
- Ensured tests accurately reflect active vs inactive Discord users.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

</details>

https://github.com/user-attachments/assets/dd9bc6d8-24e1-44e9-ba21-82d439c91672

## Test Coverage
<details>
Screenshot 2025-11-07 050338

</details>


models/discordactions.js
```diff
@@ -311,12 +311,32 @@ const enrichGroupDataWithMembershipInfo = async (discordId, groups = []) => {
...
+        return snap.docs.map((doc) => doc.data());
+      })
+    );
+
+    const usersInDiscordSet = new Set(
+      usersSnapshots
+        .flat()
+        .filter((user) => user?.roles?.in_discord === true)
+        .map((user) => user.discordId)
+    );
```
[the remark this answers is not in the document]
JAHANWEE: Implemented the changes. Please check

models/discordactions.js
```diff
@@ -311,12 +311,34 @@ const enrichGroupDataWithMembershipInfo = async (discordId, groups = []) => {
...
+
+    const usersInDiscordSet = new Set();
+    for (const userList of usersSnapshots) {
+      for (const user of userList) {
+        if (user?.roles?.in_discord === true) {
+          usersInDiscordSet.add(user.discordId);
+        }
+      }
+    }
```
MayankBansal12: can we avoid nesting here? find a way to simplify this?

MayankBansal12: you can try flattening the usersSnapshots to avoid nesting here

eg:
```js
const usersInDiscordSet = new Set(
  usersSnapshots
    .flatMap(userList => userList) // flatten the array of arrays into one array
    .filter(user => user?.roles?.in_discord === true)
    .map(user => user.discordId)
);

Comment (JAHANWEE) [reply models/discordactions.js]: image
 
I was doing this before, but after getting pointed at by @korbit ai and discussing with @JC-Coder , we changed this. How to solve this? Attaching link here: https://github.com/Real-Dev-Squad/website-backend/pull/2498#discussion_r2501172430

File: test/unit/models/discordactions.test.js [hunk 3]
```diff
@@ -477,7 +477,7 @@ describe("discordactions", function () {
       const result = await enrichGroupDataWithMembershipInfo(userData[0].discordId, newGroupData);
       expect(result[0]).to.deep.equal({
         ...newGroupData[0],
-        memberCount: 2,
+        memberCount: 1,
```
iamitprakash: did you test with 0 and -ve value

iamitprakash (changes requested):

test/unit/models/discordactions.test.js
JAHANWEE: "0" case one is already covered. The current implementation cannot produce a negative memberCount because it only increments by 1 when a user is in Discord. There is no subtraction being done. So I have not added a test for it.

iamitprakash (changes requested):

test/unit/models/discordactions.test.js
iamitprakash: why you modified 2 to 1

iamitprakash (changes requested):

models/discordactions.js
```diff
@@ -311,12 +311,33 @@ const enrichGroupDataWithMembershipInfo = async (discordId, groups = []) => {
...
+    const usersSnapshots = await Promise.all(
+      discordIdChunks.map(async (chunk) => {
+        const snap = await userModel.where("discordId", "in", chunk).get();
+        return snap.docs.map((doc) => doc.data());
+      })
+    );
+
+    const usersInDiscordSet = new Set();
+
+    usersSnapshots.forEach((userList) => {
```
iamitprakash: If roles.in_discord is false or 'false' (string), the if will misbehave.

type caste before checking

like below

if (Boolean(user?.roles?.in_discord) === true)

iamitprakash (changes requested):

models/discordactions.js
JAHANWEE: I guess in this case, roles.in_discord is already defined as a boolean in the schema, so it should only ever be true or false. Since it’s not expected to be a string, I think the extra Boolean() cast might not be necessary here. 
Please correct me if I’m missing something?

test/unit/models/discordactions.test.js
JAHANWEE: The enrichGroupDataWithMembershipInfo function only counts users whose user.roles.in_discord === true. In the fixture for Group 1 only one of the two mapped users had roles.in_discord: true, so the correct memberCount is 1. 
The test expectation has been updated to match the implementation.

iamitprakash: @JAHANWEE not sure, but how come this was working before @yesyash  please visit this point too

models/discordactions.js
JAHANWEE: Acknowledged and updated. I verified that Firestore does not enforce strict field types — even if a field previously stored a boolean, it can later be overwritten with a string without any schema restrictions. 
The implementation has been adjusted accordingly

test/unit/models/discordactions.test.js
JAHANWEE: [The link to 0 test case ](https://github.com/JAHANWEE/website-backend/blob/0f75af99f0b37d1507ab28070ca6dbfa873de8be/test/unit/models/discordactions.test.js#L496)Here result[2].memberCount is explicitly asserted to be 0, so the “0 members” scenario is tested.

- Regarding negative values: with the current implementation memberCount is computed only by incrementing:


```
roleIdToCountMap[mapping.roleid] =
  (roleIdToCountMap[mapping.roleid] ?? 0) + 1;
```

PR #2524, opened 2025-12-23
Fix group member count

Kavyashri217:
Date: 23-12-2025

Developer Name: Kavyashri K R

---

## Description

Changes made:

1. Updated enrichGroupDataWithMembershipInfo to compute memberCount from existing Discord role membership data.
2. No changes to Discord service integrations.
3. No frontend or schema changes.

What this changes:

1. Fixed inaccurate group member counts by updating how memberCount is computed in the backend.
2. memberCount is now derived from live Discord role membership.
3. Firestore group–user mappings continue to be used only for membership checks (isMember).

Scope: 
Backend

### Documentation Updated?

- [x] Yes
- [ ] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [ ] Yes
- [x] No

## Screenshots
<details>

image

</details>

## Test Coverage
<details>

</details>


services/discordMembersService.js
```diff
@@ -1,6 +1,8 @@
 const jwt = require("jsonwebtoken");
 const config = require("config");
-const DISCORD_BASE_URL = config.get("services.discordBot.baseUrl");
+
+const rawBaseUrl = config.get("services.discordBot.baseUrl");
+const DISCORD_BASE_URL = rawBaseUrl && rawBaseUrl !== "DISCORD_BASE_URL" ? rawBaseUrl : "http://localhost:8080";
```
MayankBansal12: why is this change required? isn't the config file already handling fallback value?

MayankBansal12:
@Kavyashri217 

- Please add screen recording for your changes not the code screenshot
- Please attach test coverage as well
- One of the Tests is failing, please fix that

models/discordactions.js
```diff
@@ -300,6 +301,8 @@ const updateDiscordImageForVerification = async (userDiscordId) => {
  * @param {Array<object>} groups - Array of group objects to process.
  * @returns {Promise<Array<object>>} - An array of group objects with enriched information.
  */
+
+/* istanbul ignore next */
```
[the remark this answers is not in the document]
Kavyashri217: /* istanbul ignore next */ was added to exclude a defensive error path that depends on external Discord failures and isn’t reliably testable. Without this, coverage reports false negatives even though the core logic is already covered.

services/discordMembersService.js
Kavyashri217: The config fallback works in prod, but in tests config.get() returns the literal key ("DISCORD_BASE_URL") when the value isn’t set. That results in invalid URLs like DISCORD_BASE_URL/discord-members and causes test failures.

This guard ensures we only use a valid URL and safely fall back to localhost in test/local environments.

test/unit/services/discordService.test.js
```diff
@@ -16,6 +16,10 @@ const discordMembersArray = require("../../fixtures/discordResponse/discord-resp
 // const { func } = require("joi");
 let fetchStub;
 describe("Discord services", function () {
+  before(function () {
+    process.env.DISCORD_BASE_URL = "http://localhost:8080";
+  });
```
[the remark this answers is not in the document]
Kavyashri217: Right. The fallback handling for DISCORD_BASE_URL is already covered in the service-level change linked above, so setting process.env.DISCORD_BASE_URL in this test is no longer required.

I’ll remove this block to avoid duplication.

.github/workflows/codeql.yml
iamitprakash: why this change needed for your task

models/discordactions.js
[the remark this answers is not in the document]
Kavyashri217: Removed the comment

services/discordMembersService.js
Kavyashri217: undone the changes as the config file was already handling the fallback value

PR #2584, opened 2026-02-26
Fix idle 7 days + calculation logic

### LINKED ISSUE #2038: Fix the flow of idle user tagging
### Issue Description
Currently there is issue in the flow of tagging the idle tag to the users. 

Existing bug that are already open:- 
- https://github.com/Real-Dev-Squad/website-backend/issues/2024
- https://github.com/Real-Dev-Squad/website-my/issues/590
- https://github.com/Real-Dev-Squad/website-status/issues/1155
- https://github.com/Real-Dev-Squad/todo-action-items/issues/210

### Reproducibility

 - [x] This issue is reproducible
 - [ ] This issue is not reproducible

### Steps to Reproduce
1. 
2. 

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [ ] Medium
 - [ ] Low

### Checklist

- [ ] I have read and followed the project's code of conduct.
- [ ] I have searched for similar issues before creating this one.
- [ ] I have provided all the necessary information to understand and reproduce the issue.
- [ ] I am willing to contribute to the resolution of this issue.

vinit717:
Date: 27-02-2026

Developer Name: @vinit717 

---

## Issue Ticket Number
 
- closes #2038 

## Description

Fix the 7-day+ idle users calculation logic

```text
CRON JOB START: Fetch all users where currentStatus.state == IDLE
  │
  ▼
FOR EACH USER: Determine Window Start
  ├─ If `idleWindowStartedAt` exists ──► windowStart = idleWindowStartedAt
  └─ If missing (fallback) ────────────► windowStart = currentStatus.from
  │
  ▼
FETCH SOURCE OF TRUTH OOO DATA
  ├─ Query Firestore `requests` collection:
  │    WHERE requestedBy == "user-id" 
  │      AND type == "OOO" 
  │      AND status (or state) == "APPROVED"
  │
  ▼
PROCESS OOO PERIODS
  ├─ 1. Filter out requests that happened before `windowStart`
  ├─ 2. Merge overlapping timestamps (e.g., 2 consecutive requests become 1 block)
  └─ Result = [Clean list of valid OOO date ranges]
  │
  ▼
CALCULATE IDLE DAYS
  ├─ Total calendar days = (Date.now() - windowStart)
  ├─ Approved OOO days = Sum of the merged OOO date ranges
  ├─ Net Idle Days = (Total calendar days) - (Approved OOO days)
  │
  ▼
APPLY OR REMOVE ROLE
  ├─ IF Net Idle Days >= 7:
  │    └─ Add `group-idle-7d+` Discord role
  ├─ ELSE:
  │    └─ Remove `group-idle-7d+` Discord role (if they currently hold it)
```

### Documentation Updated?

- [x] Yes
- [ ] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [x] Yes
- [ ] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>
<summary>
image </summary>

</details>

## Test Coverage
<details>
<summary>Screenshot 2026-03-02 at 6 29 43 PM</summary>

</details>

### DIFF
--- a/models/discordactions.js
+++ b/models/discordactions.js
@@ -8,7 +8,7 @@ const { findSubscribedGroupIds } = require("../utils/helper");
 const { retrieveUsers } = require("../services/dataAccessLayer");
 const { BATCH_SIZE_IN_CLAUSE } = require("../constants/firebase");
 const { getAllUserStatus, getGroupRole, getUserStatus } = require("./userStatus");
-const { normalizeTimestamp, checkIfUserHasLiveTasks } = require("../utils/userStatus");
+const { normalizeTimestamp, checkIfUserHasLiveTasks, computeIdleDaysExcludingOOO } = require("../utils/userStatus");
 const { userState, POST_OOO_GRACE_PERIOD_IN_DAYS } = require("../constants/userStatus");
 const config = require("config");
 const logger = require("../utils/logger");
@@ -680,26 +680,32 @@ const updateIdle7dUsersOnDiscord = async (dev) => {
     });
 
     if (allUserStatus) {
+      const nowMs = Date.now();
       await Promise.all(
         allUserStatus.map(async (userStatus) => {
-          const currentDate = new Date();
-          const lastDate = new Date(userStatus.currentStatus.from);
-          const ONE_DAY = 1000 * 60 * 60 * 24;
-          const timeDifference = currentDate.setUTCHours(0, 0, 0, 0) - lastDate.setUTCHours(0, 0, 0, 0);
-          const daysDifference = Math.floor(timeDifference / ONE_DAY);
           try {
-            if (daysDifference > 7) {
-              const userData = await userModel.doc(userStatus.userId).get();
-              const isUserArchived = userData.data().roles.archived;
-              if (userData.exists) {
-                if (isUserArchived) {
-                  totalArchivedUsers++;
-                } else if (dev === "true" && !allMavens.includes(userData.data().discordId)) {
-                  const shouldAdd = await shouldAddIdleUser(userStatus, tasksModel);
-                  if (shouldAdd) {
-                    userStatus.userid = userData.data().discordId;
-                    allIdle7dUsers.push(userStatus);
-                  }
+            const fullStatusDoc = await userStatusModel.doc(userStatus.id).get();
+            const fullData = fullStatusDoc.exists ? fullStatusDoc.data() : {};
+            const idleDays = computeIdleDaysExcludingOOO(
+              fullData.idleWindowStartedAt,
+              fullData.lastOooFrom,
+              fullData.lastOooUntil,
+              userStatus.currentStatus?.from,
+              nowMs
+            );
+            if (idleDays <= 7) {
+              return;
+            }
+            const userData = await userModel.doc(userStatus.userId).get();
+            const isUserArchived = userData.data()?.roles?.archived;
+            if (userData.exists) {
+              if (isUserArchived) {
+                totalArchivedUsers++;
+              } else if (dev === "true" && !allMavens.includes(userData.data().discordId)) {
+                const shouldAdd = await shouldAddIdleUser(userStatus, tasksModel);
+                if (shouldAdd) {
+                  userStatus.userid = userData.data().discordId;
+                  allIdle7dUsers.push(userStatus);
                 }
               }
             }
--- a/models/discordactions.js
+++ b/models/discordactions.js
@@ -8,7 +8,12 @@ const { findSubscribedGroupIds } = require("../utils/helper");
 const { retrieveUsers } = require("../services/dataAccessLayer");
 const { BATCH_SIZE_IN_CLAUSE } = require("../constants/firebase");
 const { getAllUserStatus, getGroupRole, getUserStatus } = require("./userStatus");
-const { normalizeTimestamp, checkIfUserHasLiveTasks, computeIdleDaysExcludingOOO } = require("../utils/userStatus");
+const {
+  getApprovedOooPeriods,
+  normalizeTimestamp,
+  checkIfUserHasLiveTasks,
+  computeIdleDaysExcludingOOO,
+} = require("../utils/userStatus");
 const { userState, POST_OOO_GRACE_PERIOD_IN_DAYS } = require("../constants/userStatus");
 const config = require("config");
 const logger = require("../utils/logger");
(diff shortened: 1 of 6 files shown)


models/userStatus.js
```diff
@@ -241,14 +241,10 @@ const updateUserStatus = async (userId, updatedStatusData) => {
           }
         }
       }
-      const lastOooUntilUpdate = resolveLastOooUntil({
```
MayankBansal12, line 244: previously we were storing `lastOooUntilUpdate` in userStatus, since we are removing that here can you please confirm that isn't being used anywhere?

vinit717: lastOooUntilUpdate isn't being used anywhere anymore
the fields lastOooFrom, lastOooUntil, and oooPeriods are effectively dead in terms of being updated in the usersStatus collection. The new getApprovedOooPeriods() handles everything dynamically via the requests collection now

MayankBansal12: https://github.com/RealDevSquad/website-backend/blob/a07167fa08c1cf72ce3da14bab89cda15729a1f8/models/discordactions.js#L1156

can you please check `getMissedProgressUpdatesUsers` in discordactions

vinit717: lastOooUntil is still stored in the userStatus collection and is used by getMissedProgressUpdatesUsers (around line 1180) to apply a grace period check.

Even though we stopped updating lastOooFrom/lastOooUntil during the refactor, the field was intentionally kept for backward compatibility. Existing users may still have valid lastOooUntil values, which the cron job relies on to grant grace periods.

In the future, this can be refactored to use the new requests collection (getApprovedOooPeriods()), but for now, everything remains safe and functional, and no behavior is broken

Since now we are just targeting the idle user, not the progress calculation, so I didn't update it here

utils/userStatus.js
```diff
@@ -34,30 +37,87 @@ const normalizeTimestamp = (value) => {
...
+ * @param {number} windowStart - Start of the idle window in ms
+ * @param {number} windowEnd - End of the window (typically Date.now()) in ms
+ * @returns {Promise<Array<{from:number,until:number}>>} Merged OOO periods overlapping the window
  */
-const resolveLastOooUntil = ({ previousState, previousUntil, nextState, fallbackTimestamp }) => {
-  if (nextState === userState.OOO) {
-    return null;
+const getApprovedOooPeriods = async (userId, windowStart, windowEnd) => {
+  try {
+    const snapshot = await requestsModel.where("requestedBy", "==", userId).where("type", "==", "OOO").get();
```
AnujChhikara: I think we should get using the date and status === APPROVED filter, as we don't need all OOO requests

utils/userStatus.js
```diff
@@ -34,30 +37,87 @@ const normalizeTimestamp = (value) => {
   return null;
 };
 
+const ONE_DAY_MS = 1000 * 60 * 60 * 24;
```
AnujChhikara:
```suggestion
const ONE_DAY_MS = 24 * 60 * 60 * 1000;
```

test/integration/taskBasedStatusUpdate.test.js
AnujChhikara, line 1: I think we should write couple of test or update exiting testing user status updating correctly newly added field idleWindowStartedAt

models/userStatus.js
```diff
@@ -568,16 +555,10 @@ const batchUpdateUsersStatus = async (users) => {
 
         if (timeDifferenceDays >= 1) {
           if (state === userState.IDLE) await addGroupIdleRoleToDiscordUser(userId);
-          const lastOooUntilUpdate = resolveLastOooUntil({
-            previousState: currentState,
-            previousUntil: currentUntil,
-            nextState,
-            fallbackTimestamp: currentTimeStamp,
-          });
```
AnujChhikara, line 576: is there any specific reason why we have removed last OOO until reference from everywhere

models/userStatus.js
```diff
@@ -705,19 +682,32 @@ const cancelOooStatus = async (userId) => {
...
+        .orderBy("createdAt", "desc")
+        .limit(1)
+        .get();
+
+      if (!oooRequests.empty) {
+        const requestDoc = oooRequests.docs[0];
+        const requestData = requestDoc.data();
+        const isApproved = requestData.status === "APPROVED" || requestData.state === "APPROVED";
+        if (isApproved) {
+          await requestsCollection.doc(requestDoc.id).update({ until: Date.now() });
```
AnujChhikara: This looks reasonable. But this is a product behavior let’s confirm with product before we lock this in.

models/discordactions.js
```diff
@@ -679,27 +688,42 @@ const updateIdle7dUsersOnDiscord = async (dev) => {
...
-                    allIdle7dUsers.push(userStatus);
-                  }
+            if (!userStatus?.userId) {
+              logger.warn("updateIdle7dUsersOnDiscord: skipping user status with missing userId");
+              return;
+            }
+            const windowStart =
+              normalizeTimestamp(userStatus.idleWindowStartedAt) ??
+              normalizeTimestamp(userStatus.currentStatus?.from) ??
+              nowMs;
```
AnujChhikara: why are we using from as fallback here

models/discordactions.js
```diff
@@ -679,27 +688,42 @@ const updateIdle7dUsersOnDiscord = async (dev) => {
...
+            const clampedWindowStart = Math.min(windowStart, nowMs);
```
AnujChhikara: better variable name please

models/userStatus.js
vinit717: We computed and stored lastOooUntil on usersStatus on every status change, which is redundant; it could go stale. In the new change, we query the requests collection fresh at cron time. So we don't need to store a copy.

models/discordactions.js
vinit717: Without the fallback, all legacy IDLE users (who have no idleWindowStartedAt) would always get 0 idle days and never be tagged, breaking the feature for existing users

models/discordactions.js
vinit717: updated

test/integration/taskBasedStatusUpdate.test.js
vinit717: done

utils/userStatus.js
vinit717: done

utils/userStatus.js
vinit717: The reason we don't add .where("status", "==", "APPROVED") directly to the database query is that our data has a mix of old and new formats.

Older OOO requests stored the approval under the status field, while newer requests store it under the state field.

If we add .where("status", "==", "APPROVED") to the query, we would miss all the newer requests that use state === "APPROVED". Conversely, if we query by state, we miss the old ones.

Firestore historically struggles with OR queries across different fields without complex composite indexing. Since a single user will only have a very small handful of OOO requests, fetching them and doing the status === APPROVED || state === APPROVED check in-memory is much safer, guarantees we don't miss legacy documents, and avoids DB indexing errors

models/userStatus.js
vinit717: I think this is not product behavior, as this is the implementation we can discuss more on a call

AnujChhikara (dismissed):

utils/userStatus.js
```diff
@@ -51,13 +54,87 @@ const resolveLastOooUntil = ({ previousState, previousUntil, nextState, fallback
...
+      .get();
+
+    const periods = [];
+    snapshot.forEach((doc) => {
+      const data = doc.data();
+      const isApproved = data.state === "APPROVED";
+      if (!isApproved) return;
+
+      const from = normalizeTimestamp(data.from);
+      const until = normalizeTimestamp(data.until);
```
prakashchoudhary07, line 100: @vinit717  This can be part of the query itself, get the approved requests only

prakashchoudhary07 (commented):
Approving it for now.
A follow-up PR will be raised by @vinit717  for improvements, idle time edge case fix, and also for the issue where users come back early from OOO, then the status for them

prakashchoudhary07 (approved):

PR #2602, opened 2026-07-07
refactor: reduce Firestore reads on users & usersStatus collections

AnujChhikara:
Date: 7 July , 2026

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

- https://github.com/RealDevSquad/website-backend/issues/2603
## Tech Doc Link

- NA
## Business Doc Link

- NA
## Description
Most reads scanned entire collections when only non-archived (active) users are needed. Switch the hot paths to active-users-first, paginated, batched reads.

GET /users/status:
- Paginate non-archived users (fetchNonArchivedUsers, doc-id cursors) then batch-fetch their statuses (getUserStatusForUserIds, chunked `in`) instead of scanning all statuses with per-user lookups.
- links.next/prev only emitted when a page exists in that direction.

Discord idle crons (group-idle, group-idle-7d):
- Active-users-first: page non-archived users -> their statuses -> filter IDLE, avoiding stale IDLE statuses of archived users and per-user reads.
- Fix 7-day idle calc: align the OOO fetch window with the idle-day window (idleFrom ?? currentStatus.from ?? now) so past OOO is excluded when idleFrom is missing.

Other:
- getTaskBasedUsersStatus: per-user tasks query -> chunked `in`.
- fetchAllUsers({ nonArchivedOnly }): active-only option, used by markUnverified and updateDiscordNicknames.
- Reuse getPaginationLink util (basePath param) instead of a duplicate helper.

Adds unit/integration tests and docs/firestore-read-optimization.md.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [ ] Yes
- [x] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>

image
image

</details>

## Test Coverage
<details>

image
image

</details>

### DIFF
--- a/controllers/userStatus.js
+++ b/controllers/userStatus.js
@@ -1,8 +1,9 @@
 const { Forbidden, NotFound } = require("http-errors");
 const { getUserIdBasedOnRoute } = require("../utils/userStatus");
+const { getPaginationLink } = require("../utils/users");
 const { INTERNAL_SERVER_ERROR } = require("../constants/errorMessages");
-const dataAccess = require("../services/dataAccessLayer");
 const userStatusModel = require("../models/userStatus");
+const usersModel = require("../models/users");
 const { userState, CANCEL_OOO } = require("../constants/userStatus");
 const ROLES = require("../constants/roles");
 const firestore = require("../utils/firestore");
@@ -75,25 +76,49 @@ const getUserStatus = async (req, res) => {
  */
 const getAllUserStatus = async (req, res) => {
   try {
-    const { allUserStatus } = await userStatusModel.getAllUserStatus(req.query);
-    const activeUsers = [];
-    if (allUserStatus) {
-      const allUsersStatusFetchPromises = allUserStatus.map(async (status) => {
-        //  fetching users from users collection with the help of userID in userStatus collection
-        const result = await dataAccess.retrieveUsers({ id: status.userId });
-        if (!result.user?.roles?.archived) {
-          status.full_name = `${result.user.first_name} ${result.user.last_name}`;
-          status.picture = result.user.picture;
-          status.username = result.user.username;
-          activeUsers.push(status);
-        }
+    const { state, next, prev, size } = req.query;
+    const pageSize = parseInt(size) || 100;
+
+    const { users, nextId, prevId } = await usersModel.fetchNonArchivedUsers({ next, prev, size: pageSize });
+
+    if (!users.length) {
+      return res.json({
+        message: "No users found.",
+        totalUserStatus: 0,
+        allUserStatus: [],
+        links: { next: "", prev: "" },
       });
-      await Promise.all(allUsersStatusFetchPromises);
     }
+
+    const userIds = users.map((u) => u.id);
+    const statusMap = await userStatusModel.getUserStatusForUserIds(userIds);
+
+    let allUserStatus = users.map((user) => {
+      const status = statusMap[user.id];
+      return {
+        id: status?.id ?? null,
+        userId: user.id,
+        currentStatus: status?.currentStatus ?? null,
+        monthlyHours: status?.monthlyHours ?? null,
+        idleFrom: status?.idleFrom ?? null,
+        full_name: `${user.first_name} ${user.last_name}`,
+        picture: user.picture,
+        username: user.username,
+      };
+    });
+
+    if (state) {
+      allUserStatus = allUserStatus.filter((entry) => entry.currentStatus?.state === state);
+    }
+
     return res.json({
       message: "All User Status found successfully.",
-      totalUserStatus: activeUsers.length,
-      allUserStatus: activeUsers,
+      totalUserStatus: allUserStatus.length,
+      allUserStatus,
+      links: {
+        next: nextId ? getPaginationLink(req.query, "next", nextId, "/users/status") : "",
+        prev: prevId ? getPaginationLink(req.query, "prev", prevId, "/users/status") : "",
+      },
     });
   } catch (err) {
     logger.error(`Error while fetching all the User Status: ${err}`);
--- a/models/discordactions.js
+++ b/models/discordactions.js
@@ -7,7 +7,7 @@ const admin = require("firebase-admin");
 const { findSubscribedGroupIds } = require("../utils/helper");
 const { retrieveUsers } = require("../services/dataAccessLayer");
 const { BATCH_SIZE_IN_CLAUSE } = require("../constants/firebase");
-const { getAllUserStatus, getGroupRole, getUserStatus } = require("./userStatus");
+const { getGroupRole, getUserStatus, getUserStatusForUserIds } = require("./userStatus");
 const {
   getApprovedOooPeriods,
   normalizeTimestamp,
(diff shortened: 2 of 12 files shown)


models/discordactions.js
```diff
@@ -397,14 +397,32 @@ const shouldAddIdleUser = async (userStatus, tasksModel) => {
   }
 };
 
+const getNonArchivedIdleUsersWithStatus = async () => {
```
MayankBansal12, line 400: instead of a separate do-while loop, why not use fetchAllUsers to get all users at once?

controllers/userStatus.js
```diff
@@ -75,25 +76,49 @@ const getUserStatus = async (req, res) => {
...
+    const pageSize = parseInt(size) || 100;
+
+    const { users, nextId, prevId } = await usersModel.fetchNonArchivedUsers({ next, prev, size: pageSize });
+
+    if (!users.length) {
+      return res.json({
+        message: "No users found.",
+        totalUserStatus: 0,
+        allUserStatus: [],
+        links: { next: "", prev: "" },
```
MayankBansal12, line 89: should next and prev be null or empty string?

models/discordactions.js
```diff
@@ -689,46 +686,38 @@ const updateIdle7dUsersOnDiscord = async (dev) => {
...
+          const idleDays = computeIdleDaysExcludingOOO(
+            status.idleFrom,
+            status.currentStatus?.from,
+            currentTime,
+            oooPeriods
+          );
+          if (idleDays < 7) {
+            return;
+          }
+          const discordId = user?.discordId;
```
MayankBansal12, line 708: why not check if the user is part of discord? like in func above? and place this check at start so user who are not part of discord are not avoided for idle period computation

models/discordactions.js
```diff
@@ -689,46 +689,39 @@ const updateIdle7dUsersOnDiscord = async (dev) => {
...
-            logger.error(`error updating discordId in userStatus ${error.message}`);
-            throw new Error("error updating discordId in userStatus");
           }
-        })
-      );
-    }
+        } catch (error) {
+          logger.error(`error updating discordId in userStatus ${error.message}`);
+          throw new Error("error updating discordId in userStatus");
+        }
```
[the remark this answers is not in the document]
MayankBansal12: @AnujChhikara please take a  look

controllers/userStatus.js
AnujChhikara: we are following the same patter in the /users api that why pick it from there

models/discordactions.js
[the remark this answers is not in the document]
AnujChhikara: it is already handled , check the lines 716-718 already logs and continues (logger.error(... idle-7d user ...)),

models/discordactions.js
AnujChhikara: The goal of this PR is only to optimize Firestore reads, not to refactor the existing code logic. I’m happy to create a separate ticket and address those improvements there.

As you can see, if I start handling additional edge cases in this PR, the scope will expand beyond the original intent.

models/discordactions.js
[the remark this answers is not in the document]
AnujChhikara: @coderabbitai please verify this it is already handled , check the lines 716-718 already logs and continues (logger.error(... idle-7d user ...)), if this comment is resolved

models/discordactions.js
MayankBansal12: it's a small change to maintain correctness, it's up to you if you want to update logic later.

AnujChhikara: We'll create another ticket for refactoring this function and see the ways we can optimise or refactor the logic itself, and will assign to you

test/unit/models/discordactions.test.js
```diff
@@ -1033,7 +1033,8 @@ describe("discordactions", function () {
       expect(res.totalIdle7dUsers).to.be.equal(2);
       expect(res.totalUserRoleToBeAdded).to.be.equal(2);
       expect(res.totalUserRoleToBeRemoved).to.be.equal(1);
-      expect(res.totalArchivedUsers).to.be.equal(1);
+      // Now fetches active-users-first, so archived users are never read/counted (was 1).
+      expect(res.totalArchivedUsers).to.be.equal(0);
```
MayankBansal12: i guess you can remove this comment?

test/integration/userStatus.test.js
```diff
@@ -97,6 +97,17 @@ describe("UserStatus", function () {
       expect(response.body.allUserStatus).to.be.a("array");
       expect(response.body.allUserStatus.length).to.equal(1);
     });
+
+    it("Should return pagination links with empty next/prev on a single page", async function () {
```
MayankBansal12, line 101: test for pagination in multiple pages case?

test/unit/models/users.test.js
```diff
@@ -313,6 +313,58 @@ describe("users", function () {
...
+      });
+      await userModel.add({
+        username: "archived-gamma",
+        first_name: "Archived",
+        last_name: "Gamma",
+        roles: { archived: true },
+      });
+    });
+
+    it("returns only non-archived users and empty cursors on a single page", async function () {
```
MayankBansal12, line 348: negative case for archived=true? and in discord=false?

test/unit/models/users.test.js
```diff
@@ -313,6 +313,58 @@ describe("users", function () {
...
+      const result = await users.fetchNonArchivedUsers({ size: 100 });
+      expect(result.users).to.be.an("array");
+      expect(result.users.length).to.equal(2);
+      result.users.forEach((user) => expect(user.roles.archived).to.equal(false));
+      expect(result.nextId).to.equal("");
+      expect(result.prevId).to.equal("");
+    });
+
+    it("paginates forward with the nextId cursor", async function () {
```
MayankBansal12, line 373: test for prevId and size limit?

AnujChhikara: added

MayankBansal12 (dismissed):

models/users.js
```diff
@@ -764,9 +764,10 @@ const getDiscordUsers = async () => {
   }
 };
 
-const fetchAllUsers = async () => {
+const fetchAllUsers = async ({ nonArchivedOnly = false } = {}) => {
```
prakashchoudhary07: Can we rename nonArchivedOnly to activeMember or activeUsers or some thing more appropriate?

models/users.js
```diff
@@ -764,9 +764,10 @@ const getDiscordUsers = async () => {
...
   const users = [];
-  const usersQuerySnapshot = await userModel.get();
+  const query = nonArchivedOnly ? userModel.where("roles.archived", "==", false) : userModel;
```
prakashchoudhary07: Also should this include in_discord filter?

prakashchoudhary07 (changes requested):

models/users.js
AnujChhikara: both are co related , every 30 minutes cron job trigger this POST /external-accounts/users, which keep these filed in sync if user is not found in discord it set archived true. Still we can add that condition in the query only thing is we might have to create new composite index for that.

models/users.js
AnujChhikara: I avoided activeMember because activeMember = false can be a bit confusing - it might sound like we're fetching inactive members. I think something like `filterActiveUsers` is a bit clearer since it indicates we're applying the active users filter.

prakashchoudhary07: okay please update it

models/users.js
prakashchoudhary07: I don't think we would need index here, but its its fine then ok

models/users.js
AnujChhikara: updated

prakashchoudhary07 (approved):

PR #2626, opened 2026-09-11
fix: the config/default in default.js

anupamme:
## Summary
Fix critical severity security issue in `config/default.js`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | CRITICAL |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `config/default.js:45` |
| **Assessment** | Likely exploitable |
| **Chain Complexity** | 2-step |

**Description**: The config/default.js file contains a hardcoded Firebase service account private key. This key grants administrative access to the Firebase project and appears to be a real RSA private key rather than a placeholder. Exposure of this key enables complete compromise of the Firebase/Firestore database.

## Evidence

**Exploitation scenario**: An attacker who gains read access to this configuration file through repository access, path traversal vulnerability, or log exposure can use the private key to authenticate as the firebase-adminsdk.

**Scanner confirmation**: multi_agent_ai rule `V-001` flagged this pattern.

**Production code**: This file is in the production codebase, not test-only code.

## Threat Model Context

This is a web service - vulnerabilities in request handlers are directly exploitable by remote attackers.

## Changes
- `config/default.js`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -48,14 +48,14 @@ module.exports = {
   firestore: `{
     "type": "service_account",
     "project_id": "<project-name>",
-    "private_key_id": "<private-key>",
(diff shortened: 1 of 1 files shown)


prakashchoudhary07:
The key that is in the config is a placeholder key; it was used for representational purposes. 

The issue is not relevant. So will be closing this PR. 

If there are any concerns or different thoughts on it, we can chat. And if required, reopen it later on

anupamme:
Thanks for clarifying. If the Firebase key in the previous revision is confirmed to be a non-functional/representational placeholder and is not associated with any active Firebase service account, then I agree that the credential-exposure finding does not apply.

My original finding treated the PEM material as a potentially valid service-account private key based on its presence in config/default.js; I did not independently establish that it was an active credential. I’ll consider the finding invalid under that clarification.

The replacement with placeholders is still a reasonable hardening improvement, but I agree it should not be characterised as remediation of a confirmed critical credential compromise.

